PR #1, opened 2023-03-04
Added basic API tests to the new Flask endpoint for neuroconv

garrettmflynn:
Implemented basic tests for the Flask endpoint in pytest after talking with @CodyCBakerPhD about the implementation and PR process.

PR #2, opened 2023-03-04
Added a basic API to request schema information from neuroconv

garrettmflynn:
This PR introduces a basic REST API to request schema information from `neuroconv`. The latest YAML environment files should install the latest version of `neuroconv` from Github, allowing users to simply follow the installation guidelines for SODA to get started.

This also includes additional changes to the SODA interface to align more with NWB GUIDE branding, as well as a fix for the broken functionality of the Guided Mode button.


package.json
```diff
@@ -167,7 +167,9 @@
     "husky": "^8.0.1",
     "lint-staged": "^13.0.3",
     "megasanjay-devmoji": "^1.1.0",
-    "prettier": "^2.3.2"
+    "npm-run-all": "^4.1.5",
+    "prettier": "^2.3.2",
+    "vite": "^4.1.4"
```
CodyCBakerPhD, line 172: Can you summarize what `npm-run-all` and `vite` do?

sections/guided_mode/guided_curate_dataset.html
```diff
@@ -121,45 +112,54 @@ <h2 class="guided--text-sub-step" id="guided-continue-curation-header"></h2>
...
+            <h1 class="guided--text-sub-step">Welcome to NWB GUIDE!</h1>
           </div>
           <div class="guided--section">
             <p class="guided--help-text">
-              Guided Mode is designed to be the simplest and most efficient method of curating and
-              sharing a SPARC dataset if you are starting from scratch. Guided Mode guides you
-              through the entire SPARC curation and sharing process via intuitive user interfaces
+              NWB GUIDE is designed to be the simplest and most efficient method of converting and
+              sharing NWB datasets if you are starting from scratch. NWB GUIDE guides you
+              through the entire NWB file conversion and sharing process via intuitive user interfaces
```
CodyCBakerPhD, line 121: It's fine for these first few PRs where we're just trying to clean everything up, but in the future I'd isolate changes to a single page like this into separate PRs for faster review and merge, especially if it's just text/wording instead of code

tools/anaconda-env/environment-MAC.yml
```diff
@@ -24,3 +24,4 @@ dependencies:
       - protobuf == 3.20.0
       - werkzeug == 2.0.2
       - opencv-python == 4.5.3.56
+      - neuroconv === git+https://github.com/catalystneuro/neuroconv.git@main
```
CodyCBakerPhD, line 27:
```suggestion
      - neuroconv @ git+https://github.com/catalystneuro/neuroconv.git@main
```

CodyCBakerPhD: This didn't work for me when I tried it - it seems it follows the same syntax as `pip install -r ...`  when you're under the `pip` heading of the `.yml`, in which case the syntax for dev branch installs is `<name of package alias> @ <the git+url.git@branch_or_commit>`

tools/anaconda-env/environment-Linux.yml
```diff
@@ -24,3 +24,4 @@ dependencies:
       - gevent == 21.12.0
       - werkzeug == 2.0.2
       - opencv-python == 4.5.3.56
+      - neuroconv === git+https://github.com/catalystneuro/neuroconv.git@main
```
CodyCBakerPhD, line 27:
```suggestion
      - neuroconv @ git+https://github.com/catalystneuro/neuroconv.git@main
```

yarn.lock
```diff
@@ -0,0 +1,10258 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
+# yarn lockfile v1
```
CodyCBakerPhD, line 2: Well, this is a quite a large file...

Can you summarize what this is for and why it has so many dependencies? Are all of them really necessary?

scripts/others/renderer.js
```diff
@@ -352,8 +352,105 @@ const wait = async (delay) => {
...
+    },
+  });
+
+  // Darwin executable starts slowly
+  // use an exponential backoff to wait for the app server to be ready
+  // this will give Mac users more time before receiving a backend server error message
+  // ( during the wait period the server should start )
+  // Bonus:  doesn't stop Windows and Linux users from starting right away
+  // NOTE: backOff is bad at surfacing errors to the console
+  //while variable is false keep requesting, if time exceeds two minutes break
```
CodyCBakerPhD, line 381:
```suggestion
  // while variable is false keep requesting, if time exceeds two minutes break
```

scripts/others/renderer.js
```diff
@@ -352,8 +352,105 @@ const wait = async (delay) => {
...
+
+  if (!status) {
+    //two minutes pass then handle connection error
+    // SWAL that the server needs to be restarted for the app to work
+    clientError(error);
+    ipcRenderer.send("track-event", "Error", "Establishing Python Connection", error);
+
+    await Swal.fire({
+      icon: "error",
+      html: `Something went wrong while initializing NWB GUIDE background services. Please restart SODA and try again. If this issue occurs multiple times, please email <a href='mailto:help@fairdataihub.org'>help@fairdataihub.org</a>.`,
```
CodyCBakerPhD, line 406:
```suggestion
      html: `Something went wrong while initializing NWB GUIDE background services. Please restart the application and try again. If this issue occurs multiple times, please raise an issue.`,
```

scripts/others/renderer.js
```diff
@@ -352,8 +352,105 @@ const wait = async (delay) => {
...
+    app.relaunch();
+    app.exit();
+  }
+
+  console.log("Connected to Python back-end successfully");
+  log.info("Connected to Python back-end successfully");
+  ipcRenderer.send("track-event", "Success", "Establishing Python Connection");
+
+  // inform observers that the app is connected to the server
+  sodaIsConnected = true;
```
CodyCBakerPhD, line 424:
```suggestion
  guideIsConnected = true;
```

CodyCBakerPhD: Or are there a bunch of other things right now that depend on it being called 'sodaIsConnect'?

scripts/others/renderer.js
```diff
@@ -432,176 +529,177 @@ const run_pre_flight_checks = async (check_update = true) => {
...
-        });
-        resolve(false);
-      }
+      // // Check for an API key pair first. Calling the agent check without a config file, causes it to crash.
+      // account_present = await c();
+      // if (account_present) {
+      //   // Check for an installed Pennsieve agent
+      //   await wait(500);
+      //   [agent_installed_response, agent_version_response] = await check_agent_installed();
+      //   // If no agent is installed, download the latest agent from Github and link to their docs for installation instrucations if needed.
```
CodyCBakerPhD, line 538: Do you think this code will be useful to refactor later on? Otherwise, feel free to remove it instead of just commenting

scripts/others/renderer.js
```diff
@@ -621,6 +719,78 @@ const serverIsLiveStartup = async () => {
...
+  try {
+    responseObject = await client.get("/startup/minimum_api_version");
+  } catch (e) {
+    clientError(e);
+    ipcRenderer.send("track-event", "Error", "Verifying App Version", userErrorMessage(e));
+
+    await Swal.fire({
+      icon: "error",
+      html: `Something went wrong while initializing NWB GUIDE background services. Please try restarting your computer and reinstalling the latest version of SODA. If this issue occurs multiple times, please email <a href='mailto:help@fairdataihub.org'>help@fairdataihub.org</a>.`,
+      heightAuto: false,
```
CodyCBakerPhD, line 741: Same rephrase as above -also, would it be helpful to attach some identifier for this error to distinguish it from the other one?

scripts/others/renderer.js
```diff
@@ -621,6 +719,78 @@ const serverIsLiveStartup = async () => {
...
+    ipcRenderer.send(
+      "track-event",
+      "Error",
+      "Verifying App Version",
+      "Server version does not match client version"
+    );
+
+    await Swal.fire({
+      icon: "error",
+      html: `${appVersion} ${serverAppVersion}The minimum app versions do not match. Please try restarting your computer and reinstalling the latest version of SODA or check to see if a previous version is running in the background with the instructions on our <a href='https://docs.sodaforsparc.io/docs/common-errors/pennsieve-agent-is-already-running' target='_blank'>documentation page.</a> If this issue occurs multiple times, please email <a href='mailto:help@fairdataihub.org'>help@fairdataihub.org</a>.`,
```
CodyCBakerPhD, line 769: So, to clarify, what is a situation that would trigger this error? How would the versions ever mismatch?

scripts/others/renderer.js
```diff
@@ -662,120 +832,120 @@ const check_internet_connection = async (show_notification = true) => {
...
+//     });
+//     log.warn("Pennsieve agent not found");
+//     return [false, userErrorMessage(error)];
+//   }
 
-  if (latest_agent_version != agent_version) {
-    notyf.dismiss(notification);
-    notyf.open({
-      type: "warning",
-      message: "A newer Pennsieve agent was found!",
```
CodyCBakerPhD, line 764: Same question as above - do you think you'll need this later? Or can we just remove?

Specific to the Pennsieve stuff, I don't think we'll need anything analogous to it going forward

pyflask/manageNeuroconv/manage_neuroconv.py
```diff
@@ -0,0 +1,46 @@
...
+
+for category in interfaces_by_category:
+    for name in interfaces_by_category[category]:
+        if (name not in interface_info.keys()): 
+            interface = interfaces_by_category[category][name]
+            interfaces[interface.__name__] = interface
+            interface_info[name] = {
+                'tags': [], 
+                'name': interface.__name__, 
+                'category': os.path.basename(os.path.dirname(os.path.dirname(inspect.getfile(interface))))
```
CodyCBakerPhD, line 17:
```suggestion
                'category': category,
```

pyflask/manageNeuroconv/manage_neuroconv.py
```diff
@@ -0,0 +1,46 @@
...
+from neuroconv import NWBConverter
+import inspect
+import os
+
+interfaces = {}
+interface_info = {}
+
+for category in interfaces_by_category:
+    for name in interfaces_by_category[category]:
+        if (name not in interface_info.keys()):
```
CodyCBakerPhD, line 11: A few things here to note about Python and Pythonic practices - I hope to learn an equal amount from you regarding JavaScript practices 😀

1) Parentheses not necessary: `if (name not in interface_info.keys())` -> `if name not in interface_info.keys()`
2) `.keys()` not necessary, querying `in` on a dictionary acts on the keys: `name not in interface_info.keys()` -> `name not in interface_info`
3) Early return/skip conditions are preferred to reduce [cyclomatic complexity](https://en.wikipedia.org/wiki/Cyclomatic_complexity) by avoiding multiple nesting levels of `if` statements:

```python
if name not in interface_info: 
    # do something
```

->
```python
if name in interface_info:
    continue  # `continue` is a special built-in you can use inside an iterable that causes it to skip to the next iteration
# do something, but you'd only reach this point if you didn't skip
```

Also, just thinking through the logic, what is a case where this would ever be true? (I can think of cases right now where it actually is)

And in those cases when it is true, is skipping really the correct action to resolve the conflicting key (thereby deferring to accepting the 'interface info' of whatever the first found case was)?

pyflask/manageNeuroconv/manage_neuroconv.py
```diff
@@ -0,0 +1,46 @@
...
+            interfaces[interface.__name__] = interface
+            interface_info[name] = {
+                'tags': [], 
+                'name': interface.__name__, 
+                'category': os.path.basename(os.path.dirname(os.path.dirname(inspect.getfile(interface))))
+            }
+
+        info = interface_info[name]
+        # if (info.category != category): info['tags'].append(category)
+        info['tags'].append(category)
```
CodyCBakerPhD, line 22: `category` and `tags` are both outer levels of your `info` schema - so isn't including the category as a tag unnecessary duplication? What are some other things you would include in `tags`?

pyflask/manageNeuroconv/manage_neuroconv.py
```diff
@@ -0,0 +1,46 @@
...
+        
+def get_all_interfaces():
+    """
+    Function used to get the list of all interfaces
+    """
+
+    return interface_info
```
CodyCBakerPhD, line 29: Global variables tend to be frowned upon when not necessary - it might seem a bit wasteful computationally, but I'd recommend moving lines 9-22 into this function so that `interface_info` is something only this function evaluates and returns

pyflask/manageNeuroconv/manage_neuroconv.py
```diff
@@ -0,0 +1,46 @@
...
+                'tags': [], 
+                'name': interface.__name__, 
+                'category': os.path.basename(os.path.dirname(os.path.dirname(inspect.getfile(interface))))
+            }
+
+        info = interface_info[name]
+        # if (info.category != category): info['tags'].append(category)
+        info['tags'].append(category)
+        
+def get_all_interfaces():
```
CodyCBakerPhD, line 24:
```suggestion
def get_all_interface_info() -> dict:
```

CodyCBakerPhD: Two things

i) adjusting name of function to better represent what it does (recommend adjust docstring as well since it's no longer returning just a 'list' of interfaces but rather a dictionary-of-dictionaries containing all interface info

ii) Python has this thing called 'annotation typing' and I'll try to suggest it whenever I see it missing - in this case, any time you define a  function you can summarize it's output structure using `->` and then pointing to the expected return type. You can get more sophisticated with it than I'm suggesting here but dict-of-dicts get a little complicated so just summarizing as a `dict` is fine in that case

Note that this is not something explicitly checked or asserted at runtime, or any time really, unless you use certain optional parsers like mypi, pydantic decorators, or what we'd usually do here is JSON validation against the schema for your 'interface_info' dictionary

Instead, this is done for readability by your fellow developers, who can then tell at a glance what this function is expected to return as output

pyflask/manageNeuroconv/manage_neuroconv.py
```diff
@@ -0,0 +1,46 @@
...
+    if (not interface): interface = interface_info.keys() # get schema for all interfaces
+
+    # Single Interface
+    if (isinstance(interface, str)): return interfaces[interface_info[interface]['name']].get_source_schema()
+    
+    # Combine Multiple Interfaces
+    class CustomNWBConverter(NWBConverter):
+        data_interface_classes = dict(zip(interface, map(lambda name: interfaces[interface_info[name]['name']], interface)))
+
+    return CustomNWBConverter.get_source_schema()
```
CodyCBakerPhD, line 45: This is already quite enough for a single introductory PR, so I'd recommend saving implementation of this functionality for a subsequent PR where we can really dig into and debug the way we want the source schema to be rendered

pyflask/manageNeuroconv/manage_neuroconv.py
garrettmflynn: Since the keys provided by each sub-object of interfaces_by_categories are not the same as their entries elsewhere, I'd argue that an `interface_info` global object is necessary to provide a lookup table mapping the entries in `interfaces_by_categories` to the `interfaces` global object. See the usage of interfaces_info in the `get_schema` function on lines 26, 39, and 43.

Alternatively, `interface_info` could hold the interface object directly, removing the need for `interfaces` at all—but the new `interface` key (a class) would need to be removed before being sent to the client since it is not serializable. But this still leaves us with a global object. Unless we decide to provide an object with equivalent functionality directly from `neuroconv`

pyflask/apis/apiNeuroconv.py
```diff
@@ -0,0 +1,52 @@
...
+class Schema(Resource):
+  @api.doc(responses={200: 'Success', 400: 'Bad Request', 500: "Internal server error"})
+  def get(self, interface):
+
+    try:
+      return get_schema(interface)
+    except Exception as e:
+        if notBadRequestException(e):
+            api.abort(500, str(e))
+        raise e
```
CodyCBakerPhD, line 38: Again, my current recommendation is that we start with the 'AllInterfaces' fetching first and save this schema stuff for a follow-up

One key change that we will look towards in that follow-up PR is:

a) the user selections a series of interfaces via the selection menu
b) all of these interfaces get combined into a single converter
c) the 'source schema' we have been referring to is the result of calling `.get_source_schema()` on that converter class - it already handles integration of all the sub-schemas from each interface, so that complexity can simply be handed off to the backend - all the manager/api needs to do is ask the user to pick the interfaces and then NeuroConv can assemble them

pyflask/manageNeuroconv/manage_neuroconv.py
garrettmflynn: Currently, there's no way to distinguish levels from the `interfaces_by_category` export (e.g. ephys -> sorting) on the client level aside from making assumptions about the received message. As such, I've prioritized, in this example, the "ephys" tag as a category by reference to the related DataInterface's path in the filesystem. 

We can avoid duplicating the tag described as a category (see comment on line 21), I'd just experienced some errors when trying to implement this and the current repo setup isn't conducive to Python debugging as a newbie. It just breaks without error on an Electron subprocess.

Since the `interfaces_by_category` object was exported with multiple category assignments (i.e. an interface can be both in ephys and sorting), so I've implemented this with an assumption that more categories could be added to an interface moving forward. Is this accurate?

If we wanted our categorization scheme to be even more explicit from `neuroconv`, we could nest sub-categories as additional dictionaries, rather than flattening that hierarchy:
```python
interfaces_by_category = {
    "ephys": {
        "sorting": {
                 interface_name: DataInterface
        }
    }
}
```

pyflask/manageNeuroconv/manage_neuroconv.py
garrettmflynn: Since individual interfaces are assigned to multiple categories (e.g. ephys and sorting), this if statement was added to populate an entry in `interface_info` if we hadn't seen the interface before. Otherwise, if this is the second category containing the interface, we simply add the category name to the pre-existing `tags` list.

To answer your question directly, it wouldn't be the correct action to skip the conflicting key since we have to add the category to the `tags` list on **either** condition, whether the statement evaluates to true or false.

The statement is constructed how it is because **there is a unique behavior when the name is not in interface_info already**, though the code that follows should still be run on each pass. Would (3) still be a required change?

package.json
garrettmflynn: I'd removed some things to clean up this PR—and forgot to remove these dependencies.

`vite` is the proposed dev tool to build the frontend and `npm-run-all` allows for running this dev server and the Flask server concurrently with a simple command. 

For this PR, these additional dependencies can be removed.

sections/guided_mode/guided_curate_dataset.html
garrettmflynn: Can do. Thank you for the feedback!

tools/anaconda-env/environment-MAC.yml
garrettmflynn: Thanks for suggesting those changes. I appreciate that you caught this on a fresh install.

yarn.lock
garrettmflynn: Yarn and NPM are interchangeable package managers that ultimately refer to the package.json file. I use `yarn` habitually, so this just snuck through since it wasn't included in the .gitignore.

This is equivalent to the package-lock.json already included in the repo. It tracks all the packages that this repo depends on and the specific versions installed on your machine so, like you mentioned during our call, we don't lapse into dependency hell when we update one of them.

Aside from those noted above, all the dependencies were pre-configured in SODA. Whether they are really necessary is a separate question we can answer later.

scripts/others/renderer.js
garrettmflynn: There are three places where sodaIsConnected is used. But they can similarly be updated.

scripts/others/renderer.js
garrettmflynn: Okay I'll just remove it. I'd originally wanted to be conservative about the changes so I could selectively uncomment (rather than undo several times or search out the snippet in SODA) to bring anything back that broke.

scripts/others/renderer.js
garrettmflynn: I think the SODA team intended for them to be equivalent—so I'd say we don't need an ID. They just copy-pasted and changed one key in the configuration object. 

This is going to be purged anyways if we aren't going to check for the API version on startup.

package.json
CodyCBakerPhD: Sounds good to split those two aspects for debugging - just be sure to have some nice documentation describing how to use them when you do decide to include these

CodyCBakerPhD:
Also, I tried running the app on this branch (in a freshly installed environment from the new environment files too) but I still get stuck on the infinite loading wheel - is this PR intended to be functional at this point?

scripts/others/renderer.js
garrettmflynn: They really shouldn't if you're packaging both with Electron. This is SODA talking, likely something they added for their dev workflow.

scripts/others/renderer.js
garrettmflynn: Agreed. Was just going to do all of this in a separate PR.

pyflask/manageNeuroconv/manage_neuroconv.py
garrettmflynn: See comment below. This is not accurate since we'd like, for instance, "ephys" to contain "sorting" and not vice-versa.

This was simply one way to achieve that without updating `neuroconv`, though we may wish to do so for additional clarity.

CodyCBakerPhD:
@garrettmflynn BTW, when accepting multiple PR 'suggestions', you can add them to a batch from the files tab so they all get added in a single commit instead of several individual commits

image

Note a huge deal since we don't have much of a CI right now, and will probably never be very heavy when we do, just figured I'd let you know about that feature

pyflask/manageNeuroconv/manage_neuroconv.py
garrettmflynn: Makes sense. Thank you for the heads up!

pyflask/apis/apiNeuroconv.py
```diff
@@ -0,0 +1,52 @@
...
+@api.route('/')
+class AllInterfaces(Resource):
+  @api.doc(responses={200: 'Success', 400: 'Bad Request', 500: "Internal server error"})
+  def get(self):
+
+    try:
+      return get_all_interfaces()
+    except Exception as e:
+        if notBadRequestException(e):
+            api.abort(500, str(e))
```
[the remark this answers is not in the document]
garrettmflynn: I agree wholeheartedly. This is just SODA boilerplate—but I'll keep this in mind as we refactor their backend.

yarn.lock
CodyCBakerPhD: Oh my, the package-lock.json is also huge, lol

So... just to clarify that I understand, this is more than just a record of what the current dependencies are, but it is actually a specification for an entire frozen environment (including tertiary dependencies and their versions?)

With Python we do occasionally do this for projects that we want to freeze in time, `pip freeze > frozen_dependencies.txt` will create such a file, if I understand the analogy correctly

pyflask/apis/apiNeuroconv.py
garrettmflynn: Sounds good. How would you recommend separating these things into separate PRs now that we're here?

(1), (2), and (3) are already handled and this flow was the intended focus of this PR. If you go to the second page of Guided Mode, you can input your selection on a dummy interface.

The only difference we may want from what's implemented now and what you've described is that **single interface requests are currently returning `.get_source_schema()` from the interface that was requested—not a converter with a single interface specified. Not sure which we'd prefer.

scripts/others/renderer.js
CodyCBakerPhD: OK planned for removal in follow-up then

scripts/others/renderer.js
CodyCBakerPhD: OK planned for removal in follow-up then

garrettmflynn:
Thanks for the heads-up about the batch commit! 

> Also, I tried running the app on this branch (in a freshly installed environment from the new environment files too) but I still get stuck on the infinite loading wheel - is this PR intended to be functional at this point?

The repo is intended to be functional. I'll try to get everything running myself on a fresh install.

yarn.lock
garrettmflynn: Exactly. This is just handled automatically by the package manager every install because web dependencies are a nightmare to deal with.

package.json
garrettmflynn: Can do!

pyflask/manageNeuroconv/manage_neuroconv.py
garrettmflynn: Makes sense. As noted below, I've already implemented a basic demo that relies on both API calls. Do you want me to remove this all from the current PR?

scripts/others/renderer.js
garrettmflynn: This will also be easier to deal with in a follow-up, as I'm not sure we'll need this server check. Since I've commented out the code that accesses user accounts from the server, it's not being used for anything substantive anyways.

pyflask/manageNeuroconv/manage_neuroconv.py
CodyCBakerPhD: > Since individual interfaces are assigned to multiple categories (e.g. ephys and sorting)

Well, not that's not supposed to be true but reflects a misunderstanding in how we want these things to be described. That outer level is a modality (icephys, ecephys, ophys, etc.) and the inner level could be a 'category' (maybe more appropriate to refer to as a 'technique')

Given how complex the interface arrangement is becoming, I strongly suggest we simplify as much as possible here and then revisit this interface info structure in a follow-up PR focused purely on that (and how NeuroConv can help with it)

Also our minimum viable product is purely to focus on the SpikeGLX + Phy interfaces to start with, so would you be OK with vastly simplifying this for now to simply use those two interfaces until we can discuss specifically how to deal with the entire interface gallery?

garrettmflynn: These two levels of categorization make much more sense. 

I'd be happy to break this into a follow-up PR. How would you propose simplifying this PR for now?

pyflask/manageNeuroconv/manage_neuroconv.py
CodyCBakerPhD: > so I've implemented this with an assumption that more categories could be added to an interface moving forward. Is this accurate?

Not really, see above comment about modality distinction.

If what you want is to define a JSON schema for an 'information' object used to describe/classify each interface class, then I think we should discuss the design and requirements for that first in a GitHub issue, then we can adapt NeuroConv as necessary and follow-up with a much simpler implementation in a subsequent PR - how does that sound?

pyflask/manageNeuroconv/manage_neuroconv.py
CodyCBakerPhD: > See the usage of interfaces_info in the get_schema function on lines 26, 39, and 43.

Yes, what I'm saying is those references can all be replaced by calls to the retrieval function that returns a dictionary instead of referring to the global dictionary itself


> but the new interface key (a class) would need to be removed before being sent to the client since it is not serializable.

We can make the interfaces serializable if needed, it's simply not something we've had to think of before


> But this still leaves us with a global object.

I still don't see why that's needed, all the API functions look like they import the retrieval functions from the manager and then call those functions return their output, so any operations on the information structure could be done within local scopes

pyflask/apis/apiNeuroconv.py
CodyCBakerPhD: > (1), (2), and (3) are already handled and this flow was the intended focus of this PR.

Hmm OK I'll save judgement until I can see it in practice then, but I'm still inclined to start small with a feature like that.

This is also a good case of that PR branching we talked about the other day, which I see you've already put into practice in #1.

This PR could be split into at least 2 separate branches - the first base branch could focus on fixing the dependencies, removing unused fluff from `renderer.js`, and deciding what to do with that massive yarn file - the second would add the NeuroConv manager, api, and specific front-end elements for those displays

That would make it easier to iterate on the design for the main NeuroConv API part - as it stands this PR is actually doing a lot of different things all at once which makes it harder to untangle

CodyCBakerPhD:
> The repo is intended to be functional.

OK, good. Given that the base SODA (+/- my bug-inducing Pennsieve disabling) was working, and the philosophy here is to start from something that works and remove everything we don't need, it stands to reason that it should still be working every step of the way

Just let me know when to try again

garrettmflynn:
@CodyCBakerPhD You should be good to try again! Just succeeded in a clean Python environment on my Mac.

At least when I got it back, it looks like it was broken because changing `get_all_interfaces` to `get_all_interface_info` had downstream consequences for imports of that function. While debugging, I enabled printing errors on the Python child_process to the terminal so we're no longer blind to the server.

pyflask/manageNeuroconv/manage_neuroconv.py
garrettmflynn: > I still don't see why that's needed, all the API functions look like they import the retrieval functions from the manager and then call those functions return their output, so any operations on the information structure could be done within local scopes

Maybe this is where we diverge. In my mind, there are no "operations" being done on the information structure—at least in the sense of write operations—since it's constructed at the top of the file and otherwise unmodified. Just like the dictionary in `neuroconv`, both `interfaces` and `interface_info` are simply maps to static information that the client would like to read. 

Constructing a reference object like that, I assumed, would only be done once to avoid regenerating the same thing with no need to maintain independent references.

If this is not properly pythonic, though, I'm happy to adjust. It'd be pretty trivial to encapsulate the globals in a generator function.

pyflask/manageNeuroconv/manage_neuroconv.py
garrettmflynn: Sounds good to me.

yarn.lock
garrettmflynn: It isn't necessary to have both a yarn.lock and package-lock.json, so I've just removed the former.

pyflask/apis/apiNeuroconv.py
garrettmflynn: Now that I've already made this PR, is there a good way to split off changes in that way?

CodyCBakerPhD:
> You should be good to try again!

To confirm, this is how I should be launching it on the latest changes for this branch? Still stalling out for me if so

```
> npm start

> nwb-guide@0.0.1 start
> NODE_ENV=development electron .


DNT disabled
App Status: dev
Killing all previous processes
child process success on port 4242
(node:25712) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
(Use `electron --trace-deprecation ...` to show where the warning was created)
Running pre-checks

# just stalls out on the spinning wheel here
```

I even tried `npm install` before that, if it changes anything

garrettmflynn:
You’re running all this with the Python env active? That is the correct NPM
command

CodyCBakerPhD:
Oh apparently it did timeout eventually


Full sequence of operations:

a) launch conda shell
b) `conda create -f tools/

CodyCBakerPhD:
Oh apparently it did timeout eventually

image

error in the shell: `[11092:0304/152213.110:ERROR:gpu_init.cc(481)] Passthrough is not supported, GL is disabled, ANGLE is`

Full sequence of operations:

a) launch conda shell
b) `conda create -f tools/anaconda-env/environment-Windows.yml`
c) `conda activate env-electron-python`
d) `npm start`

pyflask/apis/apiNeuroconv.py
CodyCBakerPhD: When using git CLI, I would do this with `git checkout -b <name of new branch>` while originating from this branch (`api`), then checkout back to `api` and remove the excess code

Now, if opening a PR for the new branch, you would simply point toward `api` as the base and keep it up to date (might have to resolve a conflict involving the lines removed, but that's a straightforward one to just accept the changes on the new branch)

garrettmflynn:
Working on a clean install on my PC.

This might not be well-formed but, when you say “remove the excess code”,
won’t it be difficult to isolate changes on branch 1 and branch 2 since
branch 1 would need to revert all the changes we want to keep isolated on
branch 2?

garrettmflynn:
Hmm it worked for me on my PC on the second startup—albeit after running
the following command.

If you continue facing issues, try running ‘npm run dev:server’ to run the
Flask server in isolation and check for errors in the console. Visit
http://localhost:4242 to view the API endpoints on a generated Swagger page.

I’m going to be afk for the rest of the day—but I can clarify things over
email intermittently.

CodyCBakerPhD:
If it helps at all, I just discovered the 'Toggle Developer Tools' which seems to give more information

image

Just repeats that connection refused over and over until timeout

> This might not be well-formed but, when you say “remove the excess code”,
won’t it be difficult to isolate changes on branch 1 and branch 2

The only times I encountered difficulty with this is if you explicitly use `git revert` to mess with the commit history; what I'm suggesting is adding commits to branch 1 that remove some number of lines, which puts its history beyond the splinter point of branch 2, so then for branch 2 to catch up it would need to update from that and add additional commit replacing the removal with the addition. If the redundancy there bothers you we can always rebase or squash, but computationally the history inflation really is not significant in the grand scheme of things

I'd be happy to do it for you if you want, it's all about simplifying the changelogs at the end of the day

CodyCBakerPhD:
> If you continue facing issues, try running ‘npm run dev:server’

OK, that looks like

image

and console

image

What am I supposed to click on to emulate the behavior of the app through that interface?

garrettmflynn:
That looks good. It should’ve already errored if there was something wrong.

You’d type the path to the request in the address bar (e.g.
https://localhost:4242/neuroconv)

Have you tried booting up the Electron app after this? Thanks for the
Developer Tools output, it looks pretty standard. It’s just sending a test
request to the Python server.

BTW in the [SODA for SPARC Project Setup page](
https://docs.sodaforsparc.io/docs/developer-documentation/project-setup), it
looks like there’s a caveat for Windows:

> For Windows, you also need the win-node-env package. This should
automatically have been installed as an optional dependency from the
package.json but if it doesn't you can use the following command to add it
to your project:

npm install -g win-node-env

Maybe you could try this? Everything looks like it’s working properly
otherwise.

CodyCBakerPhD:
I tried the following commands on my Mac as well and same thing happened

From a fresh install of conda,

```
conda env create -f tools/anaconda-envs/environment-MAC.yml  # with correction on other PR
conda activate env-electron-python
npm install  # trying to do npm start right away errors our
npm start

# Stuck on spinning wheel, dev tools console says same thing as windows
```

garrettmflynn:
Hmm one more idea:
1. Delete the node_modules folder
2. Try NPM install again
3. Test
4. If this doesn’t work, delete the package-lock.json and repeat steps 1-3

If this doesn’t work, maybe we should just schedule another time to hop on
a call?

Could also try installing from another MacBook (2015) at home before we
meet up.

CodyCBakerPhD:
Back to windows

> Have you tried booting up the Electron app after this? 

After booting the dev server with `npm run dev:server`, then `npm start` (if that's what you mean), still stalls out

> npm install -g win-node-env

Did not fix the issue

I'll try the node modules thing on Mac now

CodyCBakerPhD:
On Mac,

> Hmm one more idea:
1. Delete the node_modules folder
2. Try NPM install again
3. Test
4. If this doesn’t work, delete the package-lock.json and repeat steps 1-3

Same issue after following these steps


> If this doesn’t work, maybe we should just schedule another time to hop on
a call?

You can try grabbing a time on my calendly, but I'm pretty full on meetings Mon/Tues and away at conference thereafter

CodyCBakerPhD:
OK, ok

Did a full clean and re-installation of conda, git, and a fresh clone of the repo

Fresh round of conda environment and npm installations

App seems to work fine now, not really sure what was wrong before but there was clearly something cached somewhere somehow that it did not like

Made it as far as here

image


But trying to press the 'get_schema' button sent me back to the main page and the console spit out all kinds of stuff, the last of which being that `ERROR:gpu_init.cc(481)] Passthrough is not supported, GL is disabled, ANGLE is` thing

So I see what you're doing with the interface listing, but to simplify this PR would you be OK if we stop at building the nesting structure, with a narrowed space of possibilities in each box for the time being?

What I'm imagining for this first PR is simply

```
ecephys
    recording
        SpikeGLX
    sorting
        Phy
```

and maybe to prove that the front-end to back-end integration is working, after selecting which combo of those two interfaces we want, the next page (the page loaded after hitting 'Next' after selecting the dropboxes) could just print simple HTML text like 'you have selected XYZ interfaces'

That way the complexity of the source schema stuff can go into a follow-up PR

And then also, we can expand to other modalities and their techniques step-by-step in follow ups once the rest of the pipeline is working, which should make debugging a lot easier to manage

How does that sound?

garrettmflynn:
Awesome! Glad you got it working.

Sounds good to me. I’ll hardcode what we need for SpikeGLX and Phy on the
backend so the same series of requests is made.

This will also simplify some of the redesign so we can focus on only a
handful of pages.

garrettmflynn:
> I'd be happy to do it for you if you want, it's all about simplifying the changelogs at the end of the day

@CodyCBakerPhD Just so I have a model of this for the future, it would be fantastic if you could do this. Let me know if you have enough time this week—otherwise, I'm happy to take a stab at it. Generally, I'd see the following branches created: 

1. api: Just the Flask endpoint constrained to Phy and SpikeGLX
2. api-frontend-test: The changes to the frontend that generate the multi-select list
3. soda-purge: Changes to renderer.js and HTML files
4. api-tests: Same as current, but additional constraints added to limit to Phy and SpikeGLX

Do you agree? I attempted to start the process myself, but I'm still not confident about how to untangle mixed commits that contain changes I'd like to remove from api and, for instance, split selectively into branch 2 and branch 3

garrettmflynn:
> But trying to press the 'get_schema' button sent me back to the main page and the console spit out all kinds of stuff, the last of which being that ERROR:gpu_init.cc(481)] Passthrough is not supported, GL is disabled, ANGLE is thing

I also noticed this. The first time that request is sent to the Flask server, it appears that the command to restart the server is also triggered—which is an unexpected behavior. Navigating back to the multi-select page and pressing the button will result in the correct behavior: A printout of the combined schema in the Developer Tools. The gpu_init.cc error appears unrelated.

I'm sure this will be cleaned up once we remove all the unnecessary SODA behaviors.

CodyCBakerPhD:
> api: Just the Flask endpoint constrained to Phy and SpikeGLX
api-frontend-test: The changes to the frontend that generate the multi-select list
soda-purge: Changes to renderer.js and HTML files
api-tests: Same as current, but additional constraints added to limit to Phy and SpikeGLX
> 
> Do you agree?

Yep, exactly. Sounds like a great chain of modular PRs!


I'll take a stab at the splitting now

CodyCBakerPhD:
Putting this into draft to serve as a record of intended additions once other modular PRs are merged over time

Issue #4, opened 2023-03-04
Add pre-commit


CodyCBakerPhD:
Specific to Python/flask components; both black and isort. Might be nice to start off the project with pydocstyle (numpy settings) as well since adding that later is always a pain if it hasn't been adhered to

CodyCBakerPhD:
Pretty much done, except for docstyle

Issue #3, opened 2023-03-04
Add Codespell


CodyCBakerPhD:
Wait until main content purge is complete to avoid false flags

garrettmflynn:
I've left most of their files untouched in the repo for reference—so it's all largely there.

Would you like me to remove everything unused from the main SODA app and consolidate what is still active into `src`?

CodyCBakerPhD:
I'm not sure, what do you think?

garrettmflynn:
I think it'd be valuable to remove everything we aren't using. This would force me to visit the current SODA app to bring anything we want in explicitly. Does that sound okay?

CodyCBakerPhD:
I think at this point, that sounds fine

garrettmflynn:
Great I've just opened PR #73 for this

PR #5, opened 2023-03-04
Clean workflows

CodyCBakerPhD:
Removing some unnecessary workflows

Will replace the misspelling one with Codespell once repo is a little cleaner #3 

Will replace their custom black with pre-commit once the main purges are complete to minimize conflicts #4 

Keeping two for now - the Chromatic (which uses the storybook? not quite sure what tests it runs though) and the CodeQL (which has something to do with security vulnerabilities, so want to confirm with @garrettmflynn if we need it or not)


CodyCBakerPhD:
Also, I wonder, does 'prettier' help you at all on the JS side @garrettmflynn? Is it just a linter or do you know if it's capable of auto-fixing style changes? I noticed there is a pre-commit for it: https://prettier.io/docs/en/precommit.html

garrettmflynn:
I don't generally use Prettier myself, but it's both a linter and automatically updates code style on save. I'm agnostic to keeping it—though it could help with consistent styling.

As for Chromatic, I've never used it before. If you'd like to review frontend component changes easily outside of the context of the app itself, it might be worth keeping.

CodyCBakerPhD:
> As for Chromatic, I've never used it before. If you'd like to review frontend component changes easily outside of the context of the app itself, it might be worth keeping.

Hmmm do you have examples/instructions on how that would actually be used in practice? 

If you're unsure about it or don't think it's worth it for now, we can just remove it for simplicity and then look into adding it back once we learn more about it


> I don't generally use Prettier myself, but it's both a linter and automatically updates code style on save. I'm agnostic to keeping it—though it could help with consistent styling.

OK good to know - we are all about consistent styling here at CatalystNeuro so we'll want to keep it. Though I'm going to look into including it in the pre-commit bot instead of this mysterious SODA bot that I don't think I ever gave permissions for lol

garrettmflynn:
Sounds good!

> Hmmm do you have examples/instructions on how that would actually be used in practice?
Honestly, I don't think Chromatic is worth it right now. Once we get down to finalizing the look and feel of the application, this'll become much more important. But first, we have to actually get something running in Storybook to compare changes...and I'd assume we're more focused on functionality and user flow than how much padding we'd like a section to have at the moment.

See this [demo video from the Chromatic team](https://www.youtube.com/watch?v=9o6uB1X-LZ8&embeds_euri=https%3A%2F%2Fwww.chromatic.com%2F&source_ve_path=MjM4NTE&feature=emb_title&ab_channel=Chromatic) to see how it works in action.

garrettmflynn (approved):
Looks good to merge! This should help keep our code style consistent.

PR #6, opened 2023-03-04
Ignore yarn.lock in git

CodyCBakerPhD:
@garrettmflynn I think you recommended these be improved in the .gitignore?

I also noticed my own `package-lock.json` changed by itself after I tried running the app


garrettmflynn (changes requested):
We'll want to keep the package-lock.json, as noted in the latest comment

CodyCBakerPhD:
> For automated testing, we'd generally want to use 'npm ci' for a deterministic build straight from package-lock.json.

What exactly does `npm ci` do?

Is it separate from the backend tests run via `pytest pyflask/tests`?

CodyCBakerPhD:
> We'll want to keep the package-lock.json, as noted in the latest comment

And that even applies to the `resolved` and `integrity` fields like those seen in https://github.com/catalystneuro/nwb-guide/pull/10/files#diff-053150b640a7ce75eff69d1a22cae7f0f94ad64ce9a855db544dda0929316519L14248-R14251? I figured those were system specific

CodyCBakerPhD:
Just for my own education, could you summarize the fields of one of these, using

```
"node_modules/npm/node_modules/aggregate-error": {
      "version": "3.1.0",
      "resolved": "https://registry.npmjs.org/aggregate-error/-/aggregate-error-3.1.0.tgz",
      "integrity": "sha512-4I7Td01quW/RpocfNayFdFVk1qSuoh0E7JrbRJ16nH01HhKFQ88INq9Sd+nd72zqRySlr9BmDA8xlEJ6vJMrYA==",
      "dev": true,
      "inBundle": true,
      "license": "MIT",
      "peer": true,
      "dependencies": {
        "clean-stack": "^2.0.0",
        "indent-string": "^4.0.0"
      },
      "engines": {
        "node": ">=8"
      }
```

as an example?

I see the 'version' field being the most important, maybe the `dependencies`/`engines` as well. That's similar to how we setup requirements in Python anyhow

CodyCBakerPhD:
One final thing I'll note and then I'll head away for a while:

Just because it's in the `.gitignore` doesn't mean you can add changes to it

In particular, you can still do `git add package-lock.json` and you can commit/push it like normal

What the `.gitignore` does is blocks is from being included automatically by things like `git add *` or other catch-alls. I assume it also prevents it from showing up on the 'files changed' display for GitHub GUIs

garrettmflynn:
> What exactly does npm ci do?

It gathers the dependencies directly from the package-lock.json and is generally for closed-integration tools. See the [official docs](https://docs.npmjs.com/cli/v9/commands/npm-ci) for more details.

> Is it separate from the backend tests run via pytest pyflask/tests?
Yes, `npm ci` is because it simply installs Node dependencies. Though we might have a CI / CD pipeline that would run both Node and Python tests using a different command (e.g `npm test` where both suites are run)

garrettmflynn:
> What the .gitignore does is blocks is from being included automatically by things like git add * or other catch-alls. I assume it also prevents it from showing up on the 'files changed' display for GitHub GUIs.

Got it, I see what you mean. Since I use Github Desktop, I'm used to interpreting "add to .gitignore" to mean "ignore for commits" since it never shows up in the change list. I'm happy with whatever workflow you'd prefer.

garrettmflynn:
> Just for my own education, could you summarize the fields of one of these
> ```json
"node_modules/npm/node_modules/aggregate-error": {
      "version": "3.1.0",
      "resolved": "https://registry.npmjs.org/aggregate-error/-/aggregate-error-3.1.0.tgz",
      "integrity": "sha512-4I7Td01quW/RpocfNayFdFVk1qSuoh0E7JrbRJ16nH01HhKFQ88INq9Sd+nd72zqRySlr9BmDA8xlEJ6vJMrYA==",
      "dev": true,
      "inBundle": true,
      "license": "MIT",
      "peer": true,
      "dependencies": {
        "clean-stack": "^2.0.0",
        "indent-string": "^4.0.0"
      },
      "engines": {
        "node": ">=8"
      }
```

Can do. Since these are all auto-generated, I've honestly never taken a deeper look at this—so bear in mind that the following is just my interpretation of the [official docs] using this as a particular case.

1. **Version:** This is the version number as resolved from the package.json (which may include loose specifiers like ^3.1.0 or ~3.1.0). This specific entry is a dependency of a package noted explicitly there, which is why it isn't present if you ctrl-f there.
2. **Resolved:** This is where the assets were downloaded from
3. **Integrity:** This is a generated string to check for the integrity of the asset.
4. **Dev:** This is true because the package is specifically a part of dev dependencies (e.g. for build tools, etc.)
5. **inBundle:** This notes that this package is a bundled dependency. Most libraries are bundled at present, which places all the code into a single file rather than keeping the original file structure of the origin repo.
6. **License:** As you'd expect, this is the licence the dependency is distributed under.
7. **Peer:** This is undocumented in the official docs, but I'd assume this indicates whether this dependency is a peer dependency (i.e. this dependency only works when plugged into another of dependency, like if you publish the DataInterface for neuroconv independently and would like to note that it only works with neuroconv)
8. **Dependencies:** Direct from this dependencies `package.json` file. Notes what other packages this depends on and their version specifiers.
9. **Engines:** Direct from this dependencies `package.json` file. This is an optional field that commonly specifies what version of Node is required to run this package, as new features may be absent from old versions.

CodyCBakerPhD:
Thanks for the summary!

> Since these are all auto-generated

Now see, that's what worries me about having this in the central repository at all then - if this would be generated fresh (and different) by every dev that ever runs this on their git clone, then this is exactly the kind of file that should be in a `.gitignore` and never included packaged with the project.

A final clarification of that and we can put this to bed - is the `package-lock.json` simply a log of what dependencies were installed in the `npm` environment at that moment in time (Python equivalent, `pip freeze > current_environment.txt` on every installation), or is this file necessary for telling `npm` exactly what set of dependencies to install (Python equivalent, `pip install -r requirements.txt`)?

garrettmflynn:
The answer to your questions is slightly complicated since essentially `package-lock.json` could be seen more as a new sort of `current_environment-lock.txt` file that tracks the resolution of conditional requirements and adds metadata.

> is the package-lock.json simply a log of what dependencies were installed in the npm environment at that moment in time (Python equivalent, pip freeze > current_environment.txt on every installation)

I'd say that's what it aspires to be—though it's a bit more complicated since it's auto-generated from `package.json`. Since you can include conditional requirements in `current_environment.txt`, though, I'd say that this file is more like the `package.json` file. 

> or is this file necessary for telling npm exactly what set of dependencies to install (Python equivalent, pip install -r requirements.txt)?

As mentioned above, `package.json` is the standard for specifying dependencies and is used for `npm install`, but these are commonly loose specifications. So `package-lock.json` is not strictly necessary. `npm ci` is the only command that directly uses `package-lock.json`—and these are the exact versions that were installed. 

Consequently, most recommend including `package-lock.json` in the repo and keeping it out of `.gitignore`, as long as some changes to the developer process are met: https://stackoverflow.com/questions/48524417/should-the-package-lock-json-file-be-added-to-gitignore. If we use `npm ci` instead of `npm install`, we won't update the file when installing pre-configured dependencies. And if we use the `npm install dependency` form, the `package-lock.json` will be updated (as it should) to account for the new dependency.

CodyCBakerPhD:
> npm ci is the only command that directly uses package-lock.json—and these are the exact versions that were installed.

Aha! That's what I was looking for.

So the way we often do this is we have 'testing-specific' requirements that the continuous integration pins to.

But if the suggestion of you and

>  https://stackoverflow.com/questions/48524417/should-the-package-lock-json-file-be-added-to-gitignore

is to use `npm ci` instead to replace `npm install`, that's also perfectly fine since it won't be automatically changing the values in that file every time.

Thanks for the stackoverflow link, that was really helpful

garrettmflynn (approved):
Perfect. We'll still have version control while limiting the possibility of needless redundancy.

PR #7, opened 2023-03-04
Update environment-MAC.yml

CodyCBakerPhD:
@garrettmflynn Had to update this when testing on my Mac (M1), no distribution for `2.20.1` available with that architecture


garrettmflynn (approved):
Looks good! Good catch for the M1 processor.

PR #9, opened 2023-03-06
First NeuroConv endpoint and Interface display

CodyCBakerPhD:
@garrettmflynn OK this is a fresh PR from main that introduces fresh commits including files changed from the `api` branch as they pertained to the first establishment of the NeuroConv API + the demonstration of that using the interface display form

Fresh commits are really the best way to preserve a linear history rather than picking and choosing individual commits to prune from the other branch. I know it takes time and probably doesn't feel as productive, but that goes away once we get into the practice of always doing this super-modularly from the start

I tried not to include anything that wasn't directly related to that, but my ability to navigate what is 'necessary' or not in the `renderer.js` may have left something out here - can you help put the finishing touches on making this branch functional for the narrowed scope of what we're trying to achieve?


garrettmflynn:
Thanks for completing this. I see what I should do if this happens again—and how to avoid a monolithic PR like #2 in the first place. 

Just finished updating the branch so that it's functional. Now the multi-select list populates with the (slightly updated) data structure received from the neuroconv endpoint.

garrettmflynn (dismissed):
Seems to be working on the last push. Looks like it's ready to merge.

CodyCBakerPhD:
Lol wow, never seem pre-commit dismiss a review as stale before

CodyCBakerPhD:
Granted, it usually triggers well before anyone would hit the approve button... Just happened here since this is first time the bot turned on

garrettmflynn (approved):
This review isn't stale. Merge on!

PR #11, opened 2023-03-06
Allow for running the Python server independently

garrettmflynn:
To test, you can run `npm run dev:server` and the Flask server should start and log messages on the terminal.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,6 +6,7 @@
   "main": "main.js",
   "scripts": {
     "start": "NODE_ENV=development electron .",
+    "dev:server": "cd pyflask && python -m flask run --port 4242",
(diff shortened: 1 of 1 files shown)


CodyCBakerPhD (approved):

CodyCBakerPhD:
Very interesting how easy that is. Wonder why it wasn't already in SODA?

garrettmflynn:
They also weren't logging outputs from the Flask server when it was run by Electron—so my guess is they might have another way of testing the Python and browser integration that isn't explicit in the repo.

If they're Python devs, they might just know this command by heart. Or they didn't want to mix Node and Python commands in the `package.json` scripts field.

PR #14, opened 2023-03-07
Establish a parallel entrypoint where we can work on adaptation to SODA

garrettmflynn:
This PR introduces a parallel entrypoint for the frontend code used by the SODA application. In particular, it: 
1. Creates a new `guide.html` file that mirrors the original `index.html` 
2. Consolidates the use of JavaScript to a single entrypoint (`src/index.js`) and relinks this in `guide.html`
3. Moves the `assets/imports.js` code into the `src/index.js` file and organizes the imports for improved modularity
4. Replicates the `scripts/others/renderer.js` file and associated dependencies in the `src` directory to ensure that, moving forward, we can make changes to this dependency without changing the original.

As a whole, this PR represents a proof-of-concept for how we can continue adapting the SODA frontend for our uses in NWB GUIDE. The commits of this PR were made in a modular fashion—so it would be quite simple to break out several branches from them if this is advantageous.


garrettmflynn (commented):
Whoops, thought I was on the branch for #18 which needs its package-lock.json updated.

garrettmflynn:
@CodyCBakerPhD You're welcome to set this to a draft. This PR is going to have pretty major changes in large blocks—partly because of necessity (i.e. everything in the SODA app is connected to everything else) and partly because of speed / learning (i.e. going methodically through this the first time would take a lot of time and likely wouldn't have enough context to be done well).

Whenever you're free after the conference, we should discuss the best way to make these changes in modular PRs. It won't be hard now that this is behind us—but it'd be best to gain alignment beforehand.

garrettmflynn:
This PR is no longer needed because of recent progress on #35, #36, and #37 on other PRs.

PR #15, opened 2023-03-08
Add pre-commit configs

CodyCBakerPhD:
I messed up when I added the pre-commit, forgot to attach the config as the `pyproject.toml`


garrettmflynn (approved):
Looks good to me! Just for future reference, is there an easy way to filter out pre-commit changes aside from excluding them using the "Changes From" button in the Github PR view?

CodyCBakerPhD:
> is there an easy way to filter out pre-commit changes aside from excluding them using the "Changes From" button in the Github PR view?

No way to filter the changelog by style-only changes, sadly.

Easiest way to avoid the problem altogether is to set it up in the project first thing, and then not make changes to it like I did here 😄

PR #17, opened 2023-03-08
Added support for Web Components with Lit

garrettmflynn:
This PR introduces a minimal usage of Lit to replace the sidebar component of the NWB GUIDE application.

It encapsulates sidebar-specific HTML, CSS, and JS into a single class that can be used anywhere in the application. 

Notably, using Lit in our current Electron configuration—and likely Electron generally—requires transpiling the source code so that the `lit` package (in ESM) can be read. As such, we had to add a bundler (`vite`) to output a bundled JS file in the `dist` folder that is then referenced by the original Electron app. 

While this additional requirement could be seen as a drawback, it also opens several opportunities (not introduced in this PR):
1. With the addition of a `tsconfig.json` file, we can move forward with writing the frontend in TypeScript for better predictability, readability, etc.
2. With a single additional script specified in package.json, we can use Vite to serve the frontend by itself during development


CodyCBakerPhD:
Shall this count as one of the first 'purges'?

Tried running it, seemed to behave fine for me

garrettmflynn:
I'd say this is more of a feature addition, specifically of Web Components using Lit and Vite.

The general workflow for the "purges" will now go as follows: 
1. Make a Web Component for a desired page and sub-elements
2. Put all JavaScript and CSS it uses into it
3. Replace the original usage in the parallel guide instance

Since these Components will go into the parallel guide instance, we'll be "purging" their unused code without actually adding deletions until we wipe the original SODA instance.

garrettmflynn:
@CodyCBakerPhD Looks like the review was marked as stale again. Whoops!

CodyCBakerPhD:
Yeah, conflict resolution counts as fresh commits - it's all for maximum safety, though, so an annoyance we happily put up with

CodyCBakerPhD (approved):

PR #20, opened 2023-03-08
Add schema endpoint for single schema 

garrettmflynn:
This PR introduces a basic endpoint to request the schema of a single `neuroconv` interface. Further discussion about the returned NWBConverter schema will let us add an endpoint for multiple interfaces.

An additional change was made to interface_info keys to prioritize the class name and ensure uniqueness in the future.

### DIFF
--- a/pyflask/apis/apiNeuroConv.py
+++ b/pyflask/apis/apiNeuroConv.py
@@ -22,11 +22,10 @@ class AllInterfaces(Resource):
                 api.abort(500, str(e))
             raise e
 
+
 @api.route("/schema/<string:interface>")
 class Schema(Resource):
-    @api.doc(
-        responses={200: "Success", 400: "Bad Request", 500: "Internal server error"}
-    )
+    @api.doc(responses={200: "Success", 400: "Bad Request", 500: "Internal server error"})
     def get(self, interface):
 
         try:
--- a/pyflask/apis/apiNeuroConv.py
+++ b/pyflask/apis/apiNeuroConv.py
@@ -1,6 +1,6 @@
 from flask_restx import Namespace, Resource, reqparse
 from namespaces import get_namespace, NamespaceEnum
-from manageNeuroconv import get_all_interface_info
+from manageNeuroconv import get_all_interface_info, get_schema
 from errorHandlers import notBadRequestException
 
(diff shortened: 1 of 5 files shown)


CodyCBakerPhD:
Does this have any presence to demonstrate on the front-end, or purely back-end additions?

garrettmflynn:
I didn't see the benefit of mixing those PRs since REST APIs are pretty reliable and its tested on the Python end.

I'd be happy to move forward with some frontend integration if you'd like. Only issue is that this demo would be best represented on a new page, which isn't straightforward to create yet.

CodyCBakerPhD:
> I didn't see the benefit of mixing those PRs since REST APIs are pretty reliable and its tested on the Python end.

Sure, fair enough. Save front-end for a PR of its own

CodyCBakerPhD:
My only concern then for this PR specifically is that I think PR #21 is a generalization of it - though interfaces can generate their own stand-alone schemas, our typical approach to handling this kind of thing on the backend would simply be to always go the NWBConverter route, even if it only has a single `data_interface_class`

garrettmflynn:
Got it. Would a better solution be to use the same endpoint('/schema') but encapsulate the single interface name in a list and use the generalized function from #21? Or do you think there's no need to provide a simple endpoint for single interfaces (i.e. from `/neuroconv/schema?interfaces=PhySortingInterface` to `/neuroconv/schema/PhySortingInterface`)?

garrettmflynn:
Closing after replacement by #23

CodyCBakerPhD:
> Would a better solution be to use the same endpoint('/schema') but encapsulate the single interface name in a list and use the generalized function from https://github.com/catalystneuro/nwb-guide/pull/21?

Exactly


> /neuroconv/schema?interfaces=PhySortingInterface to /neuroconv/schema/PhySortingInterface)?

I mean, I don't really have an opinion about the particulars of how the API communicates

As long as when the user selects only a single interfaces from the frontend, that becomes a list of a single element for the input into the backend, which always makes a NWBConverter (even one of a single data interface) is all I care about there; feel free to map the endpoint calls (which I think of as REST queries, IDK if that's accurate...) however you think best

garrettmflynn:
That sounds great, you've got the right idea about the backend. It's just a REST API that we're making queries, and I'm referring to the places we're making queries to as API endpoints.

PR #21, opened 2023-03-08
Add endpoint for getting the schema from multiple combined interfaces

garrettmflynn:
This PR introduces a new endpoint and related tests for requesting the schema of multiple combined interfaces.


pyflask/tests/test_neuroconv.py
```diff
@@ -10,3 +25,17 @@ def test_get_all_interfaces(client):
         assert isinstance(value["modality"], str)
         assert isinstance(value["name"], str)
         assert isinstance(value["technique"], str)
+
+
+# Uses the NWBConverter Class to combine multiple interfaces
+def test_multiple_schema_request(client):
+    interfaces = ["SpikeGLXRecordingInterface", "PhySortingInterface"][:1]
```
CodyCBakerPhD, line 32:
```suggestion
    interfaces = ["SpikeGLXRecordingInterface", "PhySortingInterface"]
```

pyflask/tests/test_neuroconv.py
```diff
@@ -1,4 +1,19 @@
 # --------------------- Helper Functions ---------------------
+def check_schema(data):
+    assert isinstance(data, dict)
+    assert isinstance(data["required"], list)
+    assert isinstance(data["properties"], dict)
```
CodyCBakerPhD, line 5: Looks like this could be replaced a bit more rigorously by the `Draft7Validator.check_schema(schema=schema)`, imported by `from jsonschema import Draft7Validator`

pyflask/tests/test_neuroconv.py
```diff
@@ -1,4 +1,19 @@
...
+
+
+def check_converter(data, interfaces):
+    check_schema(data)
+    for interface in interfaces:
+        assert isinstance(data["properties"][interface], dict)
+        check_schema(data)
```
CodyCBakerPhD, line 12:
```suggestion
```

pyflask/tests/test_neuroconv.py
```diff
@@ -1,4 +1,19 @@
...
+    assert isinstance(data["title"], str)
+    assert isinstance(data["description"], str)
+    assert isinstance(data["version"], str)
+
```
CodyCBakerPhD, line 16: I think these are also things that would be checked by the `Draft7Validator`

pyflask/manageNeuroconv/manage_neuroconv.py
```diff
@@ -24,3 +24,26 @@ def get_all_interface_info() -> dict:
                 }
 
     return interface_info
+
+
+def get_combined_schema(interfaces):
```
CodyCBakerPhD, line 29:
```suggestion
def get_combined_schema(interfaces: Optional[List[str]]) -> dict:
```

Again some more to learn about annotation typing; not only can you annotate the expected return type, but with `:` you can indicate the expected form of the inputs.

In this particular case, `Optional` means it can be the thing in `[...]` or `None` (more precisely, `Opional[...] = Union[..., None]`)

And I would actually not think to allow `None` as an input here - if the user has not selected any interfaces, then we simply should not proceed.

So I think this could become `interfaces: List[str]`

Note you will need to add an import for `from typing import List`

pyflask/manageNeuroconv/manage_neuroconv.py
```diff
@@ -24,3 +24,26 @@ def get_all_interface_info() -> dict:
...
+    """
+    Function used to get schema from a CustomNWBConverter that can handle multiple interfaces
+    """
+
+    # Hard coded for now - eventual goal will be to import this from NeuroConv
+    interface_list_subset = [SpikeGLXRecordingInterface, PhySortingInterface]
+
+    if not interfaces:
+        interfaces = [x.__name__ for x in interface_list_subset]  # get schema for all interfaces
```
CodyCBakerPhD, line 38: As per comment above, I would say do not have a default behavior for this - always expect a list of interfaces to be specified

pyflask/manageNeuroconv/manage_neuroconv.py
```diff
@@ -24,3 +24,26 @@ def get_all_interface_info() -> dict:
...
+
+    # Combine Multiple Interfaces
+    class CustomNWBConverter(NWBConverter):
+        data_interface_classes = dict(
+            zip(
+                interfaces,
+                map(lambda name: [x for x in interface_list_subset if x.__name__ == name][0], interfaces),
+            )
```
CodyCBakerPhD, line 46: Closer to how you were doing it before, if you import `neuroconv` at the top you can clean this up a bit with


```python
data_interface_classes = {interface: getattr(neuroconv.datainterfaces, interface) for interface in interfaces}
```

Though I would also suggest at some point that we rename `interfaces` in this function to `interface_names` to more accurately specify the variables contents

pyflask/apis/apiNeuroConv.py
```diff
@@ -21,3 +21,16 @@ def get(self):
...
+
+
+@api.route("/schema")
+class Schemas(Resource):
+    @api.doc(responses={200: "Success", 400: "Bad Request", 500: "Internal server error"})
+    def get(self):
+        args = parser.parse_args()
+        interfaces = args["interfaces"]
+        try:
+            return get_combined_schema(interfaces)
```
CodyCBakerPhD, line 33: So as long as the output of any NeuroConv call is standard JSON, we're OK on compatibility with this flask APi thing?

Sometimes we have variables (we'll see cases of this in the metadata) that require special JSON decoding/encoding. See https://github.com/catalystneuro/neuroconv/blob/main/src/neuroconv/utils/json_schema.py#L16 for full details, will this be a problem we'll have to address?

pyflask/tests/test_neuroconv.py
garrettmflynn: Yeah I'll commit some new changes that do this. That should clear up some of the other suggestions as well.

pyflask/tests/test_neuroconv.py
garrettmflynn: Does `neuroconv` itself provide a schema that I could compare against? Or should I just do this ad hoc?

pyflask/manageNeuroconv/manage_neuroconv.py
garrettmflynn: Thank you for the explanation!

pyflask/tests/test_neuroconv.py
garrettmflynn: Related to this, I'm thinking we'll probably want to reduce the test here since this is really just checking 'get_source_schema()' from the NWBConverter class—which we can assume has already been tested by 'neuroconv', right?

What would you suggest as the right way to test our function rather than just 'neuroconv' output?

CodyCBakerPhD: It's a good question and occurred to me as well - for now, however, I'm happy just to get any testing in place

Now, in general, going forward, what we need to test here is really that our *usage* through *calls* to NeuroConv, especially as we define our own helper functions, return the output/structure we expect for all the cases we expect them here

So no, we don't technically need to duplicate tests like 'does this a valid JSON schema' unless the retrieval function specific to our manager is expected to change it somehow in some way in which case those differences are precisely what would be tested

pyflask/apis/apiNeuroConv.py
CodyCBakerPhD: OK, good to know

Can probably just import that helper from NeuroConv for handling that and add cases to it as we need

pyflask/tests/test_neuroconv.py
garrettmflynn: Implemented a set of updated tests just to become familiar with JSON Schema. Since it really just validates a JSON Schema returned from `neuroconv`, we can adapt `test_multiple_schema_request` as we see fit moving forward.

pyflask/manageNeuroconv/manage_neuroconv.py
```diff
@@ -24,3 +26,15 @@ def get_all_interface_info() -> dict:
                 }
 
     return interface_info
+
+
+def get_combined_schema(interface_class_names: Optional[List[str]]) -> dict:
```
CodyCBakerPhD, line 31:
```suggestion
def get_combined_schema(interface_class_names: List[str]) -> dict:
```

Now that there's no longer any default for `interface_class_name = None`, we can remove `Optional` from the annotation

pyflask/tests/test_neuroconv.py
CodyCBakerPhD: Looks great!

CodyCBakerPhD:
One minor change to the annotation typing to indicate current behavior, and then this looks good to me to proceed!

CodyCBakerPhD (approved):

PR #23, opened 2023-03-09
Implement a second endpoint for a single interface to be used in the …

garrettmflynn:
This PR is a generalization of #20 and #21 that includes two endpoints for requesting schema from the NWBConverter, one that accepts multiple interfaces (`neuroconv/schema?interfaces=InterfaceOne,InterfaceTwo`) and another that accepts one (`neuroconv/schema/InterfaceOne`).


pyflask/tests/test_neuroconv.py
```diff
@@ -53,4 +25,37 @@ def test_multiple_schema_request(client):
...
+                    "type": "object",
+                    "properties": {
+                        "modality": {"type": "string"},
+                        "name": {"type": "string"},
+                        "technique": {"type": "string"},
+                    },
+                }
+            },
+        },
+    )
```
CodyCBakerPhD, line 48: Ooh, very nice test. Nice and explicit and readable!

CodyCBakerPhD (approved):

PR #19, opened 2023-03-08
Updates to SODA text for the first few pages

garrettmflynn:
This PR brings updates to the HTML from #10 to their own branch for review.

PR #24, opened 2023-03-09
Set manual trigger for Chromatic workflow

CodyCBakerPhD:
Towards the goal of testing out Chromatic (free version) on #19, setting trigger for workflow to only occur manually


garrettmflynn (approved):
Looks good to me! Let me know when you've got an instance running, I'd love to see how it feels to interact with the Chromatic UI

CodyCBakerPhD:
> Let me know when you've got an instance running, I'd love to see how it feels to interact with the Chromatic UI

Ahhh I was going to ask if you could handle trying to get that setup 😆 I don't think I have the time right now, plus would be good experience for you to learn how that GitHub action works

CodyCBakerPhD:
The action will for sure need to set a secret variable or two, so if you aren't able to control that just send the tokens to me over a secure line and I'll add them as soon as I can

garrettmflynn:
Okay yeah, I was gonna do that originally—but I don't have the required permissions in this repo to access the Settings tab.

garrettmflynn:
Just to note, Chromatic cannot test #19 as those elements are not components in Storybook. I've created #26 to install Chromatic and have manually tested this installation with a change to the multi-select form component.

PR #26, opened 2023-03-10
Install Chromatic

garrettmflynn:
This PR includes the installation and manual test of Chromatic. Changes can be reviewed [here](https://www.chromatic.com/build?appId=640a4d93e34604f275368983&number=5).

We may want to replace #24 with this PR since it will provide an appropriate environment for our test of the CI workflow.


CodyCBakerPhD:
@garrettmflynn Wow, that is pretty cool. Super easy to use

So what would it take to make this kind of thing work on the changes in #19?

CodyCBakerPhD (approved):

PR #28, opened 2023-03-11
Scope all active JavaScript in the first part of the app

garrettmflynn:
This PR scopes all JavaScript in the application and all interactions should work up to the Guided Mode "Before Getting Started" page.

It makes an effort to separate Node imports and top-level globals into separate files. Where possible, Node imports are also given checks for existence before being used in case they aren't loaded using Electron (e.g. with Vite or other build tools).

The only globals left in the application should be:
1. Encapsulated in unused functions and, therefore, unknown to be global
2. Used in the HTML sheets injected into the DOM (e.g. made global at the bottom of guided-curate-dataset.js)

While there are many changes, this PR followed a predictable development workflow of scoping all the JavaScript using native `require` imports and triaging all errors up to the desired point.


CodyCBakerPhD (approved):

CodyCBakerPhD:
Thanks for the detailed explanation! It's great for establishing provenance even if there aren't any questions about the approach right away

PR #29, opened 2023-03-13
Convert require to ES6 imports

garrettmflynn:
This PR introduces ES6 imports for all frontend resources currently being used from SODA.

Free Form Mode has been removed again to constrain the number of edits that must be made to ensure performance on the active portion of the app. As in #28, all interactions should work up to the Guided Mode "Before Getting Started" page.

Conditional imports of Electron features were tested up to using a Vite dev server, indicating that the app can run independently of the Electron distribution.


CodyCBakerPhD:
> Free Form Mode has been removed again to constrain the number of edits that must be made to ensure performance on the active portion of the app. 

Yes, that's fine. This process should be mostly guided anyway, not sure what features would fall under 'free form' mode for us

CodyCBakerPhD (approved):

PR #32, opened 2023-03-14
Create components to replicate SODA frontend

garrettmflynn:
This PR introduces a Main component that holds pages and cleans up related code (e.g. index.html, Sidebar.js, etc.)


CodyCBakerPhD:
> Happy to set up Storybook so we can start reviewing page changes using Chromatic, before or after doesn't matter to me.

I am loving Chromatic so far, so as soon as possible on that would be great, but I'll still check this out as is

garrettmflynn:
## Issue with UI
The UI will render that way if you access the page before the Python server is connected. Maybe it isn't running on your computer? This would happen if my virtual environment isn't active or out of date.

## Loading Progress
This is possible to add back easily, it'd probably take ~1-2hrs to get the UI to render properly and the application state to be saved again. Would you like me to handle this before Monday's meeting or later?

CodyCBakerPhD:
> This is possible to add back easily, it'd probably take ~1-2hrs to get the UI to render properly and the application state to be saved again. Would you like me to handle this before Monday's meeting or later?

It's whenever you can get around to it - the principle is one of forward progress, we shouldn't be merging any PRs that lose any functionality already available, functional on `main`, and a part of the MVP - it's also no problem to continue other work on chained PRs further than #38 

That said, if there's a technical reason why you can't support the saving of page progress per dataset based on our required changes, that's a fine thing to split to a different PR

CodyCBakerPhD:
I tried a hard wipe and re-install of everything again and it works - just to note though, my dev console looks ike

Screenshot 2023-03-17 at 4 47 12 PM

and in the previous error case that 'cannot connect' just continues ad infinitum

garrettmflynn:
Looks good! Yeah, the Electron app is still set to timeout in 2min—which is
quite a long time.

CodyCBakerPhD:
Looks great, thanks for the hard work!

CodyCBakerPhD (approved):

PR #34, opened 2023-03-17
Add a search component

garrettmflynn:
This PR introduces a search component into Storybook and NWB GUIDE to replace the functionality of the multi-select form as suggested by @bendichter.

## Storybook Usage
Screenshot 2023-03-17 at 8 36 23 AM

## NWB GUIDE Usage
Screenshot 2023-03-17 at 8 35 28 AM


CodyCBakerPhD:
> replace the functionality of the multi-select form

Ultimately, both a search and multi-select could end up being useful for different reasons, so 'replacement' is a bit strong - maybe as a separate tab?

I'd also be fine tabling this for a while until the MVP is more complete. For MVP we only need a minimal form

garrettmflynn:
Sure, that was slightly stronger than I intended. I've created a _new_ Search component that is replacing the multi-select specifically for the Data Formats page. 

The Search option is actually cleaner than the multi-select anyways and, according to Ben, seems to be a common standard for selecting options like this.

The issue with a "minimal form" is that, eventually, we want to allow users to specify many DataInterfaces (likely known, at least by certain keywords) and provide arbitrary keys for them—making it possible to select many of the same interfaces. The problem with a standard form is that _options only show up once_ so we can't allow multi-selection of the same option unless we provide a similar user experience to search (i.e. select one, register it, rename it, repeat) but settle for a cluttered interface. 

Of course, the MVP _can_ be simplified from our desired goal. But since a search component is scalable and finished, I don't see a reason not to include it—aside from constraining the possible selections, which we can do with this as well.

CodyCBakerPhD:
> The Search option is actually cleaner than the multi-select anyways

It is cleaner, yes, but much more constrained and the burden is on the user to think up good search terms and on devs to associate good search terms


> I've created a new Search component that is replacing the multi-select specifically for the Data Formats page.

My issue is still with outright 'replacing', surely it's just as easy to keep the current setup as a toggled option as described below - or are tabs more difficult to implement in JavaScript than they are in `ipywidgets` (~few lines of wrapping code that routes outputs)?

Point being, especially in these early days, smart search may fail to find what the person is looking for so we do need a fallback that allows them to more easily navigate the entire library of supported interfaces

This is especially true because we do not know all the various aliases and references to techniques that people from different corners of neuroscience use (I hear a new phrase every month it seems), so we can't always guarantee 100% that what someone types into the search bar will lead them to what they need

As such I don't see the problem in having fancy search be the default of two tabs selectable at the top of the page (example: one magnifying glass, one book) with the 'book' being the full listing organized by modality+technique, as sup-optimal as that might seem


> The issue with a "minimal form" is that, eventually, we want to allow users to specify many DataInterfaces (likely known, at least by certain keywords) and provide arbitrary keys for them—making it possible to select many of the same interfaces. The problem with a standard form is that options only show up once so we can't allow multi-selection of the same option unless we provide a similar user experience to search (i.e. select one, register it, rename it, repeat) but settle for a cluttered interface.

The workflow for the form selection should match the search approach anyway for consistency, so that part is fine for now



> Of course, the MVP can be simplified from our desired goal. But since a search component is scalable and finished, I don't see a reason not to include it

Smart search is something we would only be able to fully test and fine tune when we open up the flood gate of all the other interfaces, so even though the bar and feature itself is functional, the actual practical usage of it is not fully tested yet because we're throttling the size of the search space so much

I've raised issues #35, #36, and #37 outlining the more important next 3 steps that need to be completed for the MVP, and I don't really view this feature as necessary for accomplishing that goal, but am perfectly happy and excited to either

(a) revisit this once we have the minimal workflow operational

or

(b) have this be one of two possible ways of selecting the interfaces, so we don't become too limiting

garrettmflynn:
Sounds good, thank you for the detailed explanation. I'll focus on those issues for now.

CodyCBakerPhD:
> I think a search is fine as long as an empty string shows the entire available list as a dropdown.

Ohh I didn't even think of that

Yeah no need for separate tabs if the default 'search' is the full listing, though I still think it could be beneficial *eventually* (at some point down the line) to organize even that list in a more navigable way (scrolling through 40+ items in alphabetical or random order is not as convenient) but these are all speculative UX elements for enhanced search so not an immediate concern

Will revisit this enhancement after conclusion of #37

CodyCBakerPhD:
Some merge conflicts to resolve here to test ideas from discussion #75

CodyCBakerPhD:
# Final To-Do's

The key value of the interface selector should be

(a) hidden if all the selected interfaces are unique
(b) shown and editable (I can't currently edit it) if there is at least one pair of interfaces that is duplicated (can this be restricted only to the subset of interfaces that are duplicated?)

Also for enhanced appearances

(i) On the main selector page, it would be great if we could use more human-readable aliases for the titles - I'm happy to provide these as a dictionary if that would help?

```python
{
    "SpikeGLXRecordingInterface": "SpikeGLX/NeuroPixels (single-band)",
   "PhySortingInterface": "Phy",
}
```

(ii)  Numeric increments should start with first; no underscores when appending the numeric; example, default keys should show up as

```
SpikeGLXRecording1 - SpikeGLXRecordingInterface
SpikeGLXRecording2 - SpikeGLXRecordingInterface
```

(iii) II know we 'described' this section in the early summary, but it would be great to add a text selection at the top exaplining what we mean by 'selecting data formats' - in particular, explaining that each interface is a different stream.

How hard is it to add those little blue info hover icons next to an interface to explain specific details? Like, I'd want one for SpikeGLX currently to explain that it's either the AP or LF band of a single probe



Not hugely important, but I also noticed if I type 'Phy' into the search bar it doesn't exclude SpikeGLX... Hmm it isn't because of the a keyword containing 'electroPHYsiology' is it?

garrettmflynn:
I'm curious whether the lack of editability of the key is related to the drag-and-drop issues seen in #94 [[link](https://github.com/NeurodataWithoutBorders/nwb-guide/pull/94#issuecomment-1489158019)]. If this is persistent, then I can replace the element currently using the contentEditable attribute with an Input element that is more reliable across systems.

To summarize this in the context of our talk earlier today:
1. Instead of hiding until there is a duplicate, we will instead indicate the key is editable with a small icon
2. Mock up a "label" key for each supported interface until we add this to the actual DataInterface classes
3. Remove the underscore when iterating
4. Mock up a "description" key for each supported interface until we define a way to access these automatically
    - If you look at #70, you'll see that the source data schema descriptions do this quite nicely, though they are accessed in a subsequent step at the moment
5. Initially conduct a search matching from the beginning of the keywords / name (i.e. so phy only matches Phy in this case)

CodyCBakerPhD:
> Is this correct?

Sounds good, emphasis on keys _always_ being editable so users can leverage it to name things they way they are 'used to' in their setup

CodyCBakerPhD:
Still some things to improve in the long term, but meets our current needs! Thanks for the hard work on this

CodyCBakerPhD (approved):

Issue #35, opened 2023-03-17
[MVP Step 1]: Source schema (specify file/folder locations)


CodyCBakerPhD:
We've established an entrypoint for fetching a couple minimal interfaces, and rendered those as a form in the app

Next step is for the user to select different combinations of these available interfaces, and then ask the user to specify the 'source data' for each interfaces as fetched from the source schema of an NWBConverter created containing those selected interfaces

CodyCBakerPhD:
Done in #38

Issue #36, opened 2023-03-17
[MVP Step 2]: Metadata specification


CodyCBakerPhD:
After #35 is complete, the next step is to fetch as much metadata as possible automatically from the initialized interfaces

Ultimately, this is going to be the part that can be really optimized down the road to feel like an easy TurboTax like step-by-step walkthrough with explanations of complicated metadata values and structures

I'm guessing, however, that a first-pass MVP would simply be a web form that auto-populates many fields and flags many others as required

To constrain the complexity of the metadata early on, I think it's safe enough with SpikeGLX+Phy to simply restrict the metadata fields to focus on the NWB file and Subject level, since all `ecephys` specific information is fetched automatically

That way, we can design the `ecephys` forms a little more detailed in a follow up that is compatible with less specified formats

garrettmflynn:
@CodyCBakerPhD Can you link me to a good dataset to test this combination of Phy + SpikeGLX? Otherwise I'll need to come up with a good way to mock the results of `get_metadata_schema`

CodyCBakerPhD:
All testing for all data formats in the GUIDE (by way of NeuroConv) can be done using our example data repo

Instructions on how to download data: https://neuroconv.readthedocs.io/en/main/developer_guide/testing_suite.html#on-data

For SpikeGLX and Phy, we usually use https://gin.g-node.org/NeuralEnsemble/ephy_testing_data/src/master/spikeglx/Noise4Sam_g0 and https://gin.g-node.org/NeuralEnsemble/ephy_testing_data/src/master/phy/phy_example_0

CodyCBakerPhD:
Done in #47

PR #38, opened 2023-03-17
Complete a draft schema form workflow

garrettmflynn:
This PR addresses #35 and introduces a workflow for requesting information from users based on the schemas returned from `neuroconv/schema?interfaces=xxx`. It tracks user inputs and sends these to the next page—though this can later be validated and passed through a `neuroconv/metadata` endpoint.


src/stories/pages/guided-mode/data/GuidedSourceData.js
```diff
@@ -2,15 +2,55 @@
 
 import { html } from 'lit';
 import { Page } from '../../Page.js';
+import electronImports from '../../../../electron/index'
+const { dialog } = electronImports.remote ?? {};
+
+import globals from '../../../../../scripts/globals.js';
+const { port } = globals;
+
+const base = `http://127.0.0.1:${port}`;
```
CodyCBakerPhD, line 11: Maybe a silly question, but hard-coding this URL, is that always for sure exactly the same on all platforms and boot-up styles? (I.e., if I install via the packaged/built executable, will it still run through a port on the local host like this?

src/stories/pages/guided-mode/data/GuidedSourceData.js
```diff
@@ -2,15 +2,55 @@
 
 import { html } from 'lit';
 import { Page } from '../../Page.js';
+import electronImports from '../../../../electron/index'
+const { dialog } = electronImports.remote ?? {};
```
CodyCBakerPhD, line 6: I'm just curious as I learn more JavaScript...

What does `??` mean?

CodyCBakerPhD: Also, I notice there's no terminating `;` after the import here, but there is after the definition of the `const` - based on the pattern at line 8-9, shouldn't there be a `;` after the import?

What are the rules for line termination like this in JS?

src/stories/pages/guided-mode/data/GuidedSourceData.js
garrettmflynn: Yep! That is just the standard localhost ip—though I could just swap it to localhost with no change.

CodyCBakerPhD:
OK so this all looks like front-end stuff, right?

The interfaces are selected, their schema is retrieved based on that, but that info is then sent to the next page where the info will be specified?

Principle sounds fine, but unable to test due to the interface selector issue on the parent branch, will check again once that's resolved

src/stories/pages/guided-mode/data/GuidedSourceData.js
garrettmflynn: ## Nullish Coalescing (??)
The `??` operator allows you to return the right-hand value if the left is undefined. Fun stuff, you can read more on the [official MDN docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing) (this is generally the bible for general web reference material).

## Termination (;)
Termination is not required, unless you'd like to have multiple statements inline: 
```javascript
const test = 0; function testFunction(){ return 'wow'}; 
```

It looks like the SODA developers did use them—though I generally don't. It's really just a convention, something we can probably configure in `prettier`.

garrettmflynn:
Yeah, just frontend for this. All the backend changes were made in #21.

The workflow is Specify Data Interfaces -> Specify Schema Source Data -> Specify Missing Metadata, where (1) and (2) are complete and (3) receives mocked-up data from (2) that will eventually be the returned result of #36.

src/stories/pages/guided-mode/data/GuidedSourceData.js
CodyCBakerPhD: Thanks for info!

To reciprocate, the Python `or` behaves like `??`

```python
val = None or {}
> 
```

and can even handle various other 'empty' cases though that can get a little touchy

```python
val1 = 0 or 1
> 1
val2 = "" or "abc"
> "abc"
val3 = {} or dict(a=1)
> {a: 1}
```

Python's also the same way with termination, but it's a very unrecommended practice to ever have multiple statements on one line, so you rarely ever see it used, so sounds like you've followed that yourself

CodyCBakerPhD:
OK got it working now...

Some feedback necessary to move to the next step... based on

Screenshot 2023-03-17 at 4 51 46 PM

We can ignore the 'verbose' argument entirely, that's an SDK level feature used to mimic a very basic GUI like feedback when running pure code, the GUIDE is essentially a permanent form of that verbosity (and far superior form of it)

`es_key` should have defaulted to something I believe, did it not?

`exclude_cluster_group` might also be something that we should skip for just a little while and add support for it on a separate PR

CodyCBakerPhD:
> I'll ignore everything except the required fields (file_path and folder_path) for now to keep things simple

I think that's fine - the `es_key` is the most important one there and we may have to figure it out sooner rather than later, but for a first PR the paths alone are OK

garrettmflynn:
@CodyCBakerPhD Anything holding this back from merging into `main`? The only thing I can think of is the completeness of the JSON Schema form—though I don't know if that needs to be in-scope here.

CodyCBakerPhD:
> Anything holding this back from merging into main?

I was awaiting thorough testing of it on my end, which was delayed temporarily by #42, checking it out now

CodyCBakerPhD:
Just leaving as a note (I assume it is trivial to fix and will be replaced by something later on): the order of the interfaces as rendered by the source schema path selection buttons - would prefer SpikeGLX (recording) before Phy (processed from the recording) just to maintain the natural way of thinking about the data processing workflow

CodyCBakerPhD:
Otherwise, it works, looks great!

Nice to finally have a small changelog on a front-end addition lol

PR #40, opened 2023-03-18
Save and reload progress

garrettmflynn:
This PR fixes the progress saving / reloading features of the original SODA application that was lost in #32. It completes this behavior up until the Metadata page.

PR #41, opened 2023-03-20
Minimally save and reload progress

garrettmflynn:
This PR reimplements the ability to save and load progress that was removed in #32. It also reimplements #40 in a minimal way using #32 as a base.


CodyCBakerPhD (approved):

CodyCBakerPhD:
Working for me, thanks for restoring!

PR #47, opened 2023-03-20
Complete metadata query

garrettmflynn:
This PR introduces frontend and backend code to handle a complete metadata query. It renders minimal metadata results from `neuroconv` (e.g. required properties) on the Metadata page.

PR #50, opened 2023-03-21
Track and render metadata properly

garrettmflynn:
This PR introduces a JSONSchemaForm component that allows us to test whether metadata is tracked properly for user-defined properties as well as neuroconv / NWB schema defaults.


CodyCBakerPhD:
Was this supposed to point to #47?

garrettmflynn:
Yep, my bad! Just switched where this points to

PR #52, opened 2023-03-21
Run a full conversion for Phy + SpikeGLX

### LINKED ISSUE #37: [MVP Step 3]: Run a conversion through the local app
After metadata is specified in #36, the user will want to run a conversion

There is a matter of some conversion options, but they are fairly limited for the time being - the only important one is technically the 'stub_test' but we were maybe envisioning that as being done automatically every time to form a 'preview' that displays before asking the user to 'confirm' they want to 'run the conversion'

Other than that, this MVP step will be satisfied by a simple button that when pressed, calls `.run_conversion` on the `NWBConverter` specified by all the previous steps

garrettmflynn:
fix #37 

This PR introduces a complete workflow for running a single conversion on Phy + SpikeGLX data to complete #37.

## Bugs to Resolve in Future PRs
1. Reloading a project will not maintain the navigation sidebar behaviors from the earlier session
2. Completed conversions are not transferred to the Completed tab on the start page


pyflask/manageNeuroconv/manage_neuroconv.py
```diff
@@ -39,6 +39,14 @@ class CustomNWBConverter(NWBConverter):
     return CustomNWBConverter
 
 
+def instantiate_custom_converter(source_data):
+    interface_class_names = list(
+        source_data
+    )  # NOTE: We currently assume that the keys of the properties dict are the interface names
+    CustomNWBConverter = get_custom_converter(interface_class_names)
+    return CustomNWBConverter(source_data)
```
CodyCBakerPhD, line 47:
```suggestion
def instantiate_custom_converter(source_data: Dict[str, str]) -> NWBConverter:
    interface_class_names = list(
        source_data
    )  # NOTE: We currently assume that the keys of the properties dict are the interface names
    CustomNWBConverter = get_custom_converter(interface_class_names)
    return CustomNWBConverter(source_data)
```

pyflask/manageNeuroconv/manage_neuroconv.py
```diff
@@ -52,11 +60,26 @@ def get_metadata_schema(source_data: Dict[str, dict]) -> Dict[str, dict]:
...
-    )  # NOTE: We currently assume that the keys of the properties dict are the interface names
-    CustomNWBConverter = get_custom_converter(interface_class_names)
-    converter = CustomNWBConverter(source_data)
+    converter = instantiate_custom_converter(source_data)
     schema = converter.get_metadata_schema()
     metadata = converter.get_metadata()
     return json.loads(json.dumps(dict(results=metadata, schema=schema), cls=NWBMetaDataEncoder))
+
+
+def convert_to_nwb(info):
```
CodyCBakerPhD, line 69:
```suggestion
def convert_to_nwb(info: dict) -> bool:
```

src/pages.js
```diff
@@ -112,14 +112,14 @@ dashboard.pages = {
...
-                section: sections[2]
-            }),
+            // upload: new GuidedUploadPage({
+            //     label: "Upload dataset",
+            //     section: sections[2]
+            // }),
+            // deletefiles: new GuidedDeleteFilesPage({
+            //     label: "Delete files",
+            //     section: sections[2]
+            // }),
```
CodyCBakerPhD, line 122: Planning to keep this going forward?

src/stories/pages/guided-mode/options/GuidedConversionType.js
```diff
@@ -1,27 +1,120 @@
...
+          html: "Please wait...",
+          allowEscapeKey: false,
+          allowOutsideClick: false,
+          heightAuto: false,
+          backdrop: "rgba(0,0,0, 0.4)",
+          timerProgressBar: false,
+          didOpen: () => {
+            Swal.showLoading();
+          },
+        })
```
CodyCBakerPhD, line 68: So one thing I'm curious about here, that we won't be able to test until I provide some actual full-sized data - how does this track progress from the Python side?

For instance, when using the NeuroConv API various interfaces actually have a conversion option for displaying progress but it uses a Python package called `tqdm`, but that spits out along a specifiable `stdout` pipe, can the front-end control or interact with that at all?

garrettmflynn: This doesn't track progress on the Python side at the moment. It only has two reactions at the moment: Close when completed successfully or close and throw an error if something fails from Python.

I can interact with streams from the frontend, so we should be able to find a way to pass whatever you give to `tqdm` to me. Generally, I need the total number of things that will happen and where we are (or, alternatively, the current progress from 0-1) to render a progress bar—which is easy to come by for files since they broadcast their content length before the stream begins.

src/pages.js
garrettmflynn: They will likely be renamed and their contents changed/condensed. I can delete these lines and/or the files if you would like.

CodyCBakerPhD: If they won't ever be needed, I'd say remove them in separate PR

If they're going to be edited in some way in the future then fine to keep, but why comment them out here then?

src/stories/pages/guided-mode/options/GuidedConversionType.js
CodyCBakerPhD: Understood, we'll revisit this later then as a final polish

src/pages.js
garrettmflynn: They're commented here since we're not providing users with the option to (1) upload their data or (2) delete files at present—though these are part of the Figma document that we intend to include later and I thought would be useful to mockup in the code itself.

They may no longer be useful though since we have something that works and don't need to clutter the UI with "Coming soon..." messages.

CodyCBakerPhD: > They may no longer be useful though since we have something that works and don't need to clutter the UI with "Coming soon..." messages.

I don't mind having placeholders in the MVP, the only time they'd need to be totally cleaned up is by first round of beta testing which is a decent ways out

pyflask/manageNeuroconv/manage_neuroconv.py
```diff
@@ -52,11 +60,26 @@ def get_metadata_schema(source_data: Dict[str, dict]) -> Dict[str, dict]:
...
+
+    converter = instantiate_custom_converter(info["source_data"])
+
+    converter.run_conversion(
+        metadata=info["metadata"],
+        nwbfile_path=info["nwbfile_path"],
+        # save_to_file=info.save_to_file,
+        # overwrite=info.overwrite,
+        # conversion_options=info.conversion_options,
+        # stub_test=info.stub_test,
```
CodyCBakerPhD, line 82: Just some notes

`stub_test` is actually a per-interface conversion option - it's possible to run 'partial' stubs in this way. Most common case of that is you don't want to stub the spike sorting data but you do want to stub the raw ephys data

`save_to_file` is only used for in-memory things with the API, the GUIDE should never need to use it

We will want a button for `overwrite`, or maybe a prompt in the event the file already exists? Isn't that how most apps deal with file overwrite permission?

CodyCBakerPhD:
One thing to note here is the `ecephys` is still a section of the metadata form

But it didn't allow modification of anything other than the name of the series so I guess for now that is fine; just trying to constrain the complexity of the current scope

It all works, though!

pyflask/manageNeuroconv/manage_neuroconv.py
garrettmflynn: Thanks for the clarification!

## stub_test
Alright, I'll make sure we can specify these separately

## overwrite
Yeah, I'm already spitting out an error if the file already exists. I can switch that for a message that confirms if the user would like to overwrite it.

CodyCBakerPhD (approved):

Issue #51, opened 2023-03-21
[Bug]: Chromatic and MultiSelect


CodyCBakerPhD:
As seen from last trigger of Chromatic, complains about the MultiSelect form

```
Could not resolve "./multiselect/index.js" from "src/stories/Multiselect.stories.js"
```

https://github.com/NeurodataWithoutBorders/nwb-guide/actions/runs/4481833739/jobs/7879085476

CodyCBakerPhD:
Was run on the branch for #50 if that helps any

garrettmflynn:
@CodyCBakerPhD This should be addressed in #53.

PR #53, opened 2023-03-21
Relink the multiselect form to its story

garrettmflynn:
This PR addresses #51 to fix the resolution of the MultiSelectForm class in Storybook.

### DIFF
--- a/src/stories/multiselect/Multiselect.stories.js
+++ b/src/stories/multiselect/Multiselect.stories.js
@@ -1,4 +1,4 @@
-import { MultiSelectForm } from './multiselect/index.js';
+import { MultiSelectForm } from './MultiSelectForm.js';
 
 export default {
   title: 'Example/Multiselect Form'


CodyCBakerPhD:
@garrettmflynn Now getting 'component errors' but nothing specific about what that means - does it make any sense? https://github.com/NeurodataWithoutBorders/nwb-guide/actions/runs/4483228361/jobs/7882252757

garrettmflynn:
Oh looks like there were some issues with how I'd refactored the Button component. 

We don't need `Page.js` or `Header.js` since those were Storybook defaults, so I'll just remove those.

Otherwise I'm updating the `Button.stories.js` to be compatible with class-based Web Components.

garrettmflynn:
@CodyCBakerPhD Was able to resolve these issues with [Build 20](https://www.chromatic.com/build?appId=640a4d93e34604f275368983&number=20) on Chromatic.

Components that use global styling are hidden so we snapshot the minimal number of components over time.

PR #54, opened 2023-03-21
Separate successful conversions from drafts

garrettmflynn:
This PR changes the rendering of successful / completed conversions on the main Getting Started page so that successful conversions are distinguished from drafted ones. 

A few cosmetic changes are also added to the page to align it with NWB GUIDE rather than SODA.


CodyCBakerPhD:
Maybe we should mock it up a little bit in Figma first - putting this into draft until then

CodyCBakerPhD:
Can you draft a small mockup for this? I'm envisioning that the existing pipelines looks exactly as it does now except any pipeline that didn't hit the 'run conversion' button is flagged as 'in progress'

The history tracker is probably a separate feature altogether and can be put in a different place in the app too

CodyCBakerPhD:
@garrettmflynn Should we keep this open, move to an issue, or close?

garrettmflynn:
If we don't really want this separation (at least for now), we can just close it.

Otherwise we should clearly define how we'd want a separation between existing conversions and completed conversions to look.

PR #56, opened 2023-03-21
Remove footer overlap with main page content

garrettmflynn:
This PR removes overlap between the footer and main page content.

### DIFF
--- a/assets/css/demo.css
+++ b/assets/css/demo.css
@@ -446,7 +446,7 @@ button.download-metadata:hover {
 
 /* Current interactive UI's main content */
 .dash-app {
-  padding-bottom: 10px !important;
+  /* padding-bottom: 10px !important; */
 }
 
 .dash-title {
--- a/src/stories/Main.js
+++ b/src/stories/Main.js
@@ -10,6 +10,8 @@ import { GuidedHeader } from './pages/guided-mode/GuidedHeader.js';
 const componentCSS = `
     :host {
         position: relative;
+        display: grid;
+        grid-template-rows: fit-content(100%) 1fr fit-content(100%);
     }
 `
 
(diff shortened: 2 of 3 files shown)


assets/css/demo.css
```diff
@@ -446,7 +446,7 @@ button.download-metadata:hover {
 
 /* Current interactive UI's main content */
 .dash-app {
-  padding-bottom: 10px !important;
+  /* padding-bottom: 10px !important; */
```
CodyCBakerPhD, line 449: This is a comment right? So the 'padding-bottom' actually wasn't doing anything?

src/stories/Main.js
```diff
@@ -115,14 +117,14 @@ export class Main extends LitElement {
       }
     }
 
-    const headerEl = header ? new GuidedHeader(header) : ''
+    const headerEl = header ? new GuidedHeader(header) : html`<div></div>` // Render for grid
```
CodyCBakerPhD, line 120: Is this just an empty box at the bottom to artificially produce the padding?

garrettmflynn: Not quite. Grid rows are activated on a element-by-element basis, so the 1st element will always take on the properties of the 2nd row—in this case, take up the area that your content needs.

If I use an empty string, this doesn't count as an element and, as such, rows 2/3 become rows 1/2. Adding the empty div means that these don't change.

Since the div is empty and the row height is based on content, the first row scales to 0px of height.

assets/css/demo.css
garrettmflynn: Yeah, it's a comment. I've removed it since, after the Web Component refactor, it was unnecessarily adding padding below the footer.

CodyCBakerPhD (approved):

CodyCBakerPhD:
Confirmed, it looks much better now

PR #57, opened 2023-03-21
Reload the navigation sidebar from saved project

garrettmflynn:
This PR allows for reloading the navigation sidebar (Guided Mode sidebar) with an accurate history of the pages visited by the users.

### DIFF
--- a/src/stories/Dashboard.js
+++ b/src/stories/Dashboard.js
@@ -63,10 +63,6 @@ export class Dashboard extends LitElement {
     else if (typeof page === 'object') return this.getPage(Object.values(page)[0])
   }
 
-  reset() {
-    this.#sectionStates = {} // Reset state of the navigation sidebar
-  }
-
   setMain(page, infoPassed){
 
     // Update Previous Page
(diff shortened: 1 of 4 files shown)


CodyCBakerPhD:
Sorry, I must be missing it - I went back and forth between this branch and `main` a couple times and couldn't see the difference

What specific behavior does this PR enable?

garrettmflynn:
The Guided Mode navigation bar was not reloading progress (i.e. it would not register that you visited pages 1-4 and spit you onto page 5 without the ability to backtrack on the sidebar itself)

CodyCBakerPhD:
Ohhh I see now

For future bug fixes like this, a more step-by-step set of instructions to reproduce the problem on `main` (or even a quick video demonstrating it) helps a lot

Also as soon as you detect something like this feel free to raise an issue first so everyone can be aware of it, and then we can discuss/prioritize it along with everything else on the docket

CodyCBakerPhD (approved):

garrettmflynn:
Alright will do that in the future!

PR #63, opened 2023-03-23
Render app pages in Storybook

garrettmflynn:
This PR allows for rendering all app pages in Storybook and Chromatic for simple review of frontend changes as discussed in #19.

Moving forward, we will be able to run Chromatic selectively based on when changes need to be reviewed.

## Chromatic Usage
We can currently trigger Chromatic manually in the Github Actions panel.

@CodyCBakerPhD The Chromatic CLI allows you to run Chromatic on select stories (i.e. for a single page). How do you think we should manage this moving forward given that we have a limited number of snapshots available for a given time frame. 

Unless there's a simple way to pass configuration options to a workflow, I'll probably just run the CLI locally with the required options for a particular PR.


garrettmflynn:
Fixed some issues with the resolution of assets in Chromatic. 

Based on the [latest build](https://www.chromatic.com/build?appId=640a4d93e34604f275368983&number=31), this should be good to review.

## Notes
Global styles are applied to all Pages—but not the Button or MultiSelectForm. So they will render in a way that isn't representative to their usage in the app.

Many page states (e.g. conversion successful) are missing because no progress is being represented. This could be a focus of a future PR—but isn't really necessary as we change text in the short term.

CodyCBakerPhD:
> The Chromatic CLI allows you to run Chromatic on select stories (i.e. for a single page). How do you think we should manage this moving forward given that we have a limited number of snapshots available for a given time frame.

How many snapshots are we currently at per run? One per page? One per individual component?

As long as there's only a handful of runs on a single PR, a couple PRs per day, we should be able to stay within free tier easily

CodyCBakerPhD:
This is looking really good!

But yeah, rendering appearances to mimic interactions with the backend is going to be tricky indeed - fine for now with keeping it just as visuals (and tests) of front-end components only

CodyCBakerPhD (approved):

PR #68, opened 2023-03-24
Render styles properly in Electron

### LINKED ISSUE #67: Electron Styles Missing
After #63, the global styles that were attached to the Dashboard component were no longer included in the Electron build because of inconsistent inclusion between the dev server and the built file. 

Specifically, the dev server includes the CSS while the built file has an adjacent `style.css` file that isn't linked.

To fix this, we will have to conditionally link to the built CSS in the Electron build.

garrettmflynn:
fix #67

### DIFF
--- a/index.html
+++ b/index.html
@@ -64,7 +64,14 @@
     <script src="https://bossanova.uk/jspreadsheet/v4/jexcel.js"></script>
 
     <script type="module">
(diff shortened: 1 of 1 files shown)


CodyCBakerPhD:
Confirmed it works

CodyCBakerPhD (approved):

PR #66, opened 2023-03-24
Update guided pages with NWB GUIDE content

garrettmflynn:
fix #61 

This PR updates Guided Mode pages with GUIDE-related content. Review [Build 39](https://www.chromatic.com/build?appId=640a4d93e34604f275368983&number=39) for changes.

PR #74, opened 2023-03-26
Chromatic animation fix

garrettmflynn:
This PR addresses [this comment](https://github.com/NeurodataWithoutBorders/nwb-guide/pull/66#issuecomment-1483865972) from #66 to render lottie animations consistently when using Chromatic to avoid unnecessary flagged changes.


CodyCBakerPhD:
Chromatic run seems to have failed? https://github.com/NeurodataWithoutBorders/nwb-guide/actions/runs/4532353555/jobs/7983617774

garrettmflynn:
Should be fixed now. Was accidentally referring to a variable before instantiation in #66 for a conditional that only runs in Storybook.

This is an issue that is particularly well-suited for recognition in TypeScript files—otherwise, the native linter (for VS Code at least) doesn't pick this sort of thing up. So if we want to avoid more issues like this, upgrading all our files to .ts might be the way to go.

CodyCBakerPhD:
@garrettmflynn IDK, chromatic seems to think something weird is going on with these page comparisons: [Build 43](https://www.chromatic.com/test?appId=640a4d93e34604f275368983&id=64225553320ce271441f7529)

garrettmflynn:
I think you’ll have to approve the new baseline before it will be compared
against. Have you done this yet?

CodyCBakerPhD:
> I think you’ll have to approve the new baseline before it will be compared
against. Have you done this yet?

I approved but I'm guessing it needs to be merged into main to take effect, which I'd be fine with, just a couple conflicts here now

garrettmflynn:
Now that this branch is the baseline, we just have to push another Chromatic build from the branch to check if it registers changes or not.

CodyCBakerPhD:
> Now that this branch is the baseline, we just have to push another Chromatic build from the branch to check if it registers changes or not.

It skips the rebuild because of no new commits - if you resolve the conflicts I should be able to issue a rebuild to test that

garrettmflynn:
Merged! Sorry for the delay here

CodyCBakerPhD:
Now chromatic build failed: https://github.com/NeurodataWithoutBorders/nwb-guide/actions/runs/4557457180/jobs/8039114687

garrettmflynn:
The workflow uses `yarn` (equivalent to `npm install`) to initialize the dependencies and, since the version of Storybook is loose, installed the next version which includes new mandatory peer dependencies.

Just created two successful builds, [47](https://www.chromatic.com/build?appId=640a4d93e34604f275368983&number=47) and [48](https://www.chromatic.com/build?appId=640a4d93e34604f275368983&number=48) (no actions required since identical), to reset and test the comparison since Build 43 was actually completely wrong (all pages were the homepage). 

This was all after fixing a few issues with Build 46 caused by some interactions between Storybook and the new main branch.

CodyCBakerPhD:
> The workflow uses yarn 

Would it make sense to adjust the workflow to use `npm`?

garrettmflynn:
Yeah I'd say it'd be most consistent to use `npm ci` if possible

PR #84, opened 2023-03-27
Add frontend and backend code to handle stub test

### LINKED ISSUE #77: Generate file preview before full run
Stub mode is super fast 99% of the time, and it would be great to be able to show a basic preview of the file structure before the person hits 'run' and waits for several hours

Of course, the ideal dream for this would be to call NWB Widgets on the stub file and let the user explore interactively. But I'm guessing that's a big ask to integrate, so I'd settle for any sort of intelligible rendering of the NWB file (even the print out text representation from `io.read()`)

garrettmflynn:
fix #77 

This PR introduces new pages and backend code to handle the execution of a stub test and the presentation of its results.

Additional changes were made to prettify the rendering of the new pages in the context of the application as a whole.


pyflask/manageNeuroconv/manage_neuroconv.py
```diff
@@ -71,15 +80,29 @@ def convert_to_nwb(info: dict) -> bool:
...
+    nwbfile_path = info["nwbfile_path"]
+
+    runStubTest = ("stub_test" in info) and info["stub_test"]
+
+    # add a subdirectory to a filepath if stub_test is true
+    if runStubTest:
+        filename = os.path.basename(nwbfile_path)
+        dirname = os.path.join(os.path.dirname(nwbfile_path), ".stubs")
+        nwbfile_path = os.path.join(dirname, filename)
+        check_dir(dirname)
```
CodyCBakerPhD, line 92:
```suggestion
    run_stub_test = ("stub_test" in info) and info["stub_test"]

    # add a subdirectory to a filepath if stub_test is true
    if run_stub_test:
        filename = os.path.basename(nwbfile_path)
        dirname = os.path.join(os.path.dirname(nwbfile_path), ".stubs")
        nwbfile_path = os.path.join(dirname, filename)
        check_dir(dirname)
```
Always snake case for instances of classes and variables

pyflask/manageNeuroconv/manage_neuroconv.py
```diff
@@ -71,15 +80,29 @@ def convert_to_nwb(info: dict) -> bool:
     Function used to convert the source data to NWB format using the specified metadata.
     """
 
+    nwbfile_path = info["nwbfile_path"]
+
+    runStubTest = ("stub_test" in info) and info["stub_test"]
```
CodyCBakerPhD, line 85:
```suggestion
    runStubTest =  info.get("stub_test")
```
Dictionaries in Python have a 'safe' retrieval functionality through `.get(key, default)`

`default` is `None` by, well, 'default'; but `if` logic treats `None` the same as `False` so works as is with the logic below

pyflask/manageNeuroconv/manage_neuroconv.py
```diff
@@ -5,6 +5,15 @@
...
 from neuroconv.utils import NWBMetaDataEncoder
 
+import os
+
+
+def check_dir(path):
+    if not os.path.exists(path):
+        os.makedirs(path)
+        # FILE_ATTRIBUTE_HIDDEN = 0x02
+        # ret = ctypes.windll.kernel32.SetFileAttributesW(path, FILE_ATTRIBUTE_HIDDEN)
```
CodyCBakerPhD, line 15:
```suggestion
```

Unsure what this is about

pyflask/manageNeuroconv/manage_neuroconv.py
```diff
@@ -5,6 +5,15 @@
 import json
 from neuroconv.utils import NWBMetaDataEncoder
 
+import os
+
+
+def check_dir(path):
+    if not os.path.exists(path):
+        os.makedirs(path)
```
CodyCBakerPhD, line 13: I would recommend familiarizing yourself with `Path` objects from `pathlib` (included in Python standard)

Sometimes the choice between that and `os.path` is entirely personal preference, but this helper function shows how `pathlib` can simplify things all around

Specifically, let's say you have a `Path(folder_path: str)` called `my_path`, you can simply call `my_path.mkdir(exists_ok=True)` and it will do exactly what this helper function does

pyflask/manageNeuroconv/manage_neuroconv.py
CodyCBakerPhD, line 92:
```suggestion
        stub_subfolder = (nwbfile_path.parent / ".stubs").mkdir(exists_ok=True)
        preview_path = stub_subfolder / nwbfile_path.name
```
See above comment regarding `pathlib` - this is what it would look like refactored as long as `nwbfile_path` is an instance of a `Path` object

pyflask/manageNeuroconv/manage_neuroconv.py
```diff
@@ -71,15 +80,29 @@ def convert_to_nwb(info: dict) -> bool:
...
+        dirname = os.path.join(os.path.dirname(nwbfile_path), ".stubs")
+        nwbfile_path = os.path.join(dirname, filename)
+        check_dir(dirname)
+
     converter = instantiate_custom_converter(info["source_data"])
 
-    converter.run_conversion(
+    # Assume all interfaces have the same conversion options for now
+    options = (
+        {interface: {"stub_test": info["stub_test"]} for interface in info["source_data"]} if runStubTest else None
```
CodyCBakerPhD, line 98: Hmmm; works for now but I actually don't think all interfaces support the option

To check ahead of time the official way, you'd want to see if it's in the schema from `converter.get_conversion_options_schema()`

pyflask/manageNeuroconv/manage_neuroconv.py
garrettmflynn: Was part of the StackOverflow answer for creating hidden files but required an additional dependency. Just forgot to remove the comment. Thanks for catching this!

pyflask/manageNeuroconv/manage_neuroconv.py
garrettmflynn: Added a quick check for now. Assuming we'd want to move the selection to the frontend in the future?

CodyCBakerPhD: That will be something to think about. Quite possibly

pyflask/manageNeuroconv/manage_neuroconv.py
```diff
@@ -5,6 +5,14 @@
 import json
 from neuroconv.utils import NWBMetaDataEncoder
 
+from pathlib import Path
+import os
+
+
+def check_dir(path):
+    if not os.path.exists(path):
+        os.makedirs(path)
```
CodyCBakerPhD, line 14:
```suggestion
```

This isn't being used any more, right?

CodyCBakerPhD (approved):

Issue #79, opened 2023-03-27
DANDI upload button


CodyCBakerPhD:
After the files are created from a pipeline (or, perhaps before hitting the run conversion button), we should ask if the user wants to upload the files to DANDI

If they say "yes", then we need to ask them to copy/paste the API token into the app so we can set it as an environment variable

Beyond that it's just `neuroconv.tools.data_transfers.automatic_dandi_upload` on the backend, so should be super easy to add

CodyCBakerPhD:
Done in #86

PR #86, opened 2023-03-28
Upload converted files to DANDI

garrettmflynn:
This PR is intended to address #79 and provide users with the ability to upload their datasets to DANDI.

Currently, there are a few issues to solve before this can be merged:
1. Uploading the converted file fails because "1 out of 1 files were found not containing all necessary metadata: /Users/garrettflynn/Desktop/GUIDE/conversions/converted.nwb"
2. Boolean values are not yet rendered with the JSONSchemaForm

### DIFF
--- a/pyflask/apis/apiNeuroConv.py
+++ b/pyflask/apis/apiNeuroConv.py
@@ -1,6 +1,6 @@
 from flask_restx import Namespace, Resource, reqparse
 from namespaces import get_namespace, NamespaceEnum
-from manageNeuroconv import get_all_interface_info, get_source_schema, get_metadata_schema, convert_to_nwb
+from manageNeuroconv import get_all_interface_info, get_source_schema, get_metadata_schema, convert_to_nwb, upload_to_dandi
 from errorHandlers import notBadRequestException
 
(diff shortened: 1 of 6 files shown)


CodyCBakerPhD:
As discussed, using latest NeuroConv `main` plus that one extra conditions for validating `age` +/- `date_of_birth` for subjects should be the only piece missing

CodyCBakerPhD:
@garrettmflynn Ready for review?

garrettmflynn:
Yes!

CodyCBakerPhD:
It works, and fine as-is for now

In modular (non-chained) PRs, can you make the following changes?

a) I can't change the 'staging' flag (boolean)

b) I can't change `cleanup` flag - also might want to adjust it at the NeuroConv level to specify cleanup of the DANDI folder only, but leave the original NWB files untouched...

c) You can skip the `Version` argument, I need to review if we even need it in the function itself (all uploads go to 'draft' even if there's a published version available, and all it downloads is the `metadata.yml` file anyway...)

d) We don't need to ask the user to specify the `Dandiset Folder Path`  - it's fine to set it to be the parent of the NWB data folder

CodyCBakerPhD (approved):

PR #88, opened 2023-03-28
Resolve correct icon for Electron builds

### LINKED ISSUE #87: system icon
image

This should be the GUIDE icon and should say NWB GUIDE

garrettmflynn:
fix #87

Since it's proven to be quite difficult to actually resolve a custom icon on a Mac development build, ensure that the built version of the NWB GUIDE application has the correct icon by running:

```
npm run build-[platform]
```

Where `platform` is either `win`, `mac`, or `linux` based on the desired target OS. 

You'll be able to find the built project in the `/dist` folder.

I can work more on resolving the icon in development for Mac—but I assume it might not be consequential if we can resolve it for the final build.


garrettmflynn:
I'd recommend testing this on the branch for #91, which fixes #90 (uncovered here) to ensure that the frontend is resolved properly.

bendichter:
The icon looks right, but double-clicking it gives me this error:

image

garrettmflynn:
@bendichter See #89.

CodyCBakerPhD:
So... is this still in progress? Draft? Stuck due to the M1 PyInstaller issue?

garrettmflynn:
Technically completed as much as we can since Mac generally doesn't like resolving the icons during development—though the Mac Pyinstaller issues are downstream of this.

When you build on Mac, you now get the right NWB GUIDE icon. You just also get an error about not having the Python server available.

CodyCBakerPhD:
Ah, but do you need the changes from the PR to make that a possibility through the Mac build?

CodyCBakerPhD:
> If we can push a working icon resolution without a functional Mac build, I'd suggest merging this.

Yeah, that's fine. It can be a known disconnect between dev and non-dev builds

PR #91, opened 2023-03-28
Fix frontend asset resolution for Electron builds

### LINKED ISSUE #90: Electron build broken with unresolved assets
When running `npm run build-[platform]` to create a production build of NWB GUIDE, frontend assets are unable to be resolved.
Screenshot 2023-03-28 at 10 45 16 AM

garrettmflynn:
fix #90


garrettmflynn:
Yep, I'll submit ain independent PR. 

Should I also hold off on renaming `src` to `frontend`? Semantically, moving the `dist` to `frontend/dist` rather than `src/dist` makes more sense in my mind—but it's just a nitpicky issue.

PR #92, opened 2023-03-28
Reload development server while maintaining the chosen page

garrettmflynn:
This PR allows you to reload at the same page you left off when changes are detected by the development server (`npm run dev:app`).

Generally, this also modifies the web distribution of this site to be more amenable to users, who can reload the browser while maintaining their progress.

> **Note:** This PR has **not** solved the issue of resolving progress files for Guided Mode at any point in the workflow, as the Guided Home screen currently manages all aspects of progress loading.


index.html
```diff
@@ -29,8 +29,8 @@
         stylesheet.href = "./dist/style.css"
         document.head.insertAdjacentElement('afterbegin', stylesheet) // ensure css is least priority
 
-        const url = "./dist/index.js" // avoid erroring when not present
-        await import(url /* @vite-ignore */) // Load packaged frontend for Electron
+        const dist = "./dist/index.js"
+        await import(dist /* @vite-ignore */) // Load packaged frontend for Electron
```
CodyCBakerPhD, line 33: This simply changed the name of the variable in JS right? It's not actually changing functionality?

src/stories/Dashboard.js
```diff
@@ -261,12 +287,16 @@ export class Dashboard extends LitElement {
         return acc
       }
 
-
+  #first = true
```
CodyCBakerPhD, line 290: Just curious, what are `#` in JS?

CodyCBakerPhD (approved):

src/stories/Dashboard.js
garrettmflynn: They are private class features (not accessible on instances nor within extended classes): https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields.

Generally, I use them for things that are only internal references and not intended for dev interaction.

CodyCBakerPhD: Ahh very interesting!

Sadly Python classes have no such core functionality; instead it is convention that if a class variable is to intended to be private, to append `_` at the start of the name, such as

```python
class Dashboard(LitElement):
    _first: bool = True
```

It's just a convention though, and not everyone always follows it to be truthful

CodyCBakerPhD:
Re-triggering actions, which are having problems today

PR #93, opened 2023-03-28
Register global project metadata

### LINKED ISSUE #70: Collect Project-wide Metadata
After the page where the user starts a new project, and gives it a name, we can collect some metadata that will apply to all subsequent sessions of that experiment

These include

- Institution
- Lab name
- Experimenters (list of names)
- Experiment description
- Related publications (list of DOI)

and also maybe

- protocol?
- surgery?
- virus?
- stimulus_notes?

and then perhaps we can simply as the question 'did you use the same types of subjects for all sessions of this experiment?; and if 'yes', then follow up with experiment-wide subject info like

- subject species
- subject description
- genotype
- strain
- sex (not required, could vary over sessions; but can also be the same)

Anyway, specifying these will not remove them from the session-specific metadata form, but the values here should serve as the auto-populated default in case the user needs to alter them in some way for a specific session

garrettmflynn:
fix #70 

This PR collects project-wide metadata that is merged with the displayed values on the Metadata page.

To accommodate this feature most effectively, several changes were made to the save file format—so old pipelines will not load properly.

It's worth noting that **changes to the project-wide metadata will not update after loading the Metadata page**, as there's currently no way to disentangle whether a value was set from the project metadata, the resolved values from `neuroconv`, or by an NWB GUIDE user's interactions.


CodyCBakerPhD:
> To accommodate this feature most effectively, several changes were made to the save file format—so old pipelines will not load properly.

Thanks for letting me know lol

CodyCBakerPhD:
(c) also include experiment description and protocol on the global metadata side

(d) ignore source script and source script file name, neither is relevant here

CodyCBakerPhD:
(e) Maybe include keywords as well at the global level

CodyCBakerPhD:
image

Hmmmm

garrettmflynn:
Looks like I had a few of the entries on the wrong level of the JSON Schema so they never showed up. Pushed an update for that.

c-e should be fixed. Can you confirm?

Additionally, the Age description is rendering properly on the current version for me. How about you? It may have just needed to be parsed as HTML.

CodyCBakerPhD:
Final product looks great for everything I set once at the outer level

image

CodyCBakerPhD:
Age text box is still a problem for me but I'll just raise it as a separate lower priority issue to get this through
image

CodyCBakerPhD (approved):

PR #96, opened 2023-03-29
Validate requirements using the JSON Schema Form

garrettmflynn:
This PR allows for validating requirements using the JSON Schema Form and implements this feature across the pages that use it.

This is one of two steps to ensure that we have the appropriate metadata for the DANDI conversion (#86)


CodyCBakerPhD:
@garrettmflynn So sorry, but the usual resolution is necessary again lol

I'd do it myself but I'm scared I might break something that I don't understand 😨

CodyCBakerPhD:
a) session start time is not flagged as required for any interface selection, but it should be

Especially if I select only Phy, there is no default fetched so I can't even make the NWB file


b) DOB did not auto-populate when I specified age, even though both are marked as required

If auto-filling one when the other is entered is too difficult for the front-end, is it at least possible to hide one or the other red stars when the other is specified in the form?

CodyCBakerPhD:
> Would you like me to impose this constraint arbitrarily from the frontend?

For now let's do what's needed to clear the current chain of PRs, but yes the burden should certainly ultimately be on NeuroConv

garrettmflynn:
Fixed the `session_start_time` issue. 

BTW you'll need to check out #97 to see the conditional requirement behavior noted for age and DOB. Will also be implementing the fix noted above there.

garrettmflynn:
@CodyCBakerPhD Actually it's even more complicated, I need updates from #101 to get this done if you'd like it to change on a per-change basis rather than change the nature of the check on submission (which #97 does)

CodyCBakerPhD:
@garrettmflynn OK so should I table this until after #101 and #97?

garrettmflynn:
That would be wonderful. Should be straightforward to fix otherwise.

CodyCBakerPhD:
@garrettmflynn OK other PRs have been cleared up now - just some conflicts to resolve here now

garrettmflynn:
Fixed those conflicts and generalized the existing system to handle all the NWBFile metadata. 

Let me know if things break. Otherwise would be happy to make some styling changes to alerts if we need to before merging!

CodyCBakerPhD:
Hmm.. going through an entire workflow, I'm unable to specify the source data fields

Screenshot 2023-04-06 at 10 34 25 AM

garrettmflynn:
Alright fixed this issue, which ignored the top-level properties because onlyRequired was true and those weren't required.

New changes also allow for gracefully handling suggestions with different levels of importance from NWB Inspector (critical errors or others) while ensuring that linked properties are always resolved correctly (e.g. age and DOB).  Linked properties, however, can still have non-critical suggestions from the Inspector (e.g. a non-standard age can be assigned). 

If non-critical errors remain when the user attempts to move to the next page, a pop-up appears that explicitly asks them if they'd like to move forward given these errors.

CodyCBakerPhD:
This is looking really good, and really close

Couple things of importance blocking this from being finished are...

1) DOB shows for me as 'conditionally required'

image

even when age is specified

image

and since it's red it won't let me go to the next page.

Is it possible to make the linked ones follow an 'at least 1 is specified' condition?



2) From

> New changes also allow for gracefully handling suggestions with different levels of importance from NWB Inspector (critical errors or others) while ensuring that linked properties are always resolved correctly (e.g. age and DOB). Linked properties, however, can still have non-critical suggestions from the Inspector (e.g. a non-standard age can be assigned).
> 
> If non-critical errors remain when the user attempts to move to the next page

These are blocking problems because it means we might make an NWB file that can't be uploaded to DANDI.

The source of the issue I think these might be linked, and also because I forgot to mention we should use the DANDI configuration of importance, which is tricky to do for our single-check use case (easy to do following the normal usage of the Inspector as a catch-all)

Can you adjust the check import to follow this logic?

```python
from nwbinspector.nwbinspector import configure_checks, load_config

dandi_check_list = configure_checks(config=load_config(filepath_or_keyword="dandi"))
dandi_check_registry = {check.name: check for check in dandi_check_list}
```

and fetch each check from the `dandi_check_registry` instead of the `__dict__` of the `nwbinspector`?

garrettmflynn:
Thanks for catching the first part! The linked properties _should_ generally check whether all onInput checks for a particular property and all others linked to it have no errors—and, in this specific case with the NWB Validator, this is handled as an 'at least 1 is specified' condition.

Looks like I wasn't awaiting an asynchronous call to the server, so the property was conditionally evaluating to still required when it actually wasn't. Should work now!

I tried to switch to the DANDI checks but I'm getting the following errors when mapping the list to a dict: 
```
 File "/Users/garrettflynn/Documents/Github/nwb-guide/pyflask/manageNeuroconv/manage_neuroconv.py", line 79, in <dictcomp>
    dandi_check_registry = {check.name: check for check in dandi_check_list}
AttributeError: 'function' object has no attribute 'name'
```

garrettmflynn:
Updated to __name__ and it works. Though there don't appear to be any changes to the warnings.

CodyCBakerPhD:
> Updated to __name__ and it works. Though there don't appear to be any changes to the warnings.

1) ensure `nwbinspector` is up to date, `pip install -U nwbinspector`

2) check manually with Python that the following holds

```python
> dandi_check_registry["check_subject_sex"].importance
<Importance.CRITICAL: 2>
```

CodyCBakerPhD:
Though this led to discovery of https://github.com/NeurodataWithoutBorders/nwbinspector/pull/358

pyflask/manageNeuroconv/manage_neuroconv.py
```diff
@@ -74,9 +75,12 @@ def get_check_function(check_function_name: str) -> callable:
     """
     Function used to fetch an arbitrary NWB Inspector function
     """
-    check_function: callable = nwbinspector.__dict__.get(check_function_name)
+    dandi_check_list = configure_checks(config=load_config(filepath_or_keyword="dandi"))
+    dandi_check_registry = {check.__name__: check for check in dandi_check_list}
+
+    check_function: callable = dandi_check_registry.get(check_function_name)
```
CodyCBakerPhD, line 81:
```suggestion
    check_function: callable = dandi_check_registry[check_function_name]
```

If you swap this from `get` you can remove the `if .. is None; raise ...` below, since they do the same thing

garrettmflynn:
So I've validated that I have importance values based on the DANDI configuration results and the method you provided [above](https://github.com/NeurodataWithoutBorders/nwb-guide/pull/96#issuecomment-1502159237).

Not sure if it's related to the issue you uncovered, but the function output is not aligned with the function (DANDI) properties:
Screenshot 2023-04-10 at 11 57 46 AM

When spit out of the function, I still have an Importance.BEST_PRACTICE_SUGGESTION value.

CodyCBakerPhD:
> When spit out of the function, I still have an Importance.BEST_PRACTICE_SUGGESTION value.

Ahh, this is a known separate issue; the cause is more complicated and has not yet been fully solved.

Our way around it in the rest of the inspector is to correct the inspector message post-return like in https://github.com/NeurodataWithoutBorders/nwbinspector/blob/dev/src/nwbinspector/nwbinspector.py#L575-L576

garrettmflynn:
@CodyCBakerPhD Alright this is properly gating improper age values now.

CodyCBakerPhD:
Looks great!

CodyCBakerPhD (approved):

PR #98, opened 2023-03-30
Simplify search by using neuroconv keywords and reformatting returned…

garrettmflynn:
This PR uses `neuroconv`-provided keywords to condition the results of the search bar.

It includes frontend and backend changes that make it easier to coordinate future requests between them.

The `test_get_all_interfaces` was also updated to require the new `label` and `keywords` properties.


CodyCBakerPhD:
@garrettmflynn It's looking much better!

I also discovered I can indeed edit those text fields, though it's not a very obvious interaction (the text is also very small)

However, there's some issues with the key selector that umm.. maybe we can ignore for a little while, but it did actually essentially break my pipeline without an obvious way for me to fix it. So maybe we should resolve them right away

image

and hitting 'next' fails to take me to the next page, likely because of that aberrant first interface

I need, at a minimum,

1) a 'remove' button for each interface in the list, in case I added it by accident or messed something up with it



and would prefer to have (related to how this awful mistake of mine occurred...)

2) hitting 'Enter' should complete the text entry, just like clicking outside the box does

What happened instead is it added a newline character, which shouldn't be allowed anyway in the keys

3a) a sequence of delete operations that yield an empty string should either bounce back to their original state or the interface should be flagged for having a bad key

3b) I can't even click on the text box of the empty string to add something to it, so we need an empty text box in that case

garrettmflynn:
@CodyCBakerPhD Alright I have tried to resolve the issues identified above using your suggested solutions. Let me know how this looks!

garrettmflynn:
This version will also run a complete conversion now. Previously, the new keys (only interface class names) were actually being used for future steps

CodyCBakerPhD:
Seems to be breaking one of the core tests 🤔

CodyCBakerPhD:
That said, all of the requested functionality seems to be working well for the interface selector part

CodyCBakerPhD (approved):

PR #101, opened 2023-03-31
Validate metadata during entry using nwbinspector

garrettmflynn:
fix (1) and (2) on #93 

This PR introduces a way to validate metadata against the `nwbinspector` and implements this feature on a project-wide metadata form.

### DIFF
--- a/pyflask/manageNeuroconv/manage_neuroconv.py
+++ b/pyflask/manageNeuroconv/manage_neuroconv.py
@@ -4,9 +4,13 @@ from neuroconv import datainterfaces, NWBConverter
 
 import json
 from neuroconv.utils import NWBMetaDataEncoder
+import nwbinspector
+from pynwb.file import NWBFile, Subject
+from nwbinspector.nwbinspector import InspectorOutputJSONEncoder
 
 from pathlib import Path
 import os
+from datetime import datetime
 
 
 def get_all_interface_info() -> dict:
@@ -67,6 +71,39 @@ def get_metadata_schema(source_data: Dict[str, dict]) -> Dict[str, dict]:
     schema = converter.get_metadata_schema()
     metadata = converter.get_metadata()
     return json.loads(json.dumps(dict(results=metadata, schema=schema), cls=NWBMetaDataEncoder))
+
+
+
+class objectview(object):
+    def __init__(self, d):
+        self.__dict__ = d
+
+def validate_metadata(parent: dict, function: str) -> dict:
+    """
+    Function used to validate data using an arbitrary NWB Inspector function
+    """
+    fn = nwbinspector.__dict__.get(function)
+    if fn is None:
+        raise ValueError(f"Function {function} not found in nwbinspector")
+    
+    if "subject" in function and "subject_exists" not in function:
+        cls = Subject
+        default_parent = {}
+    else:
+        cls = NWBFile
+        default_parent = dict(
+            session_description="This is a description",
+            identifier="00001",
+            session_start_time=datetime.now(),
+        )
+
+    result = fn(cls(
+        **default_parent,
+        **parent
+    ))
+
+    return json.loads(json.dumps(result, cls=InspectorOutputJSONEncoder))
+
 
 
 def convert_to_nwb(info: dict) -> str:
--- a/src/stories/pages/guided-mode/setup/GuidedNewDatasetInfo.js
+++ b/src/stories/pages/guided-mode/setup/GuidedNewDatasetInfo.js
@@ -1,11 +1,10 @@
 
 
 import { html } from 'lit';
+import { baseUrl, notyf } from '../../../../globals.js';
 import { hasEntry, update } from '../../../../progress.js';
-import { Page } from '../../Page.js';
 import { JSONSchemaForm } from '../../../JSONSchemaForm.js';
-
-import { notyf } from '../../../../globals.js';
+import { Page } from '../../Page.js';
 
 export class GuidedNewDatasetPage extends Page {
 
@@ -23,21 +22,21 @@ export class GuidedNewDatasetPage extends Page {
     onNext: async () => {
 
       const globalState = this.info.globalState.project
-
+      
       // Check validity of project name
       const name = this.state.name
       if (!name) {
-
         const message = "Please enter a project name."
         notyf.open({
           type: "error",
           message: message,
           duration: 7000,
         });
-
-        return
       }
 
+      this.form.validate()
+
+      if (!name) return
 
       // Check if name is already used
       // Update existing progress file
@@ -51,7 +50,6 @@ export class GuidedNewDatasetPage extends Page {
             message: message,
             duration: 7000,
           });
-
           return
         }
     }
@@ -60,7 +58,6 @@ export class GuidedNewDatasetPage extends Page {
     Object.assign(globalState, this.state)
 
     this.onTransition(1)
-
   }
 }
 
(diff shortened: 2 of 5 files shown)


pyflask/manageNeuroconv/manage_neuroconv.py
```diff
@@ -69,6 +73,35 @@ def get_metadata_schema(source_data: Dict[str, dict]) -> Dict[str, dict]:
     return json.loads(json.dumps(dict(results=metadata, schema=schema), cls=NWBMetaDataEncoder))
 
 
+class objectview(object):
+    def __init__(self, d):
+        self.__dict__ = d
+
+
+def validate_metadata(parent: dict, function: str) -> dict:
```
CodyCBakerPhD, line 81:
```suggestion
def validate_metadata(nwbfile_metadata: dict, check_function_name: str) -> dict:
```

Some suggestions for better readability of the code; we nearly always use full words that are as descriptive as possible for Python code

Only exceptions are usually built-in names, which I will point out that `cls` is actually one of those (used as the first argument to class methods)

pyflask/manageNeuroconv/manage_neuroconv.py
```diff
@@ -69,6 +73,35 @@ def get_metadata_schema(source_data: Dict[str, dict]) -> Dict[str, dict]:
...
+    """
+    Function used to validate data using an arbitrary NWB Inspector function
+    """
+    fn = nwbinspector.__dict__.get(function)
```
CodyCBakerPhD, line 85:
```suggestion
    nwbfile_metadata = nwbfile_metadata or dict(session_description="This is a description",
            identifier="00001",
            session_start_time=datetime.now(),
    )
    fn = nwbinspector.__dict__.get(function)
```

Suggested logical design change - even the subject-level checks are technically run upon the NWBFile object, so you will always need to make a 'placeholder' NWBFile object

Only different with the Subject checks is they will also make a `Subject` that gets added to that placeholder NWB file

pyflask/manageNeuroconv/manage_neuroconv.py
CodyCBakerPhD, line 81: I would recommend splitting this into two functions - one for NWBFile checks, and one for subject level checks

This will also help scale to other checks on other metadata fields too

In this case, the first argument could just be called `metadata` in each function, as long as the name of each function indicates the intended target

CodyCBakerPhD: Also, `_name` is more honest to the expected `str` usage; it is not uncommon to see `function: callable` in Python, but that's not how this function signature is setup (nor would I recommend it either, this approach seems fine)

pyflask/manageNeuroconv/manage_neuroconv.py
CodyCBakerPhD, line 85:
```suggestion
    check_function: callable = nwbinspector.__dict__.get(check_function_name)
```
I might make a mapping like this in the Inspector sometime - technically there is the `available_checks`, but it is a list so you'd have to do search instead of direct reference like this

pyflask/manageNeuroconv/manage_neuroconv.py
```diff
@@ -69,6 +73,35 @@ def get_metadata_schema(source_data: Dict[str, dict]) -> Dict[str, dict]:
...
+    if fn is None:
+        raise ValueError(f"Function {function} not found in nwbinspector")
+
+    if "subject" in function and "subject_exists" not in function:
```
CodyCBakerPhD, line 89: It is interesting that this appears to work since all known checks that run on subjects (except for existence) do indeed have 'subject' in their name

But that's not really intentional - instead, I would recommend using `isinstance` against the `check_function.neurodata_type` attribute (yeah, functions in Python can have attributes lol; let me know if curious why that is) to see if it matches the `NWBFile` or `Subject` class

CodyCBakerPhD, line 89: It is interesting that this appears to work since all known checks that run on subjects (except for existence) do indeed have 'subject' in their name

But that's not really intentional - instead, I would recommend using `isinstance` against the `check_function.neurodata_type` attribute (yeah, functions in Python can have attributes lol; let me know if curious why that is) to see if it matches the `NWBFile` or `Subject` class

pyflask/manageNeuroconv/manage_neuroconv.py
```diff
@@ -69,6 +73,35 @@ def get_metadata_schema(source_data: Dict[str, dict]) -> Dict[str, dict]:
...
+        cls = Subject
```
CodyCBakerPhD, line 90:
```suggestion
        nwb_class = Subject
```
`cls` is a reserved word; not that you can't technically use it here, just not conventional

src/stories/pages/guided-mode/setup/GuidedNewDatasetInfo.js
```diff
@@ -187,9 +184,37 @@ export class GuidedNewDatasetPage extends Page {
...
+
+        const message = { parent }
+
+        if (noCheck.includes(name)) return
+        else if (name === 'related_publications') message.function = 'check_doi_publications'
+        else if (name === 'experimenter') message.function = 'check_experimenter_form'
+        else if (path.slice(-1)[0] === 'Subject') {
+          message.function = 'check_subject_' + name
+          if (name === 'species') message.function += '_form'
+        }
```
CodyCBakerPhD, line 206: For this mapping, is it ever convenient to define something like this in an external YML/JSON file and load it in the JS instead of hard-coding it in pure code?

garrettmflynn: It wouldn't hurt—and the rules might be more readable. 

Generally, I hard-code things like this for the initial validation of an approach and generalize to an external JS module / YML / JSON if I'm checking in more than one place.

pyflask/manageNeuroconv/manage_neuroconv.py
```diff
@@ -69,6 +73,35 @@ def get_metadata_schema(source_data: Dict[str, dict]) -> Dict[str, dict]:
...
+    if "subject" in function and "subject_exists" not in function:
+        nwb_class = Subject
+        default_parent = {}
+    else:
+        cls = NWBFile
+        default_parent = dict(
+            session_description="This is a description",
+            identifier="00001",
+            session_start_time=datetime.now(),
+        )
```
CodyCBakerPhD, line 98: Oh, BTW it occurred to me over the weekend that it might just be easiest in each neurodata-type specific function to use the PyNWB Mock testing classes with a single field override, which is essentially what we're doing here

Example:

```python
from pynwb.testing.mock.file import mock_NWBFile  # also mock_Subject

testing_nwbfile = mock_NWBFile()  # or inject a single field override in this init
```

CodyCBakerPhD: This would also generalize to other future cases of this really well

garrettmflynn: Great adding now!

garrettmflynn:
@CodyCBakerPhD Finished adding your suggestions. Should be good to review!

pyflask/manageNeuroconv/manage_neuroconv.py
```diff
@@ -72,6 +77,65 @@ def get_metadata_schema(source_data: Dict[str, dict]) -> Dict[str, dict]:
...
+    Function used to fetch an arbitrary NWB Inspector function
+    """
+    check_function: callable = nwbinspector.__dict__.get(check_function_name)
+    if check_function is None:
+        raise ValueError(f"Function {function} not found in nwbinspector")
+
+    return check_function
+
+
+def validate_subject_metadata(subject_metadata: dict, check_function_info: str or callable):
```
CodyCBakerPhD, line 91: Proper `or` typing in annotations is actually done with `typing.Union`; as in `check_function_name: Union[str, callable]`

Why is this either `str` or `callable`? In what case is it the actual `callable`? If it can be limited to just `str`, which it looks like it can below, then we probably should until it's necessary

CodyCBakerPhD, line 91:
```suggestion
def validate_subject_metadata(subject_metadata: dict, check_function_info: str or callable) -> Union[None, InspectorMessage, List[InspectorMessage]]:
```

And ditto for the other validate helper

Note will have to import like `from nwbinspector.register_checks import InspectorMessage`

CodyCBakerPhD:
As I test this out in the actual app, should it be preventing me from moving to the next page if there's a metadata issue? Or should it be allowing non-valid entries to these fields of the metadata form to begin with?

Currently I can enter lots of non-compliant stuff in these fields and it allows me to proceed

src/stories/pages/guided-mode/setup/GuidedNewDatasetInfo.js
CodyCBakerPhD: I'd strongly recommend it since it would also allow non-JS familiar people to easily read and modify the ruleset for which checks run on which form fields

CodyCBakerPhD:
> As I test this out in the actual app, should it be preventing me from moving to the next page if there's a metadata issue? Or should it be allowing non-valid entries to these fields of the metadata form to begin with?
>
> Currently I can enter lots of non-compliant stuff in these fields and it allows me to proceed

Nevermind, as per meeting today I checked and ensured that functionality is working well on the project level metadata side

src/stories/pages/guided-mode/setup/GuidedNewDatasetInfo.js
garrettmflynn: See src/stories/pages/guided-mode/setup/validation.json

CodyCBakerPhD:
@garrettmflynn This is really, really great. Can't wait to see it integrated with all the other metadata fields!

CodyCBakerPhD (approved):

PR #103, opened 2023-03-31
Remove dandiset_folder_path

### LINKED ISSUE #102: DANDI Upload improvements
It works, and fine as-is for now

In modular (non-chained) PRs, can you make the following changes?

a) I can't change the 'staging' flag (boolean)

b) I can't change `cleanup` flag - also might want to adjust it at the NeuroConv level to specify cleanup of the DANDI folder only, but leave the original NWB files untouched...

c) You can skip the `Version` argument, I need to review if we even need it in the function itself (all uploads go to 'draft' even if there's a published version available, and all it downloads is the `metadata.yml` file anyway...)

d) We don't need to ask the user to specify the `Dandiset Folder Path`  - it's fine to set it to be the parent of the NWB data folder

_Originally posted by @CodyCBakerPhD in https://github.com/NeurodataWithoutBorders/nwb-guide/issues/86#issuecomment-1492140092_

garrettmflynn:
fix #102 (d)

### DIFF
--- a/pyflask/manageNeuroconv/manage_neuroconv.py
+++ b/pyflask/manageNeuroconv/manage_neuroconv.py
@@ -116,8 +116,6 @@ def upload_to_dandi(
     dandiset_id: str,
     nwb_folder_path: str,
     api_key: str,
-    dandiset_folder_path: Optional[str] = None,
-    version: Optional[str] = None,
     staging: Optional[bool] = None,  # Override default staging=True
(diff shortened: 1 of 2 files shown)


CodyCBakerPhD (approved):

CodyCBakerPhD:
Had to debug a could arguments to the NeuroConv helper as they were being overridden to incorrect default values

Working well now 👍

PR #109, opened 2023-04-03
Remove all unused python helpers

garrettmflynn:
This PR removes all unused Python code from the backend.

At some point, we'll also want to pare down the requirements to remove those that are unused. @CodyCBakerPhD Is there a simple way to do this in Python?


CodyCBakerPhD:
> At some point, we'll also want to pare down the requirements to remove those that are unused. @CodyCBakerPhD Is there a simple way to do this in Python?

Simplest way is to uninstall one of the named parent packages from the requirements and try it out (usually by running the test suite)

I can take on that task since I have more global familiarity with the the dependency stack

CodyCBakerPhD (approved):

PR #121, opened 2023-04-12
Visually group linked properties

garrettmflynn:
This PR organizes linked properties together visually so their association is clearer. It also represents incompleteness as a single (now red) indicator rather than two yellow indicators.


CodyCBakerPhD:
Hmmm not really sure how to resolve the conflicts here with how the previous priority+alphabetical ordering affected things

garrettmflynn:
Merged! These changes now sort linked property boxes between required and non-required properties.

CodyCBakerPhD:
Chromatic suddenly failing on this branch: https://github.com/NeurodataWithoutBorders/nwb-guide/actions/runs/4701439623/jobs/8337462283

Also, do you think it would be possible to have some kind of default render of the metadata form that could show up in chromatic? Maybe as a pre-step to the call to chromatic in the CI, have a testing utility that makes API calls as necessary to generate such a page on the fly?

garrettmflynn:
Will see what went wrong. Similar to our discussion in #122 (https://github.com/NeurodataWithoutBorders/nwb-guide/pull/122#discussion_r1165988060), we'll want to somehow link to a dummy JSON Schema file—if we want to maintain any relation to the actually rendered information.

CodyCBakerPhD:
> a dummy JSON Schema file—if we want to maintain any relation to the actually rendered information.

Yeah, a dummy JSON wouldn't be a bad idea to have setup for Chromatic since that is effectively the testing suite for the front-end

Also makes more sense for the test instance there to be fully hard-coded for readability

Do you think it would be possible to include that in this PR?

garrettmflynn:
Done. Just tested a complete Chromatic run too—though we'll need to run that again to see the rendered dummy metadata.

CodyCBakerPhD:
Working great, the new chromatic looks beautiful

CodyCBakerPhD (approved):

PR #122, opened 2023-04-13
Split JSON Schema across Generated Pages

### LINKED ISSUE #72: First page of metadata form should only cover NWB file
Right now the metadata form is rendered in its entirety on a single page of the pipeline; I think it would be a better user experience to not overwhelm with so much on one page, and split it into thematic sections according to the outer level of the NeuroConv `metadata` object

That is,

Page 1: NWB file
Page 2: Subject
Page 3: Ecephys (for current MVP constraints)

garrettmflynn:
fix #72

This PR begins to address the issue of splitting JSON Schema rendering across several pages. It involves the programmatic addition of new pages (1) at initialization for Project Setup and (2) when receiving schemas for NWB Metadata. 

At present, this is a minimal solution to the problem of generating multiple pages from a single JSON Schema. Additional improvements may include:
1. Rendering the new pages as sub-pages of the original page, rather than sibling pages.
 - Smaller capsules beneath the original
 - Another layer of nesting in the navigation sidebar
2. Improvements to page/state reloading, where new pages are sometimes flagged incorrectly as active

Are there any other things we'd like to implement immediately for better UX?


CodyCBakerPhD:
> Are there any other things we'd like to implement immediately for better UX?

Looks good to me so far, but only time will tell; if I notice anything over time I'll raise an issue to discuss but this looks like a good first step

src/stories/pages/guided-mode/setup/project-metadata.schema.json
CodyCBakerPhD, line 1: While I am not against hard-coding this stuff -yet-again- (it's hard-coded at the nwb-schema level, the PyNWB level, and again at the NeuroConv level already) somewhere at the GUIDE level; would it be possible to move this and similar files to an outer level like `/schemas` to make it easier to find and modify from the top of the package?

Also how robust is that? If I wanted to add a new field to the `NWBFile` level, would it simply be rendered automatically in the forms?

Also also, note NeuroConv has an area very similar if not identical to this located here: https://github.com/catalystneuro/neuroconv/blob/main/src/neuroconv/schemas/metadata_schema.json

If there's any way we can harness NeuroConv to avoid duplication (and make any necessary improvements/adjustments to NeuroConv of course) we should try to do it

garrettmflynn: This is a robust solution. Adding a new property here will automatically render on the Project Setup page. You can test this by running `npm run dev`, visiting that page, and changing the code—which will automatically re-render the page with the changes resolved.

It would be possible for me to grab the JSON from neuroconv (probably a url)—though I'd have to (1) remove the properties I don't want to register as project-wide metadata, as well as (2) add the name property and require it. 

This would be pretty simple. I'd convert this JSON file into a JS file and programmatically implement those changes.

We definitely move these files to a `schemas` folder for better findability.

garrettmflynn: Unfortunately, Github doesn't host files using the correct mimeType (`application/json`, rather than `text/plain`) to import JSON files directly—so even the raw link doesn't work for this purpose. 

However, we could alternatively host the `neuroconv` main branch using Github Pages and grab https://catalystneuro.github.io/neuroconv/src/neuroconv/schemas/metadata_schema.json, which would make it available to JS scripts.

CodyCBakerPhD: I mean, they are also physical files somewhere in the application environment. I could even put together a little API through NeuroConv to conveniently fetch it from the backend - would that make more sense? (also then you don't have potential misalignment with the GitHub version and local NeuroConv version)

garrettmflynn:
Ah right, that would also be fine. I was thinking it'd be difficult because each developer's environment might be located in a different place—but, if we assume the Python backend is always there for devs and users alike, we can simply create a Flask endpoint to serve the file.

src/stories/pages/guided-mode/setup/project-metadata.schema.json
garrettmflynn: Ah right, that would also be fine. I was thinking it'd be difficult because each developer's environment might be located in a different place—but, if we assume the Python backend is always there for devs and users alike, we can simply create a Flask endpoint to serve the file.

CodyCBakerPhD: OK I'll get a dev branch up soon for this - it's something I've been noticing would be nice anyway

CodyCBakerPhD:
Huh, can you take a look at https://github.com/NeurodataWithoutBorders/nwb-guide/actions/runs/4703409213/jobs/8341867829 and see if that matches what we're supposed to be seeing here?

garrettmflynn:
Fixed some merge issues. Also added some extra pre-configured metadata to see two pages.

Since all the form data for the NWB File Metadata page should be rendered as separate pages, I've made it so that it immediately jumps to the next page—so you'll see that you actually end up at the NWBFile page (/guided/metadata/NWBFile).

garrettmflynn:
> OK I'll get a dev branch up soon for this - it's something I've been noticing would be nice anyway
We're going to handle this separately from this PR, right? Just curious if we have other blockers here that need to be fixed.

CodyCBakerPhD:
I suppose no need to hold this up, can you raise an issue though so we don't forget about it?

CodyCBakerPhD:
Though now that I test it again I'm seeing some very strange behavior - the metadata forms appear to be rendering before my selection of interfaces and source data?

garrettmflynn:
Fixed after our meeting today. Interfaces were not being correctly tracked.

garrettmflynn:
@CodyCBakerPhD Should be good to merge in. Removed the changes to the File Metadata page—and actually did have to update how interfaces were being passed (#125) before running a full conversion.

CodyCBakerPhD (approved):

CodyCBakerPhD:
Lookin' good 👍

PR #114, opened 2023-04-10
Create frontend and backend for handling session iteration + metadata extraction

garrettmflynn:
This PR begins to address #78 by creating a frontend page and backend endpoint to handle automatic subject and session source data + metadata detection. It was tested on SpikeGLX data from the `cloud-and-guide-testing-data` S3 bucket.

When testing this PR, make sure to use the latest version of [Feature Branch 377](https://github.com/catalystneuro/neuroconv/pull/377), which allows for using an fstring to detect subjects and sessions.

Note that the current implementation of autodetection funnels the returned results into **one subject and session** to avoid issues with rendering further down the GUIDE workflow, allowing us to separate UI improvements into a separate PR.

PR #125, opened 2023-04-17
Fix conversion behavior which requires interfaces to be passed

garrettmflynn:
Pass interfaces to the final conversion to avoid server-side error. Additionally, block transitions if an error occurs.


CodyCBakerPhD:
No longer needed, specific to #114 being out of date

PR #133, opened 2023-04-19
Hot reload with Electron

garrettmflynn:
This PR allows for immediate reflection of changes to the code in the Electron app (similar to the web version) with minimal load-time.

### DIFF
--- a/main.js
+++ b/main.js
@@ -13,6 +13,11 @@ require('./shortcuts.js')
 
 autoUpdater.channel = "latest";
 
+// Enable hot reload for Electron
+try {
+  require('electron-reloader')(module)
+} catch (_) {}
+
 /*************************************************************
  * Python Process
  *************************************************************/
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,11 +1,11 @@
 {
-  "name": "neuroconv-app",
+  "name": "nwb-guide",
   "version": "0.0.1",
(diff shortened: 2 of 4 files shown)


CodyCBakerPhD (approved):

CodyCBakerPhD:
IDK why but this took me forever to install with `npm ci`

CodyCBakerPhD:
Also can't quite get the reload functionality to work as I would have expected (I change some code on the Flask backend, hit reload, and the change does not apply to the app unless I quite and re-launch)

But if it's working for you go ahead and merge

garrettmflynn:
This is just looking at the JavaScript code. I can see if it will watch Python code as well.

CodyCBakerPhD:
> This is just looking at the JavaScript code.

Ah, that would explain why lol


> I can see if it will watch Python code as well.

Let me know if that works - if so, it could act as a really convenient way of doing certain dynamic imports/installs going into the future

garrettmflynn:
@CodyCBakerPhD So [`electron-reloader`](https://github.com/sindresorhus/electron-reloader) does not allow for manual configuration of what to watch, though their code would be quite simple to replicate ourselves using [`chokidar`](https://github.com/paulmillr/chokidar). 

I've opened a new Issue (#136) where we can talk about the "certain dynamic imports/installs" in more detail and see if we'd like a custom solution.

PR #134, opened 2023-04-19
Add additional cases of JSONSchemaForm to Storybook for Chromatic review

garrettmflynn:
This PR introduces a set of JSONSchemaForm stories that comprehensively show the available functionality of the component.

Global styles from the app are not yet applied—though these are not necessary for a review of the functionality


CodyCBakerPhD:
I see now - I like this approach and think it makes a lot of sense as the closest we can get to 'unit' tests for visualizing different cases of front-end rendering given certain states of the app that we can hard-code as the 'inputs' in each story

CodyCBakerPhD:
Now the question will be, when Ryan adds codecov - is codecov smart enough to estimate coverage of front-end components baed on chromatic stories? I wonder...

CodyCBakerPhD (approved):

PR #139, opened 2023-04-26
Add subject table

### LINKED ISSUE #130: Implement a Subject Metadata page
Insert a Subject Metadata page for specifying subject metadata that persists between sessions. 

Screenshot 2023-04-17 at 12 00 14 PM

Using the returned metadata after the Source Data has been submitted, we can auto-populate with estimated values (if any).

garrettmflynn:
fix #130 

This PR introduces a subject table and related changes that enable this workflow.


CodyCBakerPhD:
A couple notes...

i) group `age` and `date_of_birth` next to each other, or in general the way all fields are grouped on the form for consistency
ii) add units to the column name, maybe in parentheses? Mostly for `weight` -> `weight (kg)`; might need input from @oruebel @rly on how to handle weight, it's normally a free-text string where the unit is added after the number in the string; `kg` is not mentioned in the NWB schema that I can see, but is at [this place in PyNWB](https://pynwb.readthedocs.io/en/stable/pynwb.file.html#pynwb.file.Subject.weight)

CodyCBakerPhD:
iii) When using the programmatic inference, it did not auto-populate all the subjects correctly

CodyCBakerPhD:
iv) In the following screenshot - the numbered itemization of seessions can look awkward for certain types of purely numeric `session_id`'s, such as those that are also datetimes

image

I also don't know that numerical ordering of sessions per subject is necessary; are there other options for formatting that?

CodyCBakerPhD:
v) When I specified a global species, it did not auto-populate in this table

CodyCBakerPhD:
vi) `date_of_birth` selector widget does not allow time, only date (the actual role of the field is an ISO 8601 with full timestamps capability)

vii) Can't remove a row that I added accidentally

garrettmflynn:
For (iii), can you add a screenshot of the current behavior? On the test cases I've used, it seems to work.

As for (iv), we can use a `ul` (unordered list) tag instead of an `ol` tag. This will allow us to do anything for the [list style](https://developer.mozilla.org/en-US/docs/Web/CSS/list-style), though I'd just default to bullets or dashes.

Thanks for reminding me about (v), I still have to link the project and session-specific metadata into this

CodyCBakerPhD:
That's all I can find for now. As far as actual interaction with the table, it's pretty nice and smooth so I do like that. This should work well with the Electrodes table too

CodyCBakerPhD:
Copy/pasting from other sheets into the app also worked flawlessly!

garrettmflynn:
vi) I'll have to add a custom editor so it uses the native pop-up—but that shouldn't be too hard.
vii) Did you try ctrl-clicking? This should show a context menu that allows for addition and removal. The button is just to highlight the affordance of adding a suject

CodyCBakerPhD:
> As for (iv), we can use a ul (unordered list) tag instead of an ol tag. This will allow us to do anything for the [list style](https://developer.mozilla.org/en-US/docs/Web/CSS/list-style), though I'd just default to bullets or dashes.

Unordered is the way to go I think - up to you as what you think looks best (though given that I'm biased towards YAML, I'd say dashes)


> Thanks for reminding me about (v), I still have to link the project and session-specific metadata into this

Do you think we should do that in this PR?

garrettmflynn:
100%. This isn’t really functional without that. I’ll finish that first
thing today.

CodyCBakerPhD:
> vii) Did you try ctrl-clicking? This should show a context menu that allows for addition and removal. The button is just to highlight the affordance of adding a suject

ctrl-cliking allows me to select multiple cells, as I'd expect of a spreadsheets; deleting an entire row however, does not delete the row itself (just leaves a blank row)

garrettmflynn:
> ctrl-clicking allows me to select multiple cells, as I'd expect of a spreadsheets; deleting an entire row however, does not delete the row itself (just leaves a blank row)

delete

So this doesn't work for you? Or are you talking about a different behavior?

CodyCBakerPhD:
> So this doesn't work for you? Or are you talking about a different behavior?

OK, on Windows it's right-clicking.

Two methods to resolve this, which do you think is easier/better?

a) Include instructions for this below the table (if mac/linux, use ctrl-click to delete a row; if windows, use right-click) because otherwise it's not obvious 

b) Add an explicit button on the side  of each row, much like how we remove accidentally added Data Interfaces on the previous screen

CodyCBakerPhD:
Looks like that's my last qualm about it (the thing about removing rows). All the rest feels really good; it's a very nice table element, very interactive and interoperable!

CodyCBakerPhD:
Also tests are breaking, should probably fix that before merging

garrettmflynn:
Alright fixed the tests! 

A description at the bottom will likely be the simplest moving forward, I've already mocked it up. Also, right-clicking seems to work across operating systems—so I've simply noted that at the bottom.

Screenshot 2023-04-28 at 2 17 12 PM

CodyCBakerPhD:
Spotted one thing during Chromatic review, otherwise good to go

CodyCBakerPhD:
Also tried leaving the comment specifically on the Chromatic review, so let me know how that goes or if I should just stick to the GitHub channel for that

garrettmflynn:
Didn't get a notification, but I'll see if I can enable those. That'd be the only downside to Chromatic right now.

garrettmflynn:
Fixed! Just requires a new global metadata field that I didn't add to Chromatic after merging in the new views here.

CodyCBakerPhD (approved):

rly:
Re: weight in kg, let's keep "weight (kg)". In general, we want to encourage SI units for all fields, with maybe a few exceptions. For now, let's append " kg" to the weight entered in this table.

People who are converting data from invertebrates will probably not feel great putting weights in kg, but also, I am pretty sure most people studying invertebrates using neurophysiology are not weighing their animals anyway.

PR #142, opened 2023-04-26
Normalize path registered on browser

garrettmflynn:
This PR ensures that paths are converted to contain forward slashes rather than backslashes


CodyCBakerPhD:
Hmm.. still showing up differently on at least the conversion preview page

image

garrettmflynn:
Fixed. I’d saved the modified path internally but didn’t pass it through
the callback that updates the globalState

CodyCBakerPhD (approved):

PR #144, opened 2023-04-26
Emphasize status indicator

garrettmflynn:
This PR introduces larger status indicators

## Before
Screenshot 2023-04-26 at 4 35 22 PM

## After
Screenshot 2023-04-26 at 4 34 18 PM


CodyCBakerPhD:
Definitely better, but I was thinking about the following more visually distinctive approach

Current color indicator - meaning - desired appearance

Green - No errors or warnings - Green check mark

Yellow - No errors, some warnings - Yellow yield sign

Red - At least one error present - Red X or something similar

White - not started yet? - I don't know if we even need to indicate this; wouldn't everything start in one of the previous states depending on how much global metadata was pre-specified?

garrettmflynn:
Sounds good to me. Now that you mention it, a symbolic approach will be
more accessible and straightforward.

As for the Not Started Yet state, that’s really a “no state provided”
option if there isn’t one specified—as we may not always take advantage of
this feature.

I’d shifted from a transparent bar to a white bar for visual consistency
between all list items, but I’ll remove any representation moving forward
with this approach.

garrettmflynn:
Here are a few screenshots of the latest:
Screenshot 2023-04-27 at 2 19 02 PM
Screenshot 2023-04-27 at 2 18 35 PM

Note that the Accordion component parallels the Instance Manager component—though it uses a darker colored border instead of a background beneath the icon.

CodyCBakerPhD:
It looks great! New bug though; not sure how this cropped up but even though I have subject age specified, it says `date_of_birth` is still required

garrettmflynn:
Good catch. This became an issue since I'm now checking whether a property is required during validateOnChange checks—and was counting both of the linked properties as such.

CodyCBakerPhD:
Looks fantastic! Also, huge thanks for the Chromatic stories that do a really good job showcasing how this looks in the app!

CodyCBakerPhD (approved):

PR #147, opened 2023-05-01
Remove extra field (sessions) from subject metadata merge with session metadata

garrettmflynn:
This PR ensures that the sessions field from the subject metadata doesn't end up corrupting the session metadata passed to the Python backend for conversions.


CodyCBakerPhD:
This seems like it would be a higher importance than the other PRs; some conflicts to resolve then g2g on my side

Is there any way you can think of to write a test (with Python or as a Chromatic story) to catch this kind of thing automatically going forward? That's usually policy whenever a bug occurs

garrettmflynn:
Yeah I'll break out a test here. This is all data manipulation—so the easiest way would be running vitest every commit (like pytest). 

This will decouple it from the upstream workflow—but would test this particular merge. Might even be able to add a JSON Schema check in that test to ensure it doesn't break when passed to neuroconv.

Would you want the test infrastructure to be created in a follow-up PR?

CodyCBakerPhD:
> Would you want the test infrastructure to be created in a follow-up PR?

Depends on how critical of a fix this is; is it needed to test out the other active PRs?

garrettmflynn:
The feature is required for a complete end-to-end test of the conversion workflow. 

The test is not required. Let's enable `vitest` in a follow-up. I'll use this example as the test-case to show it's working.

I'm sure once we have this infrastructure in place, we'll have a lot more ideas about testing for the frontend.

CodyCBakerPhD:
OK sounds good

PR #149, opened 2023-05-01
Enable prettier on CSS and JS

CodyCBakerPhD:
Trying again from fresh branch, only including `.css` and `.js` files (or should be anyway)


.prettierignore
```diff
@@ -1,2 +1,3 @@
 curate.html
 tools/**/*
+src/renderer/src/stories/JSONSchemaForm.js
```
garrettmflynn, line 3: Any particular reason this was added to the ignore list? I ran things locally and cleaned this file up a bit (i.e. ignored dist and build folders)

CodyCBakerPhD: pre-commit bot errored out on it for some reason, can try to debug in follow-up

.prettierignore
```diff
@@ -1,2 +1,4 @@
-curate.html
 tools/**/*
+**/dist/**/*
+**/build/**/*
+src/renderer/src/stories/JSONSchemaForm.js
```
garrettmflynn, line 4: Looks good to me! Let's just make sure to try without this at some point.

garrettmflynn (approved):
Looks like this is great for a first pass.

.prettierignore
garrettmflynn: This does still modify .yml and .md files, so we'll have to decide if that's alright

garrettmflynn: > This does still modify .yml and .md files, so we'll have to decide if that's alright

Though this is with reference to a local run using `npm run format`

CodyCBakerPhD: I'd rather just use it for `.css` and `.js` suffixes for now; YAML, JSON and others each follow slightly different pre-commit rules

This is not intended to be run by `npm run format`, but rather any time you commit via `git`

PR #151, opened 2023-05-01
Unignore certain JS file

CodyCBakerPhD:
@garrettmflynn Let's see if it still complains

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,4 +1,3 @@
 tools/**/*
 **/dist/**/*
 **/build/**/*
-src/renderer/src/stories/JSONSchemaForm.js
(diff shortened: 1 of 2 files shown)


CodyCBakerPhD:
```
[error] src/renderer/src/stories/JSONSchemaForm.js: Error: Comment "Check self" was not printed. Please report this error!
```

CodyCBakerPhD:
Does that make any sense?

garrettmflynn:
> ```
> [error] src/renderer/src/stories/JSONSchemaForm.js: Error: Comment "Check self" was not printed. Please report this error!
> ```

There's an inline comment in that file with that text—though I have no idea why it would trigger an error.

Can you point me to where I can inspect this error myself? I couldn't see it in the pre-commit.ci details—though I may be looking in the wrong place

CodyCBakerPhD:
I imagine running `pre-commit run --all-files` should do what the bot is doing, but not 100% sure

garrettmflynn:
@CodyCBakerPhD This is such a weird error—but I was able to get rid of it

PR #152, opened 2023-05-02
Subject table updates

garrettmflynn:
This PR improves the subject table with the following updates:
1. Block from moving forward for a row without a subject_id + subjects without sessions
2. Apply a hardcoded species dropdown
3. Apply a little popup for information about each column
4. Convert variable names to capitalized headers
5. Group strain and genotype before weight


schemas/subject.schema.ts
```diff
@@ -0,0 +1,51 @@
...
+    "Mus musculus - House mouse",
+    "Homo sapiens - Human",
+    "Rattus norvegicus - Norway rat",
+    "Rattus rattus - Black rat",
+    "Macaca mulatta - Rhesus monkey",
+    "Callithrix jacchus - Common marmoset",
+    "Drosophila melanogaster - Fruit fly",
+    "Danio rerio - Zebra fish",
+    "Caenorhabditis elegans"
+  ]
```
CodyCBakerPhD, line 19: We're going to want a way that this matches the `dandi` version used by NeuroConv environment perfectly; can you use the map imported by the dandi schema that I pointed you to?

Would have to be done as a query to the API but I'd prefer that to hard-coding it since it is constantly being updated over time.

garrettmflynn: So you'd want me to query an endpoint on the GUIDE backend that hosts this information? It would there be another persistent endpoint to request from?

src/renderer/src/stories/Table.js
```diff
@@ -27,12 +27,68 @@ export class Table extends LitElement {
...
-    `;
+
+      .handsontable {
+        overflow: unset !important;
+      }
+`;
+
+        //       .ht_clone_top {
+        //         pointer-events: none;
+        //       }
```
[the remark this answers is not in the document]
garrettmflynn: This CSS was to create a custom tooltip. Are you happy with the native one? It's small and can be a bit finicky—but can overlay the page itself and, therefore, is better suited for arbitrary columns / positions of the table.

schemas/subject.schema.ts
garrettmflynn: Perfect. Yeah that's the usage we defined. Handontable also had a nice way to support this (an Autocomplete cell), so this is a feature at the table-level too (where Ryan had discussed it being only before or after that you could override the defaults).

CodyCBakerPhD:
Tests (including storybook) failing, but the icons look good on my side now

src/renderer/src/stories/Table.js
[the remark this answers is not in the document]
CodyCBakerPhD: So yeah, fine with removing the comments for a custom tooltip, the native one is good for now

CodyCBakerPhD (approved):

PR #161, opened 2023-05-04
Handle Ephys metadata as a set of editable tables

### LINKED ISSUE #131: Render electrode tables in Ecephys metadata
In order to effectively render ecephys metadata, we'll need to provide a way to render electrode tables. 

Previous discussions have settled on rendering this as a table where:
1. Users can manually add new columns
2. A CSV can be loaded to populate the table with known values
3. Users can Ctrl-C and Ctrl-V cells into the GUIDE from Google Sheets or Excel (which seem to copy raw .tsv information)

garrettmflynn:
fix #131 

This PR introduces a set of editable tables to handle Ecephys metadata. These tables link into the existing hooks for JSONSchemaForm to alert the app of its current state (e.g. valid, warning, error) and includes improved validation from the previous implementation.

There remain two important issues to keep in mind with the current implementation:
1. We cannot add table columns yet. It seems like Handsontable is blocking this action for named columns
2. Table performance is terrible with lots of rows, so I'm limiting the rendered rows to 20 for now

Nevertheless, this should allow us to pass electrode table data between Python and JavaScript, ensuring there are appropriate handlers in `neuroconv` to accept table edits.

When testing this branch, make sure to use the related dev branch (https://github.com/catalystneuro/neuroconv/pull/431) in `neuroconv`


CodyCBakerPhD:
Small error in pre-commit

```
[error] src/renderer/src/validation/index.js: Error: Comment "Can specify alternative client-side validation" was not printed.\
```

Also, for performance issues, just remember that as a fallback plan we can always swap to using an export copy of table + import of changed table as external CSV files and rely on the user to use a secondary program (such as Google Sheets) to perform the interactive modifications.

I know that seemed a little awkward when I first proposed it, but it would definitely work better than a table that doesn't scale to the number of elements we need.

Also, keep in mind the Subject table can also reach sizes of > 100 subjects and eventually will also need the capability of adding extra columns so it can also be affected by this too

garrettmflynn:
Makes sense. Thanks for reminding me of that approach! 

I'm going to render a simple HTML table that populates based on JSON Schema without all the interactivity. This should allow us to explore any low-level performance issues and support the "deferred interactivity" approach you note above.

CodyCBakerPhD:
OK so the current table for this PR only has 20 elements; can you expand that so I can play with the full table (using whatever implementation you're going with atm)

garrettmflynn:
I'd suggest running `npm run storybook` locally and changing the `subjects` variable to a different number in `Table.stories.js`. This will let you choose whatever number you like.

The SimpleTable implementation here would be ready to roll for a read-only system. Currently implementing additional renderers (e.g. sessions array) and editors (e.g. date-time picker) to round things out. 

At that point, the only things left are (1) mutating the original data object and (2) adding columns.

CodyCBakerPhD:
### Design ToDo's

- Do you need to display the 'name' column that combines the 'channel_name' and 'group_name'? Do you actually need it at all technically? Since it's not an actual property that would get passed back, and since we want to keep the other names as separate columns (which they currently are as well)

- For the temporary device validation matching; when I alter the device name of an electrode group, the cell glows red since it doesn't match a device, but no error message pops up; could be confusing to a user, is there a way to communicate that they need to select a defined device, or define a new device altogether? Or is this just such a temporary placeholder for the preferred dropdown alternative that we can defer this until later?

- The same logic for validating electrode group to device linkage applies to the electrode table 'group_name' column; that is, for each electrode, we only want the user to be able to select a defined electrode group from the previous table. Can you set that up?

CodyCBakerPhD:
> I'd suggest running npm run storybook locally and changing the subjects variable

I meant for electrodes, which I'm trying to use to visualize on SpikeGLX sizes

garrettmflynn:
> > I'd suggest running npm run storybook locally and changing the subjects variable
> 
> I meant for electrodes, which I'm trying to use to visualize on SpikeGLX sizes

Ah sure. If you change maxRows in `Table.js`, that will allow you to increase the number of rows represented.

CodyCBakerPhD:
Current approach takes ~20 seconds for me to load the entire electrode table for 1 band of SpikeGLX

I'm assuming it scales linearly, so multi-probe might take up to 1-2 minutes to render

I'm perfectly fine with that TBH, the only thing needed to make it feel smoother would be a spinning progress wheel (+/- descriptive text like 'loading probe information...' or similar) to indicate what is going on; otherwise that tab of the accordion just stays gray and the entire app is unresponsive during that time

garrettmflynn:
Did you try interacting with the table at all? I'm curious if you had similar issues with lag there.

CodyCBakerPhD:
> Did you try interacting with the table at all? I'm curious if you had similar issues with lag there.

I did, no lag during interaction. Just the initial load time

> The SimpleTable component is almost ready to roll and should fix these issues / allow for appropriate intervention. Maybe we should compare the approaches together early next week?

Sounds good

CodyCBakerPhD:
This is looking really great now! And performant!

Last thing about the name/description of the electrode column; I was imaging it to look more like this for organization

image

garrettmflynn:
@CodyCBakerPhD Aside from your last comment, this should be good to review. 

One final update I've added is **on-demand table data rendering** rather than the bulk loading scheme that requires an initial loader to freeze the page for over a minute. 

Let me know what you think of that and whether I should switch back to the original scheme.

garrettmflynn:
> This is looking really great now! And performant!
> 
> Last thing about the name/description of the electrode column; I was imaging it to look more like this for organization
> 
> image

Regarding this, would it make sense to sort these properly on the `neuroconv` side (i.e. the returned default array) or should we focus on sorting on the frontend / GUIDE?

garrettmflynn:
Other than this, we should be good to go. You're now able to export / import a .tsv file that allows for declaring custom columns if the encapsulating schema has `additionalProperties` set to True.

CodyCBakerPhD:
Some last changes before this can be merged

1)

Can you design the header seen here (on my side, latest changes to this PR)

image

to be more like this

image


2)

Can the Device and ElectrodeGroup still be interactive tables? They felt perfectly fine last time I checked this PR, and maybe even from the start of this PR

They will never reach the size of the electrode table; the TSV import/export is specifically for the large electrode table and not the others

CodyCBakerPhD:
> Regarding this, would it make sense to sort these properly on the neuroconv side (i.e. the returned default array) or should we focus on sorting on the frontend / GUIDE?

NeuroConv currently does not allow control of the order of custom column additions to the electrode table; predefined columns are added in a predefined order; the metadata might also not be the best way to go about controlling that, especially since eventually we plan to remove most arrays in the schema in lieu of key-value mappings

So I'd say that's a GUIDE specific thing, picking an order that is convenient for front-end display

src/renderer/assets/css/test.css
CodyCBakerPhD, line 1: Is this test.css with handsontable stuff still necessary?

src/renderer/src/stories/JSONSchemaForm.js
```diff
@@ -452,6 +474,47 @@ export class JSONSchemaForm extends LitElement {
...
+                                    this.#checkAllLoaded();
+                                },
+                            }));
+                        }
+                        // {
+                        //   return html`
+                        //   <div class="guided--array">
+                        //     <div class="guided--array-items">
+                        //       ${parent[name]?.map((item, i) => html`
+                        //         <div class="guided--array-item">
```
CodyCBakerPhD, line 498: This commented block still necessary?

src/renderer/src/stories/pages/guided-mode/data/GuidedMetadata.js
```diff
@@ -27,12 +28,14 @@ export class GuidedMetadataPage extends ManagedPage {
...
+        const schema = this.info.globalState.schema.metadata[subject][session]; // TODO: Order the Electrodes schema properties differently once present
+
         const form = new JSONSchemaForm({
             identifier: instanceId,
             mode: "accordion",
-            schema: this.info.globalState.schema.metadata[subject][session],
+            schema,
             results,
-            ignore: ["Ecephys", "source_script", "source_script_file_name"],
+            ignore: ["ElectricalSeriesAP", "source_script", "source_script_file_name"],
```
CodyCBakerPhD, line 38: I'd make this logic to ignore anything with the name 'ElectricalSeries' in it for now

Though the truth of it is more than we'd want to _include_ only the Device/ElectrodeGroup/Electrodes fields of `Ecephys` since technically speaking the ElectricalSeries field can have any key name under `Ecephys` (does not need to include the phrase 'ElectricalSeries, though all default ones do)

src/renderer/src/stories/pages/guided-mode/data/GuidedMetadata.js
```diff
@@ -54,7 +62,41 @@ export class GuidedMetadataPage extends ManagedPage {
...
+
     render() {
+        this.#resetLoadState(); // Reset on each render
+
+        // Swal.fire({
+        //     title: "Waiting for your metadata to render",
+        //     html: "Please wait...",
+        //     allowEscapeKey: false,
+        //     allowOutsideClick: false,
+        //     heightAuto: false,
```
CodyCBakerPhD, line 90: Comments needed?

garrettmflynn:
This should be fully functional. All the changes discussed on Wednesday (e.g. using two tables to represent the electrodes table, maintaining parity between the visualized and exported tsv files, etc.) have been finalized.

CodyCBakerPhD:
Other than that, just that one issue regarding creating a new ElectrodeGroup row (and see if it happens when adding a new Device row too) and this should be good to go!

garrettmflynn:
Looks like this between-table validation is working great now. I've a clear strategy for doing this in the future, so we shouldn't have an issue later on.

CodyCBakerPhD (approved):

CodyCBakerPhD:
One final problem that I just went ahead and raised a separate issue for so we can get this through

Thanks for sticking through all this; thankfully, this should resolve any general problem we have with metadata tables that follow this similar structure, so the time spent was fairly generic even though the current implementation is only for the Ecephys subsection

PR #189, opened 2023-05-15
Simplify pages and scale

### LINKED ISSUE #162: [Windows only?]: inner box not expanding with window size
image

But I think I saw, when Garrett runs it on his Mac, that is does scale with window size there

garrettmflynn:
fix #162 for height

### DIFF
--- a/src/renderer/assets/css/guided.css
+++ b/src/renderer/assets/css/guided.css
@@ -423,7 +423,7 @@
     display: flex;
     flex-direction: column;
     align-items: center;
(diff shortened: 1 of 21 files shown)


CodyCBakerPhD:
Ohhh yeah...

That feels so much better lol

There are occasional buttons now that maybe look a little awkward when stretched across the screen but I can raise issues for those later on as I see them

CodyCBakerPhD (approved):

PR #192, opened 2023-05-16
Basic setup for sending events related to conversion progress

### LINKED ISSUE #80: [MVP] Better progress bar for single session
This might get a little tricky as per previous conversations, but putting it on the eventual to-do list with low priority initially

It just wouldn't be a very nice experience if the user tries to run a conversion of a file that's > 10 GB and then the app just freezes for an indeterminate amount of time

Existing progress bars already exist under the hood via `tqdm`, but unclear how to pipe that out in real time asynchronously to the front-end

garrettmflynn:
fix #80 

This PR implements a basic server-side events (SSE) endpoint at `neuroconv/events` that can be used to monitor aspects of the neuroconv backend that are not returned directly from related HTTP requests.

Getting this PR beyond the draft stage will require some changes on `neuroconv` to allow hooking into `tqdm` to receive updates. After this is done, we can test a complete integration with the frontend.


garrettmflynn:
Added a separate demo for this feature that can be run using `python3 sse.py` at the root of the project. This will serve an HTML page using a Flask server, which connects using SSE and requests a series of asynchronous updates from `tqdm`. This will be printed as percent completed — along with an accompanying rate — on the page.

Let me know if there's other caveats to consider on the Python end. Otherwise this would be good to integrate into the ProgressBar component on the frontend and we can start planning how to pass messages like this from the GUIDE's dependencies.

CodyCBakerPhD:
Very cool!

Can you do a couple things for future usability of the test?

1) rename `sse` to `test_sse_display_of_tqdm` or something else more identifiable
2) move it into one of the testing folders, not project root
3) maybe add a brief README or descriptive header in the file itself mentioning how to run the demo (call `python` on it, expect instructions to be printed to console, navigate to the page in those instructions in a web browser, etc)

CodyCBakerPhD (approved):

PR #194, opened 2023-05-16
Filter out empty rows in an array

### LINKED ISSUE #172: [Idea] Clean up newlines from multi-line free text inputs
For certain text boxes, like keywords, the input is a list of strings

I noticed I can skip lines here and that results in empty strings being added to the list

image

image

I can't think of any cases where the newlines for fields like this would be kept (that is, for fields that are lists of strings); I can think of it in cases where the big text box is just a single string though, like the experiment description (or any other description field, technically)

garrettmflynn:
fix #172


garrettmflynn:
Pushed a test in vitest

CodyCBakerPhD (approved):

PR #197, opened 2023-05-19
Fix group check for Electrodes table

### LINKED ISSUE #195: [Improved validation] Check if electrode groups referenced by electrodes are defined in form
image

In this setting you can see I've manually assigned two random electrodes to different electrode groups, one of which doesn't even exist on the form

Not necessary for getting #161 through but would be needed eventually (eventually the field should be a dropdown anyway); though not a huge priority. NeuroConv creates the missing electrode group by default so the conversion itself is not hindered

garrettmflynn:
fix #195

### DIFF
--- a/src/renderer/src/stories/Table.js
+++ b/src/renderer/src/stories/Table.js
@@ -300,7 +300,7 @@ export class Table extends LitElement {
 
         table.addHook("afterValidate", (isValid, value, row, prop) => {
             const header = typeof prop === "number" ? colHeaders[prop] : prop;
-            let rowName = (this.keyColumn) ? rowHeaders[row] : row;
+            let rowName = this.keyColumn ? rowHeaders[row] : row;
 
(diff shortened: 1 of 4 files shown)


CodyCBakerPhD:
I'm unable to add new rows to the ElectrodeGroup table here, is that expected?

garrettmflynn:
I noticed some weird behavior too. If I went all the way back to the
Subject table and returned to the metadata page, it starts working again…I
can look into it

garrettmflynn:
This is an issue with `working-ecephys` generally—and not one that I've been able to figure out.

All I can say is it's related to Handsontable instances that are contained within a Web Component. All of the appropriate hooks are being set, so it seems like the context menu buttons are not effectively triggering any actions.

For now, I've swapped over to SimpleTable and caught a few bugs so it should be fully functional.

PR #198, opened 2023-05-19
Only delete subject if no sessions to place in the empty list

### LINKED ISSUE #196: [Bug] Manual modification of session_id from session manager table
If I go to this part of the workflow

image

then I interact with the table and change the name of the session like so

image

then my entire pipeline breaks like so on the next page (and this applies if I had retroactively gone back to the session manager after filling in later parts of the pipeline, in which case all my data entry was lost)

image

Definitely a highish priority to fix this

Right now, 2 avenues for fixing this

1) desired behavior would be for the values of the session ID field to match the downstream forms (if downstream forms changed) or otherwise if I go back and change them in the session manager that change should also propagate to the form

2) OK, (1) is a bit complicated. For now feel free to just make that cell non-editable (if that's possible or easier, which I assume it is) so that it can't be modified in a way that causes this issue

garrettmflynn:
fix #196


CodyCBakerPhD:
Is there an easy way to write a vite test for this?

CodyCBakerPhD:
Trying this out, the listing at least did not deplete however when I did the following basic set of actions

1) 'locate data programmatically = no'
2) set a source data for one SpikeGLX recording
3) proceed to form
4) go back to the session manager and change the name of that one session ID

--> the source file that I specified is now lost for that session

garrettmflynn:
As I've been interpreting it, the `sessions` property is an unordered list of the sessions that a particular subject has available. In other words, a session's metadata will be cleared if its entry is removed from the list. I thought this was the most elegant way of handling this input's behavior.

Do you want to generally maintain some understanding of the position of the session in the list and attach metadata to that rather than the actual session_id? It seems like as this scales up to more than one session removed and replaced, this scheme could be more confusing than useful (e.g. someone wants to remove one of the sessions in the middle and shifts all their metadata to different sessions).

I could handle this case *only* when one session is replaced by a session of a different name, where the new session now has the data from the previous session. Would that be an effective solution here?

CodyCBakerPhD:
I'll think more on the core problem and get back to it in an issue

Back to import question for this PR though

> Is there an easy way to write a vite test for this?

garrettmflynn:
Added. Not too complicated, just have to separate the data structure manipulation from the component.

CodyCBakerPhD (approved):

PR #200, opened 2023-05-22
Bring updates from Table to SimpleTable that support intertable validation

### LINKED ISSUE #199: [Bug]: Ecephys metadata validation of newly added ElectrodeGroup
Manually added `s1`

Validation flags anything other than `s0`

image

garrettmflynn:
fix #199

### DIFF
--- a/src/renderer/src/stories/SimpleTable.js
+++ b/src/renderer/src/stories/SimpleTable.js
@@ -395,7 +395,14 @@ export class SimpleTable extends LitElement {
                 label: "Remove Row",
                 onclick: (path) => {
                     const cell = this.#getCellFromPath(path);
-                    const { i } = cell.simpleTableInfo; // TODO: Support detecting when the user would like to remove more than one row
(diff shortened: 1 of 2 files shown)


CodyCBakerPhD:
That works! Thanks

Err.. At risk of sounding like a broken record... 😅

Is there any easy way to add a test for this?

garrettmflynn:
No worries, I assumed that was coming lol

This one will be more involved—but should be doable.

garrettmflynn:
It took quite a lot of doing—but it works now.

CodyCBakerPhD:
Hmm.. the tests are now failing however

garrettmflynn:
Ah the `fetch` API is only available on Node v18. I've set a conditional to ignore that call (which isn't used anyways) for the tests.

CodyCBakerPhD:
@garrettmflynn Was this test simplified as we discussed the other day?

garrettmflynn:
Yes. The test still instantiates a JSONSchemaForm rather than just two tables, but it only renders the two tables that are needed to render the simplified scheme (now declared entirely inside the test itself).

PR #203, opened 2023-05-30
Editable electrode columns

garrettmflynn:
This PR changes the Electrode Columns table from a BasicTable (read-only) to a SimpleTable (read/write)


CodyCBakerPhD:
Will save extensive testing for this after the generator demo is complepte

PR #205, opened 2023-06-05
Hide DANDI API key

garrettmflynn:
In preparation for the demo video, this PR hides the DANDI API key using a native password input (i.e. xyz is rendered as ***)

### DIFF
--- a/src/renderer/src/stories/pages/guided-mode/options/GuidedUpload.js
+++ b/src/renderer/src/stories/pages/guided-mode/options/GuidedUpload.js
@@ -36,6 +36,7 @@ export class GuidedUploadPage extends Page {
             properties: {
                 api_key: {
                     type: "string",
+                    format: 'password'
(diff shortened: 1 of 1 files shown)


CodyCBakerPhD:
Hmm... OK, not a bad idea

But in the video how do you think you'll show how to find/copy the key before pasting it in the GUIDE?

CodyCBakerPhD:
Also note that you can very easily regenerate that DANDI key with the little refresh symbol, which would be the approach I'd normally take when doing a recorded tutorial on this kind of thing

garrettmflynn:
> Hmm... OK, not a bad idea
> 
> But in the video how do you think you'll show how to find/copy the key before pasting it in the GUIDE?

Great point. I was thinking I'd keep it offscreen—but to show the full workflow, I'll switch to the approach you've indicated here.

PR #206, opened 2023-06-05
Generate informative global metadata page headers

garrettmflynn:
This PR provides informative headers on the NWBFile and Subject page, which previously had "Data Formats" as their header because of a careless copy-paste operation that we didn't notice.


CodyCBakerPhD:
Thanks, I did notice this a while back but forgot to make an issue for it, and also wasn't super bothered by it at the time

But for a round of polishing yeah we should clean this kind of stuff up where possible and easy

CodyCBakerPhD (approved):

PR #207, opened 2023-06-05
Show all neuroconv interfaces

garrettmflynn:
This PR opens up the interfaces provided to the GUIDE on the backend and instead limits the valid options on the frontend. 

This allows us to display the full list of interfaces we intend to support while ensuring users can only select the ones we've validated.


CodyCBakerPhD:
Looks good! That's a lot of work awaiting us lol

CodyCBakerPhD (approved):

PR #208, opened 2023-06-05
Fix state updates, rendering, and saving

garrettmflynn:
This PR fixes some issues with global state updates that were not rendering and saving properly.

The changes are as follows: 
1. No longer clear the global state when users go to the `guided/start`, which provides some information about the workflow
2. Check that the page history has a state before trying to write to it
3. Set the local state for the `guided/details` page so that `name` always renders and isn't cleared when going back to this page

### DIFF
--- a/src/renderer/src/progress.js
+++ b/src/renderer/src/progress.js
@@ -42,9 +42,9 @@ export const save = (page, overrides = {}) => {
 
     params.set("project", guidedProgressFileName);
 
+    // Update browser history state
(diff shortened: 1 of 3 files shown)


CodyCBakerPhD:
@garrettmflynn Not sure how to fix the conflict here

garrettmflynn:
Fixed

CodyCBakerPhD (approved):

PR #209, opened 2023-06-06
Simplify final conversion workflow

garrettmflynn:
This PR proposes a simplified workflow for users of the GUIDE in preparation for our demo video.

Instead of requiring users to specify the output folder, we can instead create a `Conversions` subdirectory in the main `NWB GUIDE` directory that already holds the progress and stub files. With this change, we can always refer to the output of a particular pipeline simply by its project name.

This change allowed for the removal of the Conversion Options page, which only had the output_folder field


CodyCBakerPhD:
Some conflicts here; it's a pretty common use case for people to want to control where the data gets written to though (usually to a separate disk)

What we could do as a compromise is use this strategy as the _default_ save location so that people don't _need_ to specify it if they don't care

But we really should keep the capability of choosing where the files get written if the user wants it to be different from the proposed default here

garrettmflynn:
Sounds good. Centralizing the project output files presents a significant amount of benefits for accessing files down the line—though, now that you're bringing this up, I think it'd be possible to add a symlink at the default location and write the real files to an arbitrary location if the user specified it.

I will try this approach today. Would it be alright to ask for an output folder for the project on the first page (Project Setup) rather than adding a new page just for this info?

CodyCBakerPhD:
>  Would it be alright to ask for an output folder for the project on the first page (Project Setup) rather than adding a new page just for this info?

That could work, yeah

The symlink records could be nice as you say, but just be aware they aren't guaranteed to be persistent (what happens when they delete the files in the true destination folder outside of the app?) - rather it seems the intention behind that approach is to maintain a log of what was written where and when, which we should just use actual logging for

garrettmflynn:
To some extent, yeah. Though I'd say a primary benefit of the actual filesystem representation is hosting files simply as in #213.

garrettmflynn:
This feature is implemented. The user-defined output folder can be referenced using a symlink in the default location

CodyCBakerPhD (approved):

Issue #185, opened 2023-05-15
[Idea] Better file content previews


CodyCBakerPhD:
The current behavior is fine for MVP (essentially a dump of what you get from calling `print` on an NWB file in-memory)

But @garrettmflynn mentioned it might not be that hard to run NWB Widgets or WebNWB visualizations on a stubbed NWB file directly through the app

I know many users would greatly appreciate this, since they often think of their data not in the way that it's stored in the file but how it is rendered by visualization software (especially for microscopy data)

Would also give a good early indicator of temporal alignment via PSTH's (for sorting data anyway)

garrettmflynn:
Just started experimenting with rendering NWB Widgets on the browser on the `visualize-with-widgets` branch. Based on discussions with Ben and others, I tried using (1) PyScript / Pyodide as well as (2) `voila`—but came up short using both methods. 

## Pyscript
`pyscript` is very particular to inconsistent dependencies and requires you to specify a `*py3-none-any.whl` file for dependencies you'd like to pre-resolve. This allowed me to fix an issue with `numpy` brought on by `ruamel.yaml`—though I wasn't able to resolve a similar problem for `multidict` from `nwbwidgets`. 

Based on this, I was not able to run NWB Widgets or stream a file using `ffspec`. However, I was able to fix Oliver's original demo and print out the string representation of a remote file.

## Voila
`voila` is very close to working—though rendering hangs for properties other than `file_create_date` and `subject`. No errors are thrown.

The original notebook can render all other properties when accessed via VS Code, so the code appears to be functional.

CodyCBakerPhD:
@garrettmflynn  At a minimum, can we leverage the new HTML repr capabilities of NWBFiles instead of the pure-text printout?

garrettmflynn:
Just added the HTML representation capacity in #233

CodyCBakerPhD:
I saw, testing it out now (also appended to name of the base PR)

CodyCBakerPhD:
Closed by #232

PR #213, opened 2023-06-06
Host conversion files with Flask

garrettmflynn:
This PR allows us to access stub and final .nwb files using the Flask server. This will be useful for #185.

It was tested locally using WebNWB to ensure that these files can, in fact, be downloaded or streamed using the URL returned from the conversion.

> **Note:** The true base of this PR is `[simplify-final-conversion-workflow](https://github.com/NeurodataWithoutBorders/nwb-guide/tree/simplify-final-conversion-workflow)`. However, I accidentally pushed these changes to that branch and had to revert them manually—and they continue to register as too similar to compare against.

### DIFF
--- a/pyflask/manageNeuroconv/manage_neuroconv.py
+++ b/pyflask/manageNeuroconv/manage_neuroconv.py
@@ -229,6 +229,8 @@ def convert_to_nwb(info: dict) -> str:
     """
 
     nwbfile_path = Path(info["nwbfile_path"])
+    output_folder = info.get("output_folder")
+    project_name = info.get("project_name")
 
     run_stub_test = info.get("stub_test")
 
@@ -283,6 +285,11 @@ def convert_to_nwb(info: dict) -> str:
         overwrite=info.get("overwrite", False),
         conversion_options=options,
     )
+
+    if output_folder:
+        os.symlink(
+            Path(output_folder) / project_name, conversion_save_path / project_name
+        )  # Have a scoped pointer to the converted project
 
     return dict(preview=str(file), file=str(resolved_output_path))
 
--- a/schemas/json/path-expansion.schema.json
+++ b/schemas/json/path-expansion.schema.json
@@ -5,7 +5,7 @@
             "format": "directory",
             "description": "Enter the base directory of your data."
         },
-        "file_path": {
+        "format_string_path": {
             "type": "string",
             "description": "Enter a format string to locate data.",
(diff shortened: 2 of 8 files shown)


pyflask/manageNeuroconv/info/__init__.py
CodyCBakerPhD, line 1: Not recommended practice to have actual code in an `__init__.py`

Can this path configuration be done any other way?

CodyCBakerPhD:
> Note: The true base of this PR is [simplify-final-conversion-workflow](https://github.com/NeurodataWithoutBorders/nwb-guide/tree/simplify-final-conversion-workflow). However, I accidentally pushed these changes to that branch and had to revert them manually—and they continue to register as too similar to compare against.

OK, should we just proceed with this as the complete base then? and close the other one? Or is there a better way

pyflask/manageNeuroconv/__init__.py
```diff
@@ -8,3 +8,5 @@
     upload_to_dandi,
     listen_to_neuroconv_events,
 )
+
+from .info import *
```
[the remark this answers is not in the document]
garrettmflynn: I've replaced this with an explicit export

garrettmflynn:
The changes seem to have hit now—so I've handled them in a merge commit for the new base.

pyflask/manageNeuroconv/info/urls.py
```diff
@@ -0,0 +1,13 @@
...
+import json
+
+project_base_path = Path(__file__).parent.parent.parent.parent
+path_config = Path(
+    project_base_path, "paths.config.json"
+)  # NOTE: You're going to have to ensure that this copies over to the Python distribution
+f = path_config.open()
+data = json.load(f)
+stub_save_path = Path(Path.home(), *data["stubs"])
+conversion_save_path = Path(Path.home(), *data["conversions"])
```
CodyCBakerPhD, line 11: OK I see; another common convention is for variables like this that are truly global in the sense of the Python side of things, is to all-capitalize them

I'd also recommend adding file or folder specification to it's really clear to someone looking at it what it means; I'm assuming they are directories?

```suggestion
STUB_SAVE_FOLDER_PATH = Path(Path.home(), *data["stubs"])
CONVERSION_SAVE_FOLDER_PATH = Path(Path.home(), *data["conversions"])
```

pyflask/manageNeuroconv/info/__init__.py
CodyCBakerPhD: Yeah it's better - thanks!

pyflask/manageNeuroconv/info/urls.py
garrettmflynn: This should be good to go now.

CodyCBakerPhD (approved):

PR #219, opened 2023-06-07
Update app copy to accurately reflect behavior and content

garrettmflynn:
This PR updates some of the app's text and behavior to ensure that this accurately reflects its behavior.

1. The sidebar options for Overview and Guided Mode have been replaced by Home and Conversions respectively.
    - Relatedly, all `guided/x` paths are now `conversion/x` paths
    - "Guided Mode" is still used to refer to the conversion process, but not the landing page that we use to open it. Since we don't have any other mode to contrast it with, for now, it seems a bit weird to refer to it too often.
2. The Start a New Dataset button on the start page skips the first Conversions page (with the list) and acts as if the user pressed the Convert a New Dataset button on the skipped page
3. The Conversions In Progress tab was replaced with Existing Conversions based on Ben's update to the demo script

### DIFF
--- a/src/renderer/src/pages.js
+++ b/src/renderer/src/pages.js
@@ -84,11 +84,11 @@ style="margin-right: 30px; margin-bottom: -5px"
 
 const pages = {
     "/": new GettingStartedPage({
-        label: "Overview",
+        label: "Home",
         icon: overviewIcon,
     }),
-    guided: new GuidedHomePage({
-        label: "Guided Mode",
+    conversion: new GuidedHomePage({
+        label: "Conversions",
         icon: guidedIcon,
         pages: {
             start: new GuidedStartPage({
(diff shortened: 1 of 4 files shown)


CodyCBakerPhD (approved):

CodyCBakerPhD:
The storybook for the home page is blank https://www.chromatic.com/test?appId=640a4d93e34604f275368983&id=648109812fbe92a5c2cf6ac6

Is this intended? Or should it appear as it used to when it was called 'Overview'?

garrettmflynn:
Good catch. I'd replaced `guided/x` with `conversion/x`, but not just `guided` with it.

This will be something that should be handled when we merge this with #218 as well, since it has one of those transitions.

CodyCBakerPhD (approved):

PR #232, opened 2023-07-12
Prepare for using all interfaces and add HTML repr

garrettmflynn:
This PR bumps the neuroconv target for environments and checks for the existence of the Ecephys helper functions added in https://github.com/catalystneuro/neuroconv/pull/447 to show Ecephys metadata at all.


CodyCBakerPhD (approved):

pyflask/manageNeuroconv/manage_neuroconv.py
```diff
@@ -103,6 +104,12 @@ def get_first_recording_interface(converter):
             return interface
 
 
+def is_supported_recording_interface(recording_interface, metadata):
```
CodyCBakerPhD, line 107:
```suggestion
def is_supported_recording_interface(recording_interface, metadata):
    """
    Temporary conditioned access to functionality still in development on NeuroConv.
    
    Used to determine display of ecephys metadata depending on the environment.
    
    Alpha build release should therefore always return False for this.
    """
```

Just adding a note to the docstring so we remember

CodyCBakerPhD (dismissed):

CodyCBakerPhD:
Looks great! This is a nice compromise between the string printout and the eventual NeuroSift integration

CodyCBakerPhD (approved):

PR #233, opened 2023-07-12
Use HTML representation

garrettmflynn:
This PR uses the HTML representation provided by `hdmf` on the stub test preview


CodyCBakerPhD:
Recommend pinning the version of HDMF requirements to be >= 3.7 to ensure the HTML feature will be included

PR #234, opened 2023-07-12
Add Neuroscope recording interface

garrettmflynn:
This PR adds the NeuroScopeRecordingInterface to the list of supported interfaces. A file is outputted from a pipeline using this interface.

@CodyCBakerPhD How would you like to go about validating that this actually works?


CodyCBakerPhD:
When the PR for each new interface is added, you try it once before opening the PR (the entire pipeline) then I'll try it before merge

I really can't think of any other better way to get an entire end-to-end integration test using ground truth test files (could use a series of Chromatic stories with hard coded payloads but that isn't really the same thing... would also greatly inflate our snapshot usage)

How does that sound?

garrettmflynn:
Sounds good to me. Would there ever be cases where the pipeline is successful but the output file is incorrect?

CodyCBakerPhD:
> Would there ever be cases where the pipeline is successful but the output file is incorrect?

Depends on what you mean be 'incorrect'

If that means 'user specified a bunch of metadata fields in the GUIDE but they did not end up in the file' then the issue could be in either the GUIDE passing things to NeuroConv or a problem in NeuroConv itself

If that means 'data fields were written incorrectly' then I really hope not because that would be a big NeuroConv problem. I like to think everything is OK there

If that means 'NeuroConv default values are actually incorrect and we can't modify them in GUIDE right now' yeah, that is technically possible (w.r.t electrode metadata and the like) but the current solution is just to not use GUIDE in that case since they would _need_ metadata customization. Anyone that needs that will be directed to use NeuroConv directly until GUIDE supports all the metadata customization

CodyCBakerPhD:
@garrettmflynn on this PR I get an error when generating preview

```JavaScript
Failed to load resource: net::ERR_CONNECTION_REFUSED
index.ts:72 Connected to Python back-end successfully
127.0.0.1:4242/neuroconv/convert:1          Failed to load resource: the server responded with a status of 500 (INTERNAL SERVER ERROR)
utils.js?t=1689345824441:35 Uncaught (in promise) Error: Request to convert failed: 'ElectrodeColumns'
    at run (utils.js?t=1689345824441:35:33)
    at async GuidedMetadataPage.runConversions (Page.js?t=1689345824441:122:28)
    at async GuidedFooter.onNext [as __onNext] (GuidedMetadata.js?t=1689345824441:27:29)
```

using `main` of NeuroConv - I thought we had merged in a fix for that?

garrettmflynn:
You are correct, though this branch was not up-to-date with `main`. 

Just worked on my end. Can you try again?

CodyCBakerPhD:
Had to do a complete wipe and reinstall of GUIDE on my system, but yes it works on this branch now

The last thing I'd request specific to this PR is adding a source data field for the `gain` argument: https://github.com/catalystneuro/neuroconv/blob/main/src/neuroconv/datainterfaces/ecephys/neuroscope/neuroscopedatainterface.py#L119

It's technically optional so schematically doesn't _need_ to be filled in - but it's encouraged

This is actually a good example of your earlier question

> Would there ever be cases where the pipeline is successful but the output file is incorrect?

Right now, I'd say yes because the default `gain` from `neo` has always been incorrect as far as we can tell, which is why we always override it when using it in practice

garrettmflynn:
The changes in the `intan` branch expose non-required properties and hide those we don't like seeing right now.

Rebasing to that branch exposes `gain` properly.

CodyCBakerPhD:
Very minor - how difficult is it to get rid of the up and down arrows to the right of the form entry box for the `float` valued gain? It increments in +/- 1 which isn't really applicable for this field

image

garrettmflynn:
Would you want it to increment at a different interval? Or would you just
prefer it to be gone.

Both are fairly straightforward.

CodyCBakerPhD:
> Would you want it to increment at a different interval? Or would you just prefer it to be gone.

I'm not against the general idea of incrementing a float (in terms of the generic web component) whenever the resolution exists in a reasonably stable range, but it's this specific float that should probably not be incrementable. So for this case, 'gone' is preferred

Usually the gain value is a very particular number and can potentially go into several (up to 6+ or more) decimal places out, so typing it explicitly really is the way to go here to make sure it's accurate

garrettmflynn:
Added the relevant interval increment features to #270.

CodyCBakerPhD:
Pipeline looks good!

CodyCBakerPhD (approved):

PR #235, opened 2023-07-12
Add NeuroScopeSortingInterface and BiocamRecordingInterface + bug fixes

garrettmflynn:
This PR adds the NeuroScopeSortingInterface to the supported interfaces list.

Because of the structure of the test data, we had to enable additional options (non-required properties) on the source data specification page.


pyflask/manageNeuroconv/manage_neuroconv.py
```diff
@@ -310,7 +311,7 @@ def update_conversion_progress(**kwargs):
         # Update with the latest metadata for the electrodes
         ecephys_metadata["Electrodes"] = electrode_column_results
 
-    del ecephys_metadata["ElectrodeColumns"]
+    ecephys_metadata.pop("ElectrodeColumns", "No key found")  # Might not exist
```
CodyCBakerPhD, line 314: What role is the second argument intended to play here?

`pop` returns either the value at that key in the dict, or it returns a default (the second argument) - in this case the string, but the returned value is not kept to be used for anything else and IDK that it would even get dumped to stdout - did you intend for this to raise an error, or be silent?

CodyCBakerPhD:
Comments from https://github.com/NeurodataWithoutBorders/nwb-guide/pull/236 now apply to this branch, which cannot be accidentally merged this time 😅

garrettmflynn:
Now that this is up-to-date with `main` and additional keys have been added to the ignore list, this should be displaying the appropriate metadata form correctly.

CodyCBakerPhD:
Both of these interfaces look great!

CodyCBakerPhD (approved):

PR #237, opened 2023-07-12
Add Blackrock Interfaces

garrettmflynn:
This PR adds the Blackrock recording and sorting interfaces to the GUIDE's list of supported interfaces.


CodyCBakerPhD (approved):

CodyCBakerPhD:
I had a made a couple of minor changes

Added `nsx_override` to ignored source arguments; it only has special usage in one conversion and I'm not entirely sure why or if it still needs to exist except for back-compatability (which we could definetely soft deprecate)

Also added a temporary fix for an issue with the sorting interface not auto-populating the Ecephys key unlike the other two existing sorting interfaces

Go ahead and merge if you agree with these changes

CodyCBakerPhD (approved):

CodyCBakerPhD (approved):

PR #240, opened 2023-07-12
Add Intan Interface

garrettmflynn:
Add IntanRecordingInterface to the list of supported interfaces


pyflask/manageNeuroconv/manage_neuroconv.py
```diff
@@ -310,7 +312,7 @@ def update_conversion_progress(**kwargs):
         # Update with the latest metadata for the electrodes
         ecephys_metadata["Electrodes"] = electrode_column_results
 
-    del ecephys_metadata["ElectrodeColumns"]
+    ecephys_metadata.pop("ElectrodeColumns", "No key found")  # Might not exist
```
CodyCBakerPhD, line 315: What role is the second argument intended to play here?

`pop` returns either the value at that key in the dict, or it returns a default - in this case the string, but the returned value is not kept to be used for anything else

CodyCBakerPhD:
Starting the review with Intan since that's a good clean format (usually) to try out

First thing I notice is this

image

We had  previously agreed that we should not expose `es_key` as an option in the GUIDE since we ought to be able to internally handle all those references (it plays no role in reading source data nor is it a conversion option)

We also should not expose `stream` info without either needing to, as well as providing a selectable dropdown of available streams (class method `get_stream_names` should or at least could exist for any interface that needs it, but might only have built it into a couple so far).

And actually, as I look at this interface, I'd rather not expose `stream_id` in this case at all, could you suppress it for the Intan interface just for now?

CodyCBakerPhD:
Conversion pipeline ran without a hitch but when I tried to  upload to DANDI staging `200560` I got

```JavaScript
Failed to load resource: the server responded with a status of 500 (INTERNAL SERVER ERROR)
utils.js:35 Uncaught (in promise) Error: Request to upload failed: failed validation
    at run (utils.js:35:33)
    at async GuidedFooter.onNext [as __onNext] (GuidedUpload.js:25:29)
```

CodyCBakerPhD:
Running `dandi validate` independently gives

```
[DANDI.NO_DANDISET_FOUND] C:\Users\Raven\Downloads\GIN-Intan — Path is not inside a Dandiset
[pywnb.GENERIC] C:\Users\Raven\Downloads\GIN-Intan\sub-001\sub-001_ses-1.nwb — Index (0) out of range for empty dimension
```

CodyCBakerPhD:
`nwbinspector` confirms the second is an actual `pynwb.validation` issue, something quite wrong in the file. Investigating

CodyCBakerPhD:
OK the good news is it's not a problem with this data format itself - if I don't specify a bunch of random metadata throughout the forms (only the base requirements) then it works fine

So I'll try to track down the issue in the metadata and raise a separate issue for it. All this specific PR needs is 

(i) modularization to just the Intan interface
(ii) do not expose `es_key` at all
(iii) Just for Intan, do not currently expose `stream_id`

garrettmflynn:
(ii) and (iii) have been completed. 

When you say "modularization to just the Intan interface", are there any specific places that needs to be accomplished? The only changes introduced—aside from the inclusion of the Intan interface—are those just introduced.

CodyCBakerPhD:
> When you say "modularization to just the Intan interface", 

I might be hallucinating, or it might have been due to the changing target - I thought at some point this branch included changes for Biocam (maybe something else too?)

Anyway it all looks good now, just ran through the pipeline and source data page looks good

CodyCBakerPhD (approved):

PR #243, opened 2023-07-12
Add OpenEphys Interface

garrettmflynn:
This PR adds support for the OpenEphysRecordingInterface to the GUIDE.


CodyCBakerPhD:
Huh - I'm not seeing it show  up in the interface listing, any ideas why that would be?

CodyCBakerPhD:
Works well - one caveat; while we do technically expose `stream_name` here, so technically if the user knows the exact name of their stream, they can type it in and get it to work; the schema should however be adjusted on the NeuroConv side to be an enum of all the available streams

I assume once the NeuroConv source schema changes to an enum of strings, this element will transform to a dropdown?

CodyCBakerPhD (approved):

garrettmflynn:
> Works well - one caveat; while we do technically expose `stream_name` here, so technically if the user knows the exact name of their stream, they can type it in and get it to work; the schema should however be adjusted on the NeuroConv side to be an enum of all the available streams
> 
> 
> 
> I assume once the NeuroConv source schema changes to an enum of strings, this element will transform to a dropdown?

You got it. Should be a smooth transition to the desired render if the schema is updated properly.

PR #244, opened 2023-07-12
Add Spike2 Interfaces

garrettmflynn:
This PR adds support for the Spike2RecordingInterface and CEDRecordingInterface.

Based on initial tests, it seems like these interfaces need to be tested on separate pipelines.

### DIFF
--- a/guideGlobalMetadata.json
+++ b/guideGlobalMetadata.json
@@ -3,6 +3,8 @@
         "SpikeGLXRecordingInterface",
         "PhySortingInterface",
         "NeuroScopeSortingInterface",
-        "BiocamRecordingInterface"
+        "BiocamRecordingInterface",
+        "Spike2RecordingInterface",
+        "CEDRecordingInterface"
     ]
 }


CodyCBakerPhD:
@garrettmflynn Can you either resolve the conflict in the `JSONSchemaForm.js` or re-open from a fresh checkout off of `main`? Sorry for taking so long to try this out (also see quick comment above)

guideGlobalMetadata.json
```diff
@@ -3,6 +3,8 @@
         "SpikeGLXRecordingInterface",
         "PhySortingInterface",
         "NeuroScopeSortingInterface",
-        "BiocamRecordingInterface"
+        "BiocamRecordingInterface",
+        "Spike2RecordingInterface",
+        "CEDRecordingInterface"
```
CodyCBakerPhD, line 8:
```suggestion
        "Spike2RecordingInterface",
```

CEDRecordingInterface = Spike2RecordingInteface

However, CEDRecordingInterface is the deprecated version of the name: https://github.com/catalystneuro/neuroconv/blob/main/src/neuroconv/datainterfaces/ecephys/spike2/spike2datainterface.py#L76-L79

CodyCBakerPhD:
Well...

I knew we'd run into this eventually but now it's sooner rather than later.

This is the first major example of an interface whose external dependencies are very, shall we say, particular in terms of what kind of environments are supported

I forget if M1 works at all in the given ranges, I can test that later.

But the particulars are seen here: https://github.com/catalystneuro/neuroconv/blob/main/src/neuroconv/datainterfaces/ecephys/spike2/spike2datainterface.py#L9-L14

Basically, to get this interface to work we'd need to downgrade our current environment specifications to `3.9` and that ought to work cross-platform

CodyCBakerPhD:
Found a couple unrelated bugs while testing this, raised issues

The one thing I will leave as note for something for us to consider is, this interface does not support `.smr` file suffixes; only `.smrx`

Not sure the best way to communicate or handle that on the UX level but we can leave it for a broader discussion across interfaces and suffixes maybe

Rest looks good

CodyCBakerPhD (approved):

PR #245, opened 2023-07-12
Add SpikeGadgets Interface

garrettmflynn:
This PR adds support for the SpikeGadgetsRecordingInterface to the GUIDE.


CodyCBakerPhD:
OK this mostly looks good, but if I manually input `gains` like so

image

then when I reach the end of the metadata form and try to generate preview I get

```javascript
utils.js:35 Uncaught (in promise) Error: Request to convert failed: can't multiply sequence by non-int of type 'float'
    at run (utils.js:35:33)
    at async GuidedMetadataPage.runConversions (Page.js:129:28)
    at async InstanceManager.onClick (GuidedMetadata.js:165:50)
```

garrettmflynn:
This should be handled more gracefully after merging #249, though I can implement a quick fix to recognize lists of numbers and convert them properly.

garrettmflynn:
I've done as much as I can on my end.


Though this is actually a problem with the schema for the SpikeGadgetsRecordingInterface, which specifies that `gains` should be an array but doesn't indicate the type of items contained within it:
Screenshot 2023-08-08 at 2 32 17 PM

CodyCBakerPhD:
Ahhh, yes. I see the problem.

Would you prefer to use this solution or await the correct on the NeuroConv side?

garrettmflynn:
So the latest changes actually depend on those `neuroconv` corrections to work. Once those are applied, though, this should be fine.

The only stand-in solution would be to use `try / catch` to determine if a certain input could be parsed as a number, then use that numeric value instead if it can.

CodyCBakerPhD:
OK, might as well table it alongside the TDT fix as well then

garrettmflynn:
Updated this since @stephprince wanted to use it. We have an issue in NeuroConv where Gains isn't specified explicitly, otherwise this should be good to go:
Screenshot 2024-04-18 at 2 58 56 PM

CodyCBakerPhD:
Right, that's the trouble with this interface

It needs a a list of float gains to be specified

But the length of the list we cannot know until we know the file path and try to read it, so validation and typing is tricky

CodyCBakerPhD:
Awaiting ongoing work between the Frank lab and neo to resolve the stream issues

PR #246, opened 2023-07-12
Add TDT Support

garrettmflynn:
This PR adds support for the TdtRecordingInterface to the GUIDE.

### DIFF
--- a/guideGlobalMetadata.json
+++ b/guideGlobalMetadata.json
@@ -3,6 +3,7 @@
         "SpikeGLXRecordingInterface",
         "PhySortingInterface",
         "NeuroScopeSortingInterface",
-        "BiocamRecordingInterface"
+        "BiocamRecordingInterface",
+        "TdtRecordingInterface"
     ]
 }


CodyCBakerPhD:
Going to convert this to draft until NeuroConv directly exposes `gain` (should be renamed to `conversion_factor` in NeuroScope too probably): https://github.com/catalystneuro/neuroconv/issues/540

Would need to dig a bit deeper but it appears this format might not automatically record the conversion factor, meaning the data being converted here is potentially wrong, which is misleading.

If using NeuroConv directly the user would have been expected to set the gain manually after initializing the interface so not quite a bit of an issue there; but here the user has no control exposed

CodyCBakerPhD:
This was recently fixed on the NeuroConv side (gain was explicitly exposed)

However, a bizarre issue on the front end side (not altogether blocking, will probably raise as separate issue)

If I type `0` - `.` - any other number, such as a common gain of `0.195`, the form entry bugs out and the `.` moves the insertion line to the left and does not add the `.` character

To insert the correct value, I need to _start_ with the `.` instead

CodyCBakerPhD (approved):

PR #248, opened 2023-07-13
Delete Ecephys property only if it exists

garrettmflynn:
This PR ensures that the deletion of the Ecephys property will not error if it doesn't exist (e.g. on non-Ecephys interfaces)


pyflask/manageNeuroconv/manage_neuroconv.py
```diff
@@ -162,7 +162,7 @@ def get_metadata_schema(source_data: Dict[str, dict], interfaces: dict) -> Dict[
 
     # Delete Ecephys metadata if ElectrodeTable helper function is not available
```
CodyCBakerPhD, line 163:
```suggestion
    # Delete Ecephys metadata if ElectrodeTable helper function is not available (temporary solution)
```

Just adding a note how this is temporary to begin with

CodyCBakerPhD (approved):

PR #251, opened 2023-07-13
Updates for Behavior Interfaces

garrettmflynn:
This PR fixes Ecephys access during the conversion call and avoids Behavior validation on the JSON Schema Form.


garrettmflynn:
@CodyCBakerPhD Similar to the Ecephys metadata, should we also be blocking the Behavior metadata (as well as for the other data formats)?

CodyCBakerPhD (approved):

CodyCBakerPhD:
> @CodyCBakerPhD Similar to the Ecephys metadata, should we also be blocking the Behavior metadata (as well as for the other data formats)?

@garrettmflynn For the time being, yeah

PR #256, opened 2023-07-17
Mac deployment

rly:
Changes to make the mac build work on M1 and hopefully also in CI (no arm64)


package.json
```diff
@@ -22,6 +22,7 @@
     "deploy-mac": "electron-builder build --mac --publish always",
     "deploy-linux": "electron-builder build --linux --publish always",
     "python-onefile-build": "python -m PyInstaller --onefile --clean ./pyflask/app.py --distpath ./pyflaskdist --add-data ./paths.config.json;. --collect-data jsonschema_specifications --collect-all pynwb --collect-all hdmf --collect-all ci_info",
+    "python-onefile-build-mac": "python -m PyInstaller --onefile --clean ./pyflask/app.py --distpath ./pyflaskdist --add-data './paths.config.json:.' --collect-data jsonschema_specifications --collect-all pynwb --collect-all hdmf --collect-all ci_info",
```
CodyCBakerPhD, line 25: At this point the only difference between these two commands is the `'` character on the mac version - do you need that? Or are you able to use the common `python-onefile-build`?

rly: The other difference is the delimiter -- `;` for windows and `:` for mac and linux. https://pyinstaller.org/en/stable/usage.html#cmdoption-add-data Annoying. I'll try using a single delimiter and see if it actually matters.

CodyCBakerPhD:
Latest attempt looked like it failed because it already exists in the release, maybe try again with yet another different package version here?

garrettmflynn:
I was able to fix things enough to remove the popup error on startup. So the PyInstaller output files are now being targeted properly.

However, there are still issues with the generated executables themselves. When run on their own, I get the following errors:

## Mac
```
/Users/garrettflynn/Documents/Github/nwb-guide/out/python/app ; exit;
  File "app.py", line 59, in <module>
IndexError: list index out of range
[43422] Failed to execute script 'app' due to unhandled exception!

Saving session...completed.
```

## Windows
```
  File "app.py", line 14, in <module>
  File "PyInstaller\loader\pyimod03_importers.py", line 476, in exec_module
  File "setupUtils\__init__.py", line 3, in <module>
  File "PyInstaller\loader\pyimod03_importers.py", line 476, in exec_module
  File "setupUtils\configureRouteHandlers.py", line 1, in <module>
  File "PyInstaller\loader\pyimod03_importers.py", line 476, in exec_module
  File "apis\__init__.py", line 2, in <module>
  File "PyInstaller\loader\pyimod03_importers.py", line 476, in exec_module
  File "apis\apiNeuroConv.py", line 5, in <module>
  File "PyInstaller\loader\pyimod03_importers.py", line 476, in exec_module
  File "manageNeuroconv\__init__.py", line 1, in <module>
  File "PyInstaller\loader\pyimod03_importers.py", line 476, in exec_module
  File "manageNeuroconv\manage_neuroconv.py", line 2, in <module>
  File "PyInstaller\loader\pyimod03_importers.py", line 476, in exec_module
  File "neuroconv\__init__.py", line 1, in <module>
  File "PyInstaller\loader\pyimod03_importers.py", line 476, in exec_module
  File "neuroconv\basedatainterface.py", line 7, in <module>
  File "PyInstaller\loader\pyimod03_importers.py", line 476, in exec_module
  File "pynwb\__init__.py", line 15, in <module>
  File "PyInstaller\loader\pyimod03_importers.py", line 476, in exec_module
  File "hdmf\common\__init__.py", line 210, in <module>
  File "PyInstaller\loader\pyimod03_importers.py", line 476, in exec_module
  File "hdmf\common\sparse.py", line 1, in <module>
  File "PyInstaller\loader\pyimod03_importers.py", line 476, in exec_module
  File "scipy\sparse\__init__.py", line 287, in <module>
  File "PyInstaller\loader\pyimod03_importers.py", line 476, in exec_module
  File "scipy\sparse\csgraph\__init__.py", line 185, in <module>
  File "PyInstaller\loader\pyimod03_importers.py", line 476, in exec_module
  File "scipy\sparse\csgraph\_laplacian.py", line 7, in <module>
  File "PyInstaller\loader\pyimod03_importers.py", line 476, in exec_module
  File "scipy\sparse\linalg\__init__.py", line 120, in <module>
  File "PyInstaller\loader\pyimod03_importers.py", line 476, in exec_module
  File "scipy\sparse\linalg\_isolve\__init__.py", line 4, in <module>
  File "PyInstaller\loader\pyimod03_importers.py", line 476, in exec_module
  File "scipy\sparse\linalg\_isolve\iterative.py", line 9, in <module>
ImportError: DLL load failed while importing _iterative: The specified module could not be found.
[15488] Failed to execute script 'app' due to unhandled exception!
```

garrettmflynn:
The Mac executable is now working—though, for some reason, it may take several minutes to actually start the server. 

SODA originally waited 5min before canceling on the user, which didn't seem necessary until now.

This result was achieved by manually running the executable. I'm currently testing various execution methods in Node.js to see if I can replicate this behavior.

garrettmflynn:
I just tried `--onedir` instead of `--onefile` mode and now the executable will start in a few seconds. Seems like we just have too many dependencies to unpack—so things take forever to start.

Do we want to try switching to `--onedir` generally?

garrettmflynn:
Using `npm run build:mac` now works on my non-M1 Mac. Can you both confirm?

CodyCBakerPhD:
> Do we want to try switching to --onedir generally?

Can you summarize the difference? Trying to read on it doesn't quickly describe why this would be different since there's still just a single executable in both cases

CodyCBakerPhD:
Hmmm... from https://github.com/pyinstaller/pyinstaller/issues/5453#issuecomment-756081999 it seems that `onefile` = `onedir + gzip` and thus to run the `onefile` we must unzip it first each time

CodyCBakerPhD:
Success on M1! 🎉

Have yet to put it through its full paces (need to download testing data, will do that tomorrow - reminder, anyone can download it easily from [here](https://drive.google.com/drive/folders/11Op9GRJSwuBrvyGamQtSVvkYj9BUcHf3?usp=sharing))

package.json
```diff
@@ -11,19 +11,22 @@
     "dev:server": "cd pyflask && python -m flask run --port 4242",
     "build": "npm run build:app",
     "build:app": "electron-vite build",
+    "build:mac": "npm run build && npm run build:python-unix && npm run build-mac",
+    "build:win": "npm run build && npm run build:python && npm run build-win",
+    "build:python": "python -m PyInstaller --onedir --clean --noconfirm ./pyflask/app.py --distpath ./out/python --add-data ./paths.config.json;. --collect-all pynwb --collect-all hdmf --collect-all ci_info",
+    "build:python-unix": "python -m PyInstaller --onedir --clean --noconfirm ./pyflask/app.py --distpath ./out/python --add-data ./paths.config.json:. --collect-all pynwb --collect-all hdmf --collect-all ci_info",
+    "build-win": "electron-builder build --win --publish never",
+    "build-mac": "electron-builder build --mac --publish never",
+    "build-linux": "electron-builder build --linux --publish never",
```
CodyCBakerPhD, line 20: Very excited this is actually working now - one remaining comment I have for the new `npm` shorthands - can we rename this block to something more precise? The difference between `build:<platform>` and `build-<platform>` is quite minor and could be an avoidable source of confusion going into the future

garrettmflynn:
Woohoo! Yeah that’s the downside of onefile.

Good to know we have a simpler way to get the testing data. Thanks for
letting me know!

package.json
```diff
@@ -11,19 +11,22 @@
     "dev:server": "cd pyflask && python -m flask run --port 4242",
     "build": "npm run build:app",
     "build:app": "electron-vite build",
+    "build:mac": "npm run build && npm run build:python-unix && npm run build-mac",
+    "build:win": "npm run build && npm run build:python && npm run build-win",
+    "build:python": "python -m PyInstaller --onedir --clean --noconfirm ./pyflask/app.py --distpath ./out/python --add-data ./paths.config.json;. --collect-all pynwb --collect-all hdmf --collect-all ci_info",
+    "build:python-unix": "python -m PyInstaller --onedir --clean --noconfirm ./pyflask/app.py --distpath ./out/python --add-data ./paths.config.json:. --collect-all pynwb --collect-all hdmf --collect-all ci_info",
```
CodyCBakerPhD, line 17: Same qualm with this name, perhaps `build-flask-dist`? 

Also was there a significant reason to rename the output directory to `./out` instead of `./pyflaskdist` or even something better (if we can think of one)?

garrettmflynn:
For sure about those script key updates.

Do you have any ideas? Right now, the difference is that one runs just the
electron-builder command while the other runs every script required to
build the app from scratch.

If not, I’ll think about it some more tomorrow.

garrettmflynn:
out is what’s used by the vite + electron tool and I both knew it
transferred to the build and was already ignored on git—so I just
consolidated all the build assets, including the Flask server, there.

Would you want to rename that? I originally saw the consolidation as a
benefit, but we don’t have to hold to that either.

package.json
CodyCBakerPhD, line 17:
```suggestion
    "build:mac": "electron-vite build && npm run build:flask && npm run build-mac",
    "build:win": "electron-vite build && npm run build:flask && npm run build-win",
    "build:flask": "python -m PyInstaller --onedir --clean --noconfirm ./pyflask/app.py --distpath ./out/python --add-data ./paths.config.json:. --collect-all pynwb --collect-all hdmf --collect-all ci_info",
```

Perhaps I'm having a "where's Waldo" moment, but the two `build:python` look the same to me

Also suggested rename to be more explicit about this subcommand - while Python is certainly a part of what is built, the full Flask server distribution is much more than just the base Python

Also removing excess subcommands that can be included in the main `build:<platform>` commands to reduce abstraction (let me know if the original structure is actually somehow needed)

CodyCBakerPhD:
> Would you want to rename that? I originally saw the consolidation as a
benefit, but we don’t have to hold to that either.

Good to know the `out` convention is typical among other frameworks - I suppose I'd be open to adopting it based on that fact

However, not really sure what you mean by consolidation here. When I run the build process I end up with 3 new folders in the Git directory: `build`, `dist`, and `out`. The one that works (launches app and communicates interface listing from backend) for me is the `.dmg` in the `dist`

package.json
CodyCBakerPhD, line 20:
```suggestion
    "build:mac": "npm run build && npm run build:python-unix && npm run build-mac",
    "build:win": "npm run build && npm run build:python && npm run build-win",
    "build:python": "python -m PyInstaller --onedir --clean --noconfirm ./pyflask/app.py --distpath ./out/python --add-data ./paths.config.json;. --collect-all pynwb --collect-all hdmf --collect-all ci_info",
    "build:python-unix": "python -m PyInstaller --onedir --clean --noconfirm ./pyflask/app.py --distpath ./out/python --add-data ./paths.config.json:. --collect-all pynwb --collect-all hdmf --collect-all ci_info",
    "build:electron:win": "electron-builder build --win --publish never",
    "build:electron:mac": "electron-builder build --mac --publish never",
    "build:electron:linux": "electron-builder build --linux --publish never",
```

suggestion for rename on the final build step (electron; merging the electron-vite plus the flaskdist together)

package.json
```diff
@@ -11,19 +11,22 @@
...
-    "build-win": "electron-builder build --win --publish never",
-    "build-mac": "electron-builder build --mac --publish never",
-    "build-linux": "electron-builder build --linux --publish never",
-    "deploy-win": "electron-builder build --win --publish always",
-    "deploy-mac": "electron-builder build --mac --publish always",
-    "deploy-linux": "electron-builder build --linux --publish always",
-    "python-onefile-build": "python -m PyInstaller --onefile --clean ./pyflask/app.py --distpath ./pyflaskdist --add-data ./paths.config.json;. --collect-data jsonschema_specifications --collect-all pynwb --collect-all hdmf --collect-all ci_info",
+    "deploy:win": "electron-builder build --win --publish always",
+    "deploy:mac": "electron-builder build --mac --publish always",
+    "deploy:linux": "electron-builder build --linux --publish always",
```
CodyCBakerPhD, line 29:
```suggestion
    "deploy:electron:win": "electron-builder build --win --publish always",
    "deploy:electron:mac": "electron-builder build --mac --publish always",
    "deploy:electron:linux": "electron-builder build --linux --publish always",
```

would then be this for consistency, if accepted

tools/anaconda-env/environment-MAC-arm64.yml
```diff
@@ -0,0 +1,32 @@
...
+      # - pennsieve == 6.2.0
+      - configparser == 6.0.0
+      # - python-docx == 0.8.11
+      # - xlrd == 2.0.1
+      # - biopython == 1.81
+      # - gevent == 22.10.2
+      - flask == 2.3.2
+      - flask_restx == 1.1.0
+      - flask-cors == 4.0.0
+      # - protobuf == 3.20.0
```
CodyCBakerPhD, line 30:
```suggestion
  - numcodecs = 0.11.0
  - lxml = 4.9.3  # pypi build fails due to x64/arm64 mismatch so install from conda-forge
  - pyedflib = 0.1.32  # pypi build fails due to x64/arm64 mismatch so install from conda-forge
  - jsonschema = 4.18.0  # this installs jsonschema-specifications which is needed for validation
  - pydantic[email] = 2.0.2  # email validator is used by dandi
  - numpy  # may have x64/arm64 mismatch issues so install from conda-forge
  - pip
  - pip:
      - chardet == 5.1.0
      - configparser == 6.0.0
      - flask == 2.3.2
      - flask_restx == 1.1.0
      - flask-cors == 4.0.0
```

cleanup - M1 build worked fine without these

pyflask/app.py
```diff
@@ -56,8 +58,13 @@ def get(self):
 
 
 if __name__ == "__main__":
-    port = sys.argv[1]
-    api.logger.info(f"Starting server on port {port}")
-    app.run(host="127.0.0.1", port=port)
+    multiprocessing.freeze_support()  # https://stackoverflow.com/questions/32672596/pyinstaller-loads-script-multiple-times#comment103216434_32677108
```
CodyCBakerPhD, line 61: Note to future self: if ever any issues parallelizing conversions over sessions, think about this line

CodyCBakerPhD:
I managed to get the automatic CI deployment for non-M1 macs working ass well

https://github.com/NeurodataWithoutBorders/nwb-guide/releases/tag/untagged-c1e437922cb44502d5af

@garrettmflynn Can you give that a shot and see if it works?

Also @rly on your work computer when you get a chance?

bendichter:
image

package.json
garrettmflynn: Windows and Unix-like systems have different separators for the `build:python` command so there do need to be two versions.

testExecutable.js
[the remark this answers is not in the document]
garrettmflynn: Yes, I'd actually tried to remove this from the repo altogether—but now I think it will be useful to test the Pyinstaller executables on their own before attempting to create + launch the Electron build.

garrettmflynn:
> Good to know the `out` convention is typical among other frameworks - I suppose I'd be open to adopting it based on that fact
> 
> However, not really sure what you mean by consolidation here. When I run the build process I end up with 3 new folders in the Git directory: `build`, `dist`, and `out`. The one that works (launches app and communicates interface listing from backend) for me is the `.dmg` in the `dist`

So the `electron-vite` system creates the `out` folder for the `out/renderer`, `out/main`, and `out/preload` subfolders that compose the Electron JavaScript assets. 

`pyinstaller` will produce the `build` folder in addition to the desired `--distpath`, where `build`["contains some log files and working files"](https://www.askpython.com/python/pyinstaller-executable-files#:~:text=A%20build%20folder%20is%20created,the%20given%20python%20script%20name.) that we don't actually use while `--distpath` (set to `out/python`) is the final executable file and related assets.

Finally, `electron-builder` outputs the `dist` folder which contains the actual build NWB GUIDE app.

As such, you can think of these as flowing in the following way:
```
Source files —> out —> dist
```

`out` and `dist` therefore have a logical reason for their separation, while `build` really doesn't. And, in fact, I think `out` would better be characterized as `build`. Would we want to try renaming it to that using the `--outDir` option?

CodyCBakerPhD:
> Would we want to try renaming it to that using the --outDir option?

Sounds good to me for consistency

CodyCBakerPhD:
> As such, you can think of these as flowing in the following way:

Thanks for explaining! Would there be a problem though with doing `out/dist`?

garrettmflynn:
Technically no, though `build` (what `out` would become) is usually not ready for production while `dist` is. Usually they aren't nested

bendichter:
@garrettmflynn let me know when this is ready for me to test on my M1!

CodyCBakerPhD:
> @garrettmflynn let me know when this is ready for me to test on my M1. I'm excited to give it a try!

@bendichter The M1 won't be ready until @rly builds, manually signs it (he's the only one with the Mac certification). and manually deploys

garrettmflynn:
@CodyCBakerPhD Also fixing an issue where the DANDI configuration .yaml file isn't found for the nwbinspector, which renders the GUIDE only semi-functional starting on the Subjects page.

garrettmflynn:
Now should at least run a full conversion.

CodyCBakerPhD (approved):

CodyCBakerPhD:
Confirmed with Garrett that the local build works fine; the CI deployment is a follow-up issue that intermittently works but improvements to the Flask app structure should make it more robust

CodyCBakerPhD:
@garrettmflynn Whelp, my bad for not checking this on my end before merging

Since this was merged to `main` I can no longer run `npm start`

Error

```
error during start dev server and electron app:
Error: No electron app entry file found: D:\GitHub\nwb-guide\build\main\main.js
    at ensureElectronEntryFile (D:\GitHub\nwb-guide\node_modules\electron-vite\dist\chunks\lib-87772d93.js:26:23)
    at Object.startElectron (D:\GitHub\nwb-guide\node_modules\electron-vite\dist\chunks\lib-87772d93.js:118:5)
    at createServer (D:\GitHub\nwb-guide\node_modules\electron-vite\dist\chunks\lib-994f1722.js:74:21)
    at async CAC.<anonymous> (D:\GitHub\nwb-guide\node_modules\electron-vite\dist\cli.js:48:9)
```

Confirmed that checking out before this merge is perfectly fine however

What would you suggest? Is there an easy fix or should I just try to revert this

garrettmflynn:
Easy fix. There just has to be a file where the `main` field in the `package.json` points.

PR #261, opened 2023-08-03
Open a popup when the user jumps to a new page using the sidebar

### LINKED ISSUE #201: [Idea] Link session ID (and possibly other fields) across pages
From the conversation of https://github.com/NeurodataWithoutBorders/nwb-guide/pull/198#issuecomment-1555331298

> As I've been interpreting it, the sessions property is an unordered list of the sessions that a particular subject has available.

Correct; this understanding and the mapping from the main session manager (subject table) is fine

> In other words, a session's metadata will be cleared if its entry is removed from the list. I thought this was the most elegant way of handling this input's behavior.

As a principle of design, I believe we should avoid situations in which a user is allowed to enter information (in this case, potentially quite a lot of information) and then without warning cause a complete loss of that information

Even if we wanted to keep the behavior of wiping form data upon session ID change from the session manager page, it should at least give some big warnings and ask for confirmation, letting the user know they risk losing inputs

I'd prefer to avoid the situation altogether though, and have an idea introduced below for how I prefer to think about it

> Do you want to generally maintain some understanding of the position of the session in the list and attach metadata to that rather than the actual session_id?

Position does not matter, it's always an unordered list

What I envision is rather than that the value of the session ID as specified on the session manager page is not only the default of the NWBFile metadata form, but that they are the same property under-the-hood, so that changes in one place occur in the other

How complicated would it be to implement something like that? The same idea might apply to all information on that table now that I think about it

But this would help resolve those 'merge resolution' problems you had hypothesized might happen by keeping them in-sync with each other

> It seems like as this scales up to more than one session removed and replaced, this scheme could be more confusing than useful (e.g. someone wants to remove one of the sessions in the middle and shifts all their metadata to different sessions).

This part sounds like we're thinking about different things

Yes, if I remove a session from the sidebar or the session manager table (back completely backspacing out of that item in the list; again, should probably ask for confirmation somehow since that could occur accidentally), I acknowledge the loss of that information

All I'm talking about though is giving the session ID a different value as an in-place modification

> I could handle this case only when one session is replaced by a single session of a different name, where the new session now has the data from the previous session. Or some other constrained condition. Would that be an effective solution here?

Not sure about the implementation level details; that is, if you need to copy the session info in order to alter that one field

If I were doing this with Python data classes or similar, I would simply change that one value (session ID) in the structure instead of copying a new structure with that one changed value

garrettmflynn:
fix #201 (Subpoint 4)

This PR interrupts the user when they try to jump to a new page using the sidebar.

It highlights the option to keep editing, otherwise a secondary button can be pressed that will move forward with the page transition without saving.


CodyCBakerPhD:
OK, so I get the popup, which looks nice and scary to avoid accidental data loss, but a couple things I noticed (that aren't really huge deals ATM so feel free to ignore them and merge this if you'd rather improve in a separate lower-priority PR)

**a)** It seems like every page does this - including pages such as the stub preview, where there is no 'editing' being performed and thus no data to be lost

I was envisioning this notification strictly for a couple of the heaviest typing pages, such as session manager (including fstring location page if applicable), source data specification, and file/subject metadata, but see point (b) if it's even necessary now?

I could see this constant notification as being potentially annoying for users that like to hop around different stages of their pipeline willy-nilly (not sure if that's yet a use case to be concerned about, so I defer to your opinion)

**b)** I entered text into a bunch of random fields on one pipeline and without saving (and ignoring the 'keep editing' highlight), was able to navigate to a different page, as well as different pipeline, as well as exit and relaunch the entire app - all with the same information still being saved (data was not lost)

What are the precise conditions that someone would actually lose their data by switching pages?

**c)** The one way I was able to lose the data was by going back to the session manager and manually deleting the session I had entered random text into

That operation should _probably_ feel more difficult as to not be done accidentally, which is more in the spirit of what I was going for with this feature request...

...but it's also pretty self evident operation, so not a huge priority ATM

CodyCBakerPhD (dismissed):

garrettmflynn:
This PR should be ready for review now. 

The save workflow is now explicit, where pages mark themselves as having unsaved changes—resulting in a displayed popup when the user attempts to exit that page. There is now a `beforeSave` callback that fires any time the page saves, allowing linear and jump transitions to be relatively consistent without code duplication.

I've tried testing all the pages for desired behaviors. Let me know if anything has snuck through. In particular, we might still have to move some of `onNext` callbacks into the `beforeSave` callback (i.e. jump transitions might result in an unanticipated lack of change to downstream forms)

CodyCBakerPhD:
When going through a fresh tutorial pipeline, I reach the 'Populate Subject Details' button on the Path Expansion page but cannot proceed further due to

```javascript
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'results')
    at GuidedFooter.onNext [as __onNext] (GuidedPathExpansion.js?t=1691431954856:94:49)
```

garrettmflynn:
Fixed! Wasn't targeting the right object for the globalState

CodyCBakerPhD:
Hm... I'm not getting the popup now, but I also am unable to navigate away from the metadata page until I hit the 'Save' button

Likewise when on the subject table page, if I modify something I'm unable to navigate away from it unless I hit the 'Generate Data Structure' button

Error in console:

```javascript
Uncaught (in promise) ReferenceError: Swal is not defined
    at Dashboard.subSidebar.onClick (Dashboard.js:92:25)
    at NavigationSidebar.<anonymous> (NavigationSidebar.js:121:35)
    at EventPart.handleEvent (lit-html.ts:2003:29)
Dashboard.subSidebar.onClick @ Dashboard.js:92
(anonymous) @ NavigationSidebar.js:121
handleEvent @ lit-html.ts:2003
```

CodyCBakerPhD:
New error starting from a fresh tutorial pipeline

Upon reaching the subject metadata table, it fails to render and I get

```javascript
Uncaught (in promise) TypeError: Cannot convert undefined or null to object
    at GuidedSubjectsPage.render (GuidedSubjects.js:74:31)
    at GuidedSubjectsPage.update (lit-element.ts:160:24)
    at GuidedSubjectsPage.performUpdate (reactive-element.ts:1331:14)
    at GuidedSubjectsPage.scheduleUpdate (reactive-element.ts:1263:17)
    at GuidedSubjectsPage.__enqueueUpdate (reactive-element.ts:1235:25)
```

CodyCBakerPhD:
Oh man, I can't hit 'Exit' either to go back to the main page

```javascript
GuidedSubjects.js:23 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'validate')
    at GuidedSubjectsPage.beforeSave (GuidedSubjects.js:23:24)
    at GuidedSubjectsPage.save (Page.js:96:39)
    at GuidedSubjectsPage.beforeTransition (GuidedSubjects.js:32:14)
    at Main.to (Page.js:64:14)
    at GuidedFooter.to (GuidedFooter.js:51:54)
    at GuidedFooter.onExit [as __onExit] (GuidedFooter.js:33:29)
    at EventPart.handleEvent (lit-html.ts:2003:29)
```

garrettmflynn:
Is this on an existing pipeline? A new pipeline just worked for me.

CodyCBakerPhD:
> Is this on an existing pipeline?

Yep, freshly generated tutorial pipeline

Maybe I should try `npm ci` and try again?

garrettmflynn:
Let me try using the same pipeline. Could be because of the autofill for the tutorial pipeline.

garrettmflynn:
Got it. There was an issue with the metadata generation from the PathExpansion page.

CodyCBakerPhD:
I'm so sorry 😆

Starting from a fresh tutorial pipeline, I now get

```javascript
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'source_data')
    at createForm (GuidedSourceData.js:98:53)
    at utils.js:35:70
    at Array.map (<anonymous>)
    at utils.js:35:45
    at Array.map (<anonymous>)
    at mapSessions (utils.js:34:10)
    at GuidedSourceDataPage.mapSessions (Page.js:118:63)
    at GuidedSourceDataPage.render (GuidedSourceData.js:128:27)
    at GuidedSourceDataPage.update (lit-element.ts:160:24)
    at GuidedSourceDataPage.performUpdate (reactive-element.ts:1331:14)
```

on the source data specification page

garrettmflynn:
No worries. I was anticipating that we'd have a lot of back and forth here.

Was intending to send this yesterday—but it didn't go through. This should be good to review again.

CodyCBakerPhD:
Not getting any pop-up

```javascript
ncaught (in promise) ReferenceError: Swal is not defined
    at Dashboard.subSidebar.onClick (Dashboard.js:102:25)
    at NavigationSidebar.<anonymous> (NavigationSidebar.js:121:35)
    at EventPart.handleEvent (lit-html.ts:2003:29)
```

garrettmflynn:
My bad, looks like a bad merge. This responsibility has been deferred to the page itself.

garrettmflynn:
Good catch, another bad merge issue. With the separation of working and registered changes, I'm resolving the current value differently.

CodyCBakerPhD:
OK looks and feels good all the way around!

Last thing is just to adjust the back-end tests that are failing: https://github.com/NeurodataWithoutBorders/nwb-guide/actions/runs/5930539516/job/16080472450?pr=261

garrettmflynn:
@CodyCBakerPhD Should be good to go. Wasn't getting the `jsonschema_specifications` error on Windows locally—but looks like this is a flag that we need to add across platforms.

CodyCBakerPhD (approved):

PR #262, opened 2023-08-03
Ignore subject_id and session_id in File Metadata page

### LINKED ISSUE #201: [Idea] Link session ID (and possibly other fields) across pages

garrettmflynn:
fix #201 (Subpoint 1)

This PR ignores the subject_id and session_id on the File Metadata page so they can only be edited correctly, from the Subject metadata table.


CodyCBakerPhD:
Agreed, limiting the number of branching cases for setting the same field is a good solution. Now the high-level session manager is the only place such information is controlled

CodyCBakerPhD (approved):

Issue #252, opened 2023-07-14
[Bug] Session ID not being used in file name for 'Run Conversion' step


CodyCBakerPhD:
I built a pipeline with `session_id="testing_123"`, but the output file name (before DANDI organize + upload) was `sub-001_ses-1.nwb` which is the 'session ID' from the subject table, which was not updated by the downstream forms

The filename should instead show up as `sub-001_ses-testing-123.nwb` (NOTE: that underscore was converted to a dash in the name since underscores within a block get converted to dashes following the DANDI convention)

garrettmflynn:
You updated the `session_id` field in the File Metadata section, correct? At present, we've assumed that these updates are only handled on the Subjects table.

The quickest way to handle this specific issue would be to ignore the `session_id` and `subject_id` options in the File Metadata page, as changing these would require a complete re-render of that page (or at least many elements) anyways.

This solution is what we decided on in #201, where different aspects of the same problem are discussed.

In addition to the discussion on #201, a more general problem here is that we have two representations of subject metadata:
1. The Subjects table
2. One Subject accordion per subject + session combination

It is easy enough to apply (1) as the default value for each session in (2). However, it is not straightforward—nor is it conceptually accurate—to merge (2) into (1). 

As such, we need to reconcile changes on both. How would we like to do this?

At first thought, it could be most straightforward to only allow entry of immutable fields (e.g. `subject_id`, `date_of_birth`, etc.), thereby limiting the editable metadata in the File Metadata section to things like `weight` and `age`. But this is just what's on the top of my mind.

If this is a line of inquiry we'd like to pursue, we should probably move this discussion to a separate issue before going too far.

garrettmflynn:
I've published several related PRs that limit the extent of this issue.

#262 removes the ability of the user to edit these IDs on the File Metadata form.

#264 implements a proper data cascade so that the GUIDE generally only allows updates to flow to dependent forms (e.g. Subject table to NWB File Subject metadata) if the dependent value hasn't been modified by the user.

With these updates, the user is more clearly locked into their decisions made earlier in the application—though they can apply overrides to properties that don't structure the UI itself (e.g. not `session_id` and `subject_id`). 

We may wish to further lock decisions, as noted in the previous comment. But this particular bug isn't really an intended behavior to begin with—and now it can't be done.

garrettmflynn:
Closed based on the data flow updates in #262 and #264

PR #264, opened 2023-08-04
Improved Data Representation

garrettmflynn:
This PR improves the data representation for the NWB GUIDE File Metadata by implementing a proper cascade of Subject Global Metadata —> Subject table metadata —> NWB File Subject Metadata for the final results sent to `neuroconv`. 

With these changes, each project form result is kept separate until aggregated to send to the server. As such, we can now distinguish between auto-filled form values (i.e. values from higher in the cascade) and user-defined values.

> **Note:** This allows users to define, for instance, a global `genotype` property, overwrite this for a specific session, then change that global property without negating the previous change. 

This also allows us to dismiss problems related to reconciliation noted in #252, as data _should not_ flow from further down in the cascade. As such, we should make UI changes that limit our ability to edit a common bit of data (e.g. `session_id`) managed by an upstream form (e.g. the Subject table) in dependent forms (e.g. the NWB File metadata form).

### DIFF
--- a/src/renderer/src/pages.js
+++ b/src/renderer/src/pages.js
@@ -115,7 +115,7 @@ const pages = {
             }),
 
             subjects: new GuidedSubjectsPage({
-                title: "Subjects",
+                title: "Subject Metadata",
                 label: "Subject details",
                 section: sections[0],
             }),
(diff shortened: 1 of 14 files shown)


CodyCBakerPhD:
Sounds great - will try it out tomorrow

Can you fix the one little merge conflict? I wasn't sure how you wanted the `parent` specification stuff done with that FileSystemSelector

CodyCBakerPhD:
Yeah this feels like a much clearer workflow now, thanks!

CodyCBakerPhD (approved):

PR #273, opened 2023-08-07
[Mac] Basic Infrastructure to Open NWB Files with the GUIDE

garrettmflynn:
This PR allows you to receive filesystem commands to open an NWB file with the GUIDE. Since we don't have `neurosift` / other frontend visualizer integrated yet, this is simply a proof-of-concept to pass the correct absolute path to the browser.

Additional changes were made to nicely handle the rendering of the GUIDE's main browser window for all possible start states (e.g. inactive, minimized, etc.) and transition triggers (e.g. the `open-file` event, clicking the docked app on Mac, etc.)


CodyCBakerPhD:
@garrettmflynn Can you resolve these conflicts?

garrettmflynn:
Done. Just tested and should be good.

garrettmflynn:
Hmm that's weird, as it looks like a variable access issue. Since the variable is only used once, I've replaced it with the generator function for it.

Can you try again?

CodyCBakerPhD:
Yep, `npm start` works now

How should I go about testing the file association? I'm trying a local build+install from local dist to see if that does it on Windows

garrettmflynn:
Yep that's the way I've been doing it. The option to open with NWB GUIDE should be available after installation of a local build.

CodyCBakerPhD:
OK confirmed Mac works to associate linkage (though attempting to run it fails to launch the app with some vague OS error; will push more on that when stand-alone neurosift operation is the final destiation)

Windows did not work to associate an automatic association, but this is fine to start. Not a high priority yet just a proof of concept

CodyCBakerPhD (approved):

PR #276, opened 2023-08-08
Make chromatic CI manual dispatch only

CodyCBakerPhD:
To partially combat overusage of Chromatic during a billing period when there are tons of PRs that trigger an update every time they need their branch updated, I propose to make it required for merge but only triggerable manually, which the reviewer will do during final review (or perhaps multiple times if revisions are requested)


garrettmflynn (approved):
Sounds good to me

PR #279, opened 2023-08-09
Generate source data schema and update stories to use them properly.

### LINKED ISSUE #274: [Testing] Add story pages for the source data screen for each interface
@garrettmflynn To more easily catch events like the one detected in #247 I think it would be useful to have chromatic stories for the source data screen for each interface in order to track and detect unintended changes to global components

This might increase our snapshot usage, but TBH chromatic is proving invaluable for this kind of stuff so I'd say it's worth it if we have to pay for a few months during active development

garrettmflynn:
fix #274 

This PR allows us to visualize the GuidedSourceData page for each interface in order to confirm if they are rendered properly.


CodyCBakerPhD:
First off, these look great! Very happy to have visual coverage for all of these

I left comments on a couple of items on the recent Chromatic build, mostly minor

I don't suppose there's a way to provide a dummy file path for one or two of the interfaces in Chromatic just to visually check what a valid source data page _would_ look like?

This would mostly be for having a visual check for the behavior caught in #318 and #314 on the PathExpander page, but could apply to other example pages too

garrettmflynn:
Would you want those dummy file paths on the Source Data page or the Locate Data page? The parent update issue wasn't actually affecting the former.

CodyCBakerPhD:
> Would you want those dummy file paths on the Source Data page or the Locate Data page? 

At least one of both ideally

garrettmflynn:
Now that I think of it, this issue will only arise if the user updates the input—manually or programmatically. I've already implemented a test in #318 that will detect the behavior you're seeking to test.

If we provide the results to the form beforehand, they'll simply render how they're supposed to. I created a Story that contains all the interfaces in Source Data and provides a dummy path to the 2nd one. 

Would you still like me to fill in the Locate Data blank given that it'll just show how the form renders rather than confirm whether the validation is accurate?

CodyCBakerPhD (approved):

CodyCBakerPhD:
These new stories look great! Thanks a bunch

PR #285, opened 2023-08-12
Try to fix windows scipy Pyinstaller issue

CodyCBakerPhD:
@garrettmflynn Does this branch fix your specific scipy issue? Solution from: https://saturncloud.io/blog/troubleshooting-scipy-import-error-with-pyinstaller/#:~:text=This%20error%20typically%20occurs%20because,for%20PyInstaller%20to%20work%20properly.

You were right that I don't run into the same problem in my environment. Quite strange...


CodyCBakerPhD:
Solution 1 from the suggestion did not work for the deployment version - the file size of the release was also unchanged by the extra flags

Will continue trying other solutions later

garrettmflynn:
When run locally, the configuration options specified here appear to error immediately—so the Pyinstaller command will not even run.

Can you tell if that's happening for the deployment version?

PR #290, opened 2023-08-15
Custom install on flask server side

CodyCBakerPhD:
@garrettmflynn This is how I was imagining a flask endpoint for doing a custom install, let me know if it makes sense


CodyCBakerPhD:
replaced by #300

PR #291, opened 2023-08-16
[Simplify PyFlask - REFERENCE - DO NOT MERGE]

CodyCBakerPhD:
Vast simplification of workflow for pyflask, as well as utilizing localized import per API call


pyflask/app.py
```diff
@@ -1,47 +1,72 @@
...
 
-api = configureAPI()
 
-configureRouteHandlers(api)
+@api.route("/echo")
+class Echo(Resource):
+    @api.expect(echo_parser)
+    def get(self):
+        args = echo_parser.parse_args()
+        return args["arg"]
```
CodyCBakerPhD, line 53: @garrettmflynn Do we actually use this anywhere?

pyflask/app.py
```diff
@@ -1,47 +1,72 @@
...
 
-api.init_app(app)
 
+@api.route("/conversions/<path:path>")
+class SendConversions(Resource):
+    def get(path: str):
+        from .tools.info import CONVERSION_SAVE_FOLDER_PATH
```
CodyCBakerPhD, line 59: I also changed these to use relative imports, if this causes a problem feel free to switch back to `from tools.info` (which refers to the folder `tools` that is closest to this file) instead of `.tools.info` (which is explicit about where that folder should be)

CodyCBakerPhD:
Probably just going to use this first attempt as a guiding reference and implement in smaller pieces

pyflask/app.py
CodyCBakerPhD: It does cause a problem...

Apparently Flask essentially runs the entire import stack as if it were a typical Python script instead of a package, so absolute importing to local namespaces will have to used

However, the resulting cleanup reduces the occurrences of that quite a bit so hopefully it would be too confusing

PR #292, opened 2023-08-16
~~Add CI for testing pyflask distributable~~

CodyCBakerPhD:
Separate CI that can run on push to a PR


.github/workflows/test-pyflask-build.yaml
```diff
@@ -0,0 +1,94 @@
...
+      - name: Setup Node
+        uses: actions/setup-node@v3
+        with:
+          node-version: "16"
+
+      - name: Build PyFlask distribution
+        run: npm run build:flask:${{ matrix.shorthand }}
+
+      - name: Run test on distributed executable
+        run: node tests/testPyinstallerExecutable.js
```
CodyCBakerPhD, line 87: Though I see that by default, this test targets the `build` not the `dist` version - should we test both do you think?

garrettmflynn: Since the executable environment should be scoped, I don't think it matters. It's just copied to another folder—which I've never had differing behavior from.

CodyCBakerPhD: But then, testing the `dist` is at least more accurate to what would be deployed with the rest of the app?

garrettmflynn: Technically yes. you can add an argument after the file path that targets that specific exe if you'd like

CodyCBakerPhD:
replaced by #297

PR #295, opened 2023-08-17
Remove unused commands from package.json

CodyCBakerPhD:
I don't think we use either of these in practice

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,8 +32,6 @@
     "deploy:electron:win": "electron-builder build --win --publish always",
     "deploy:electron:mac": "electron-builder build --mac --publish always",
     "deploy:electron:linux": "electron-builder build --linux --publish always",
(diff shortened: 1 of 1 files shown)


garrettmflynn (changes requested):
postinstall is actually a standard command that's triggered after installation. Interestingly enough, I just learned that you can append "pre" or "post" to any of the keys underneath the "script" field to trigger before/after the main script runs. 

The format command is fine to remove though.

CodyCBakerPhD:
> postinstall is actually a standard command that's triggered after installation.

So it must be included in the `package.json` in order to allow installation? Even though it's never called explicitly by us? Is that an `electron-vite` thing or just an `npm` thing?

CodyCBakerPhD:
Oh, I see

https://docs.npmjs.com/cli/v9/using-npm/scripts#pre--post-scripts

garrettmflynn (approved):
Looks good to me!

PR #283, opened 2023-08-11
Build and Deploy Windows

CodyCBakerPhD:
The workflow is green: https://github.com/NeurodataWithoutBorders/nwb-guide/actions/workflows/Build-and-deploy-win.yml

But when I install from the release: https://github.com/NeurodataWithoutBorders/nwb-guide/releases/tag/untagged-953a574ccfe5adc959bc

Everything goes well up until the connection to the back-end, which times out

My suggestion, since we all work on different devices, and since Mac is more touchy about signing issues than the other platforms, is that we focus on debugging the sister PR #284 for linux on the DANDI Hub (remote desktop), which is the common linux environment we all have access to

PR #296, opened 2023-08-17
[Simplify PyFlask I]: local scopes around problematic import

CodyCBakerPhD:
#291 is a sketch of the eventual goal with the pyflask refactor

But this is a simpler first step that doesn't break anything

This should also allow you to proceed attempting the linux distributable flask server by not attempting the Pydantic import on launch


pyflask/manageNeuroconv/info/urls.py
```diff
@@ -10,7 +10,7 @@ def resource_path(relative_path):
         # PyInstaller creates a temp folder and stores path in _MEIPASS
         base_path = sys._MEIPASS
     except Exception:
-        base_path = os.path.abspath(".")
+        base_path = Path(__file__).parent.parent.parent.parent
```
CodyCBakerPhD, line 13: I was unable to call `npm run dev:server` since it would make the working directory the pyflask directory, which was not the proper base to find the `paths.config.json` file

This still makes the internal logic of the function relative, but instead to this file itself in the dev folder structure

CodyCBakerPhD:
> This should also allow you to proceed attempting the linux (and other) distributable flask server by not attempting the Pydantic import on launch

@garrettmflynn Likewise, if you run into the scipy issue again, here's the list of hidden imports suggested: https://stackoverflow.com/questions/35060523/how-to-use-pyinstaller-with-hidden-imports-for-scipy-optimize-leastsq

pyflask/app.py
```diff
@@ -1,33 +1,38 @@
...
 
 configureLogger(app)
 
-api = configureAPI()
+package_json_file_path = Path(__file__).parent.parent / "package.json"
+with open(file=package_json_file_path) as fp:
+    package_json = json.load(fp=fp)
+
+api = Api(
+    version=package_json["version"],
```
CodyCBakerPhD, line 32: I noticed that if I ran `dev:server` the `version` would be `None`. Not sure if that environment variable would have been set somehow by the rest of the app build, but this approach is at least more consistent when running just the flask server alone

CodyCBakerPhD:
@garrettmflynn Ready for review now, I'll continue making changes in follow-up branches

pyflask/manageNeuroconv/info/urls.py
garrettmflynn: Nevermind, it seems like the `paths.config.json` file is fine here. So if this works for you, it's fine.

Gotta actually fix the `package.json` resolution in the same way.

garrettmflynn (approved):
Fixed the `package.json` resolution. This should be good to go.

garrettmflynn:
> > This should also allow you to proceed attempting the linux (and other) distributable flask server by not attempting the Pydantic import on launch
> 
> @garrettmflynn Likewise, if you run into the scipy issue again, here's the list of hidden imports suggested: https://stackoverflow.com/questions/35060523/how-to-use-pyinstaller-with-hidden-imports-for-scipy-optimize-leastsq

BTW tried this on the Windows branch (#283) with no luck.

PR #298, opened 2023-08-17
[Simplify PyFlask II]: remove excessive namespacing

CodyCBakerPhD:
@garrettmflynn You should double check me on this, but I'm 99% sure nothing is broken by this


garrettmflynn:
Yep tested the basics and fixed my merge. Should be good to go

PR #299, opened 2023-08-17
[Simplify PyFlask III]: remove setupUtils and unused startup endpoint(s)

CodyCBakerPhD:
Do we still need the echo endpoint in 'startup'? Do we ever use it? If not we can get rid of that sub namespace completely


garrettmflynn:
We do use it for checking if the Python server is actually started. We could simply maintain a variable in the `main` Electron process and pass this on completion—but actually getting a request to pass is the most foolproof method.

CodyCBakerPhD:
Ohhh right you are: https://github.com/NeurodataWithoutBorders/nwb-guide/blob/ea8fa6fc09112a99cad305b9e5a32af87695951c/src/renderer/src/index.ts#L128-L134

OK fine to keep then

garrettmflynn:
The alternative approach may actually allow us to catch a failed Python server rather than simply waiting for a timeout. Would we want to pursue that approach or nah?

CodyCBakerPhD:
> We could always send the echo request when notified to double-check.

I defer to you on the approaches that might be best for detecting such an event

In general I don't mind doing all of them, or as many as are useful for detecting different points of failure

My general approach right now is also just to massively simplify what is required to run the Flask server itself as opposed to the question of what is required to operate individual endpoint commands

garrettmflynn:
Okay sounds good. Will file this away as a possible improvement for later.

garrettmflynn:
Implemented on #301 to Implemented on #301 to help with debugging the Mac and other builds.

garrettmflynn (approved):
Looks good to me. Not using any of the deleted endpoints and the flattened app structure looks much better.

CodyCBakerPhD:
@garrettmflynn Sorry, need re-approval on this. I've disabled the stale dismission moving forward

garrettmflynn (approved):
No worries. Let's merge it in!

PR #300, opened 2023-08-17
[Simplify PyFlask IV]: add endpoints for dynamic installations

CodyCBakerPhD:
Tested these out via the direct Swagger page, seems to work in dev environment

Next going to test distributable version through PyInstaller


pyflask/apis/neuroconv.py
```diff
@@ -146,3 +146,31 @@ def get(self):
...
+
+
+@neuroconv_api.route("/test-custom-install/<string:package_name>")
+class TestCustomInstall(Resource):
+    @neuroconv_api.doc(responses={200: "Success", 400: "Bad Request", 500: "Internal server error"})
+    def post(self, package_name: str):
+        try:
+            import os
+
+            return os.system(f"pip install {package_name}")
```
CodyCBakerPhD, line 158: In before someone says something about how `os.system` is not ideal - I know this should be replaced with subprocess in some way but I ran into 2 issues with that on my end...

first being easy to fix, the output returned is not JSON serializable, so just don't bother returning the output of the call

second was really mysterious, something about subprocess not being able to find file referenced by `_win32` environment variables

pyflask/apis/neuroconv.py
```diff
@@ -146,3 +146,31 @@ def get(self):
...
+
+
+@neuroconv_api.route("/test-dynamic-import/<string:package_name>")
+class TestDynamicImport(Resource):
+    @neuroconv_api.doc(responses={200: "Success", 400: "Bad Request", 500: "Internal server error"})
+    def post(self, package_name: str):
+        try:
+            from importlib import import_module
+
+            remfile = import_module(name=package_name)
```
CodyCBakerPhD, line 171:
```suggestion
            package = import_module(name=package_name)
```

Ahh was testing with remfile, but doesn't have to be...

pyflask/apis/neuroconv.py
garrettmflynn: Huh weird. As long as this works for now, we can simply relegate this to a future issue—if we care to track it at all.

garrettmflynn:
Looks good to me. Any luck with dynamically installing alongside the executable?

CodyCBakerPhD:
This one is still experimental, will come back to it next week

CodyCBakerPhD:
While this worked in `dev` mode (where the Flask server is really running on top of the native conda environment) this works, because new packages just get installed to that conda environment.

When I test this with the final distributable however, the PyInstaller python executable does not share the same behavior

CodyCBakerPhD:
We aren't the first to ask if this is possible

Two stackoverflows

https://stackoverflow.com/questions/66590199/can-you-use-pip-from-within-a-pyinstaller-executable

https://stackoverflow.com/questions/75719852/does-pyinstaller-support-installing-dependent-packages-via-pip-on-first-use

No accepted solutions, a lot of hypothesizing

PR #301, opened 2023-08-17
Monitor status efficiently

garrettmflynn:
This PR allows us to monitor internet and backend status efficiently and continuously.

Turning on/off Wifi will now result in an updated status icon and a pop-up window if detected off.

We will now know immediately if the Flask server fails and errors will be relayed to the frontend. No polling is required to detect a successful launch—so we only have to send on request to determine if it is working.

These behaviors are now much cleaner than I'd expected, and it was quite simple to implement these features. No doubt, they'll help debug the production builds of these apps.


CodyCBakerPhD (approved):

CodyCBakerPhD:
Awesome! Thanks for improving that

PR #284, opened 2023-08-11
Linux deployment

CodyCBakerPhD:
Replaces #257 

Sister PR to #283

I suggest, since we all work on different devices, and since Mac is more touchy about signing issues than the other platforms, we should focus on debugging this PR first on the DANDI Hub (remote desktop), which is the common linux environment we all have access to

Only problem right now is getting the deployment working, could use some help on it: https://github.com/NeurodataWithoutBorders/nwb-guide/actions/workflows/Build-and-deploy-linux.yml

PR #302, opened 2023-08-18
[Simplify PyFlask V]: full local scope for all utils

CodyCBakerPhD:
This should fix the scipy problem with regards to the Flask server launching

It does _not_ fix the actual problem when the import is triggered by an endpoint that needs to run a Python function using any library stemming from HDMF, but at least this gives us a better capacity to debug live on the Flask server


CodyCBakerPhD:
Windows passing, Linux awaiting hopeful fix on #284

garrettmflynn:
Still have that issue with `pydantic[email]`. Maybe that will also go away?

CodyCBakerPhD:
> Still have that issue with pydantic[email]. 

That should even have been fixed as of #296 

Can you re-setup your environment and re-build the build/dist from scratch and then include full traceback for this branch?

Nothing anywhere should be triggering it on launch since the import is only ever triggered on call to the upload endpoint

CodyCBakerPhD:
Oh wow, it's actually in the CI over on https://github.com/NeurodataWithoutBorders/nwb-guide/actions/runs/5905804597/job/16020592221?pr=284

investigating...

CodyCBakerPhD:
Oh I see. Outer import of YAML functionality

https://github.com/catalystneuro/neuroconv/blob/main/src/neuroconv/__init__.py#L5

triggers some minimal DANDI import

https://github.com/catalystneuro/neuroconv/blob/main/src/neuroconv/tools/yaml_conversion_specification/yaml_conversion_specification.py#L7-L8

Sorry I missed that

The conclusion is that #284 will need this branch then to fully resolve that launch issue; fix for using the endpoint remains elusive but the recent front-end additions will at least let the person know why they can't do that one step inside the GUIDE

CodyCBakerPhD:
@garrettmflynn All green

We can even, for the first time, check this with CI deployment to the release page

Parts of the Guided Mode may not work on all platforms, which is still next steps to debug, but this gives us a better position to debug and potentially fix those dynamically using #300 if it works

garrettmflynn (approved):
Woohoo! Good thinking with this, let's see what we can do

PR #303, opened 2023-08-18
Move Freeze Up 

garrettmflynn:
Ensure issues with runaway thread creation are resolved for the PyInstaller build


CodyCBakerPhD (approved):

CodyCBakerPhD:
Interesting that this fixes it. Strange behavior observed exclusively to PyInstaller build on a Flask server with lots of imports... 

To summarize renegade threads would spawn performing some task and were rather hard to kill; `sudo pkill -9 "new-guide"` worked of course but felt unnecessary.

PR #305, opened 2023-08-18
Replace HTML representation of the NWB file with Neurosift

garrettmflynn:
This PR uses an iframe to point the Neurosift application to the locally hosted stub file being previewed to provide a comprehensive look at the contents of the file.

### DIFF
--- a/pyflask/manageNeuroconv/manage_neuroconv.py
+++ b/pyflask/manageNeuroconv/manage_neuroconv.py
@@ -236,7 +236,6 @@ def validate_metadata(metadata: dict, check_function_name: str) -> dict:
 
 def convert_to_nwb(info: dict) -> str:
     """Function used to convert the source data to NWB format using the specified metadata."""
-    from pynwb import NWBHDF5IO
 
     nwbfile_path = Path(info["nwbfile_path"])
     custom_output_directory = info.get("output_folder")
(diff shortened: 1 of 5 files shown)


CodyCBakerPhD:
Looks great! Slight delay while it launched but then was A-OK

One question maybe; is it possible for `neurosift`, or the way you've integrated it, to fail?

And if so, is it possible to catch that failure and instead fallback to the basic HTML display in that event (basically treating this like icing on the cake, but always offering at some some minimal representation of the output)

CodyCBakerPhD (approved):

CodyCBakerPhD:
Actually storybook confirms what happens in event of failure I suppose: https://www.chromatic.com/test?appId=640a4d93e34604f275368983&id=64e38b652c12a1ee283afdde

but IDK if that's because the input to that page is not suitable for the story or what-not

garrettmflynn:
As far as I can think, there is only one case where the Neurosift iframe will fail and we can still get the HTML representation: when there is a bug introduced on the Neurosift site itself that is undesirable.

All other cases (e.g. server down, file not created, etc.) will result in us being unable to get any representation of the file.

garrettmflynn:
As mentioned in the last meeting, the clearest case where this could be useful is when internet is down—so we can't actually gain access to the Neurosift URL.

PR #315, opened 2023-08-21
Indicate that the list of interfaces is loading

garrettmflynn:
This PR accounts for the delay in the list population given that the server endpoint now has a startup time. 

I'm waiting for the server to return the list of valid interfaces before I render any previously selected ones. This ensures an appropriate message is provided to the user, indicating this operation is in progress—though if desired, I can just render the unvalidated list.

The Add Interface button is also hidden from the user until the search list is populated.


CodyCBakerPhD:
There's an alternative to this wait time I think we should try first - can you try to trigger `import neuroconv` as an isolated operation after initial startup of the app?

The thing about Python imports (as long as the Flask is constantly running anyway) is that there is only a delay to import something heavy the _first time_ it is ever imported, even by local scope

garrettmflynn:
You're right. Adding `neuroconv` to the top of `app.py` results in a near-instantaneous response. 

It may still be useful to have these changes in case the response takes longer than expected (e.g. if we end up hosting the backend remotely and make this request over a slow internet connection)—though we don't necessarily have to.

CodyCBakerPhD:
Yeah I'll still check it out

> Adding neuroconv to the top of app.py results in a near-instantaneous response.

Right but that runs the risk of causing the same issues we've been fixing - I'm saying still use a local scope - but inside a new endpoint that would get triggered in a safe/isolated way on server startup, if that's possible. In case the import fails due to bad environment, only want the endpoint to fail (and even then maybe silently since user didn't explicitly say to do it) not the server itself

garrettmflynn:
Was originally intending to test using #300. This would be the ideal way to do it, right?

If so, I can simply change the base for this PR.

CodyCBakerPhD:
I'm thinking I should review this and #324 before #326, based on the number of conflicts across all 3 (most in #326)?

Let me know when resolved

garrettmflynn:
Alright this is up-to-date. Closed #324 and reimplemented in #326 (also now current) to align with updated request results.

CodyCBakerPhD (approved):

PR #316, opened 2023-08-21
Run all stubs and preview the first

### LINKED ISSUE #223: Ensure that the user previews a session before moving onto the full conversion
If we'd like to ensure that a user reviews a stub preview before moving onto the full conversion, we should block their progress from the File Metadata page until they press the Preview button for a session they've finished editing.

Since this review process is a best practice and, really, one that is easy to ignore if automated anyways, we don't necessarily have to block their progress—and instead might just be more clear about this expectation to preview before moving on.

garrettmflynn:
fix #223.

All stub files are created and the first stub is always previewed. 

We _could_ instantiate a Neurosift preview in an Instance Manager (used for File Metadata) so users can peruse all stub files. Would this be a desirable feature? There would also be other UI options we could adopt—though this would certainly be the fastest to get up and running.


CodyCBakerPhD:
> We could instantiate a Neurosift preview in an Instance Manager (used for File Metadata) so users can peruse all stub files. Would this be a desirable feature?

Could be interesting to try out in a separate draft PR, see how it looks/feels

CodyCBakerPhD (approved):

CodyCBakerPhD:
Nice

Not so much applicable to the tutorial since it just copied the same data 4 times, but in practice not every file is created equally and running a complete stub conversion over all sessions lets us catch any errors that might occur in outlier sessions

Granted, if say 1 out of 4 sessions had an error, the entire operation from the front-end would only report one error, and would not allow navigating to the next page, right?

An improvement for a follow-up could be to along the lines of thought of your question and other PRs you've started, but with the addition that if a particular session errored for a particular reason, then sessions without errors still preview fine but the session with the error renders a window with the traceback instead of neurosift or inspector

PR #297, opened 2023-08-17
Create pyflask-build-and-dist-tests.yml

CodyCBakerPhD:
Re-attempt #292 from a better base

PR #320, opened 2023-08-21
Create a new script to run tests on the executable

garrettmflynn:
fix #297

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,6 +24,7 @@
     "test": "npm run test:app && npm run test:server",
     "test:app": "vitest run",
     "test:server": "pytest pyflask/tests/ -s",
+    "test:executable": "pytest pyflask/tests/ -s --executable",
     "test:coverage": "npm run coverage:app && npm run coverage:server",
     "coverage:app": "vitest run --coverage",
     "coverage:server": "pytest pyflask/tests/ -s --cov=pyflask --cov-report=xml",
--- a/pyflask/tests/conftest.py
+++ b/pyflask/tests/conftest.py
@@ -1,23 +1,60 @@
 import pytest
 import app as flask
+import sys
+import subprocess
+from pathlib import Path
+from multiprocessing import Process
+import multiprocessing as mp
+import os
+import time
+
+executable_flag = '--executable'
+use_executable = executable_flag in sys.argv
+is_windows = os.name == 'nt'
+
+args = list(sys.argv)
+
+PORT = 1234
+exe_process = None
+
+def pytest_addoption(parser):
+    parser.addoption("--executable", action="store_true", help='Run the executable instead of the standard Flask app')
+
+@pytest.fixture(scope="session")
+def client(request):
+    exe_path = Path("build/flask/nwb-guide", 'nwb-guide.exe' if is_windows else 'nwb-guide' ) 
+    if (use_executable and exe_path.is_file()):
+
+        if (not is_windows):
+            mp.set_start_method('fork')
+
+        def runExecutable(port):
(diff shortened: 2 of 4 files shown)


garrettmflynn:
@CodyCBakerPhD Are you familiar with catching errors from a subprocess? I tried doing this for one of the recent pushes—but it doesn't actually seem to be doing much.

CodyCBakerPhD:
> Are you familiar with catching errors from a subprocess?

I've done it before but it gets sophisticated with passing/controlling a system stdout or stderr PIPE object

I'd say try `os.system` instead, it returns `0` or `1` for success/failure (not necessarily in that order, or if it even works that way)

CodyCBakerPhD:
> I've done it before but it gets sophisticated with passing/controlling a system stdout or stderr PIPE object

Even then the PIPE just passes the in/out streams as string, so IDK about re-raising error outputs through that

Is it possible to asynchronously launch the Flask server via a CLI call/script, then wait a few seconds, then call pytest directly? `pytest` definetely works very well when called at the top level (not via subprocesses)

garrettmflynn:
Well, I guess before we move forward, we should define something: Do you want Pytest to fail because (1) the executable threw an error, or (2) the endpoint did not return the expected value. 

For some reason, things are still passing even though the endpoint presumably did not return anything. IMO I should probably fix that issue first—and then we can use the dev logs (which are visible, but not interrupting the main process) to fix the endpoint.

CodyCBakerPhD:
> Do you want Pytest to fail because (1) the executable threw an error, or (2) the endpoint did not return the expected value.

If the executable has a problem separate from the testing suite, that should be caught ahead of time and reported separate I'd think

Otherwise we want pytest to fail in the same way it would fail as it is currently run on the dev Flask version with regard to all the tests that we run 

If there's any weird interaction that is different from running in dev mode, I'd probably need to see a specific example but in general if there's any kind of error, it should let that be known to the CI (causing a failure in CI environment so we get the red X on GitHub)

garrettmflynn:
Looks like we have the intended behavior now? I've also improved the configuration of the tests—including some updates to `testPyinstallerExecutable.js`—based on your comments.

package.json
```diff
@@ -24,6 +24,7 @@
     "test": "npm run test:app && npm run test:server",
     "test:app": "vitest run",
     "test:server": "pytest pyflask/tests/ -s",
+    "test:executable": "concurrently -n EXE,TEST --kill-others \"node tests/testPyinstallerExecutable.js --port 1234 --forever\" \"pytest pyflask/tests/ -s --target http://localhost:1234\"",
```
CodyCBakerPhD, line 27: Wow uhh... never seen this before

Is this an NPM thing? (`concurrently`)

garrettmflynn: No, it's a library that lets you easily run NPM commands in parallel. It's useful when one command blocks the main process (i.e. any server).

We were already using it to run the app and server together.

CodyCBakerPhD: Ahh, looks like CI doesn't like: https://github.com/NeurodataWithoutBorders/nwb-guide/actions/runs/5942246443/job/16114801620?pr=320

Maybe the package lock needs to be updated to include it?

CodyCBakerPhD: Or, ohhh I see. That CI was originally only designed to run Python but now it needs to be adjusted to setup `npm` and such too

Guessing it just needs a `npm ci` call in an earlier step

CodyCBakerPhD:
Fantastic, thanks so much! This is great and will allow me or Ryan to focus solely on debugging those specific issues while you can go full steam ahead on the app in the dev environment

PR #322, opened 2023-08-22
Instance Manager to View All Stubs

garrettmflynn:
This PR allows users to view all stubs through an instance manager, as discussed in #316.

### DIFF
--- a/src/renderer/src/stories/InstanceManager.js
+++ b/src/renderer/src/stories/InstanceManager.js
@@ -75,6 +75,10 @@ export class InstanceManager extends LitElement {
                 height: 100%;
             }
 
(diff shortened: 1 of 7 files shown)


CodyCBakerPhD (approved):

CodyCBakerPhD:
Looks/feels really good; though a very large dataset might lag as per discussion

Implementing lazy loading per tab would be ideal, perhaps with some limited/throttled async pre-rendering if not too difficult to setup

I also noticed that, after disabling my internet and rerunning the conversion, I was still able to view the files in neurosift - I thought that was supposed to default to the HTML? Or maybe neurosift got cached somehow the first time?

PR #324, opened 2023-08-22
Fix Results Page

garrettmflynn:
This PR fixes the results page now that the only returned object from the conversion is the final file path.


garrettmflynn:
Incongruous with #326

garrettmflynn:
Incongruous with https://github.com/NeurodataWithoutBorders/nwb-guide/pull/326

PR #325, opened 2023-08-22
Standalone DANDI Upload

### LINKED ISSUE #308: Stand-alone DANDI upload
Since (at least in dev mode) we have working automatic DANDI upload at the end of the guided pipeline, it should be easy to add a tab to the main page for a simple stand-alone version of that page where the user specifies a folder of NWB files and wants to use the NWB GUIDE (instead of DANDI CLI) to upload everything to the archive

garrettmflynn:
fix #308


garrettmflynn:
@CodyCBakerPhD Noticed I'm generally getting an error when trying to upload to DANDI—though I feel like I've seen it before. Can you confirm by testing the new page?

CodyCBakerPhD:
> Can you confirm by testing the new page?

I caught several issues with the new page; actually unable to even execute the upload command itself

a) in the folder selector, I cannot see any `.nwb` files in the folders I point it to
b) if entered first, the DANDI set ID gets wiped after selecting the folder
c) the red invalid error that show beneath the missing DANDI set ID renders twice
d) the red invalid errors do not disappear when I enter the DANDI set ID; only when I hit the upload button
e) trying to upload, unable to find the API key (first time, this is fair)
f) I set the API key in the settings page (looks very cool BTW!) and save, everything looks green
g) Try upload again; unable to find the API key; infinite loop back to settings page

https://github.com/NeurodataWithoutBorders/nwb-guide/assets/51133164/cd2b0d58-82fb-47c5-9e5e-ee11fdf08e15

garrettmflynn:
@CodyCBakerPhD 

> d) the red invalid errors do not disappear when I enter the DANDI set ID; only when I hit the upload button
e) trying to upload, unable to find the [API](https://docs.github.com/) key (first time, this is fair)
f) I set the API key in the settings page (looks very cool BTW!) and save, everything looks green
g) Try upload again; unable to find the API key; infinite loop back to settings page

This behavior happens because the value only updates when the input element is unfocused (i.e., clicked off of)—and the button onClick event appears to happen faster than that. 

As for the other issues, (a) appears to only be an issue on Windows and I can't find any way around it. Only `openFile` or `openDirectory` can be specified at one time, and the latter appears to hide all files on Windows. 

(b) has been addressed by flipping the position of the NWB Folder selector and the Dandiset ID input. 

I haven't been able to replicate (c), though it may have been resolved by my changes as the errors are much more consistent now.
I've tried to implement fixes for (a) - (c). Can you confirm the expected behavior is now occuring?

CodyCBakerPhD (approved):

CodyCBakerPhD:
TODO for follow-ups

- fix the guided mode DANDI upload project page to not have to ask for the NWB folder path since we already know it from the project

- draft Figma for display/tracking of results and history

Anything else?

PR #334, opened 2023-09-05
Update file handling to support and block multiple file uploads

garrettmflynn:
This PR allows users to specify that they'd like to specify multiple files for the FileSystemSelector using the JSONSchemaForm / Input elements. 

Additionally, dragging the incorrect number / type of files into the element will now be blocked—where this previously was not implemented because this is automatically handled by the native selector popup.


CodyCBakerPhD:
@garrettmflynn One conflict here

Also, how would you recommend going about testing it? (on my end, in the suite, etc)

garrettmflynn:
Just added comprehensive Stories, so you can run `npm run storybook` to interact with them all. 

The GUIDE has a custom onThrow callback added that results on a popup being shown— but these look ugly in Storybook so they've been omitted. You can still view the errors thrown in the Dev Console.

CodyCBakerPhD:
Since we don't have any multi-file select mechanisms that really depend on that yet (such as interfaces that take `file_paths`) it's hard to test that particular feature

Just testing this to make sure it doesn't break anything and here's one or two things

i) Trying a file path based interface looks OK at the beginning, but once I select a file the box disappears (does not tell me what path I selected

image




ii) The stand alone NWB Inspector page is now only the 'inspect button' but clicking on that says no paths are selected

image

garrettmflynn:
This should be ready to go now.

CodyCBakerPhD:
New conflicts - all tests were also failing specific to this PR as of last commit

CodyCBakerPhD (approved):

CodyCBakerPhD:
Looks like some backend tests failing now: https://github.com/NeurodataWithoutBorders/nwb-guide/actions/runs/6189341651/job/16803234661?pr=334

CodyCBakerPhD:
But running the app felt fine, didn't notice any problems

garrettmflynn:
Alright local checks passed

PR #335, opened 2023-09-05
Validate DANDI API token

garrettmflynn:
This PR introduces changes to validate the DANDI API token to ensure that it is properly provided by the user.


src/renderer/src/stories/pages/settings/SettingsPage.js
```diff
@@ -18,6 +18,8 @@ import { merge } from "../utils.js";
 
 import { notyf } from "../../../dependencies/globals.js";
 
+const apiRegex = /^[a-f0-9]{40}$/;
```
CodyCBakerPhD, line 21: I would name this a bit more specifically to DANDI; i.e., `dandiApiTokenRegex`

Also, just curious what your preference/opinion of standard is for capitalizing abbreviations in lowerCamelCase? I.e., `dandiAPITokenRegex`?

garrettmflynn: Sounds good. I'd say I prefer `dandiAPITokenRegex` to `dandiApiTokenRegex`

CodyCBakerPhD (approved):

CodyCBakerPhD:
Yep, looks good. Technically the error message could be improved over the basic 'invalid value' but attaching instructions as per #339 should be enough for that

Issue #312, opened 2023-08-21
Stand-alone neurosift launcher


CodyCBakerPhD:
Similar to #308, would be great to have an extra tab on the left that exists simply to navigate the local file system, select an NWB file, and run local neurosift window on that file

This would also be the main mechanism that would be launched by double-clicking an `.nwb` file outside the app

PR #336, opened 2023-09-05
Create a standalone NWB Inspector page

garrettmflynn:
This PR creates a standalone NWB Inspector page similar to the DANDI Upload page. Now that we are inspecting files before preview, this is adjacent to #312—but not directly within scope of this issue.


CodyCBakerPhD:
Tried it out on a real DANDI set (locally downloaded) and worked great

Future items, as discussed, might be a separate button for single file (maybe shift/ctrl clicked multiple files, too?) vs folder mode

Would apply to DANDI upload as well

CodyCBakerPhD (approved):

PR #338, opened 2023-09-06
Fix Upload Page (Schema Mutation + Button Text)

garrettmflynn:
This PR fixes the behavior of the Guided DANDI Upload page noted on #325.


CodyCBakerPhD:
@garrettmflynn Just out of curiosity, have you seen these Inspector report items before? Not sure if it's the way the folder is being passed or something in the Inspector, but it seems to be scanning something twice

image

Also the file paths for the messages below the red ones seem to route to a tmp directory, not the current stub location in the main GUIDE folder, any idea what that's about?

CodyCBakerPhD:
The discovery above is tangential to this fix however; the fix to the conversion upload page looks good now

CodyCBakerPhD (approved):

PR #341, opened 2023-09-06
Add Support Table to Read the Docs

garrettmflynn:
This PR adds an embedded Google Sheet that lists our supported formats throughout the NWB ecosystem.


docs/support.rst
CodyCBakerPhD, line 1: minor thing, but I'd call this `.rst` the same as the title to make it clear its not a typical 'support' section (like, how to get help with an issue)

E.g., `format_support.rst`

CodyCBakerPhD:
I like the approach of having these tabs

image

but is it possible to move them to the top, make them bigger, or more nicely formatted (less HTML-like), like they are on the sheet itself

image

CodyCBakerPhD:
It looks like the embed also has some rendering issues

image

whereas the view in sheets

image

So I'm curious, if we _don't_ opt for the direct embed, what other options are you able to put together on a different branch?

What's the best looking analogous table you can make in JS/TS/ raw HTML / other sphinx things?

garrettmflynn:
I've created a new repository that simply hosts our derivative representation of the Ecosystem Support Table: https://garrettflynn.com/nwb-support-table/. This website can then be used in the `<iframe>` instead of the Google Sheets widget.

I've handled the data structure manipulation—so we just need to style this in an agreeable way. What do you think?

CodyCBakerPhD:
@garrettmflynn That is indeed much cleaner, and even gives a more collaborative and better version controlled approach to managing the table - thanks! Would you prefer to use Figma to prototype/communicate visual style improvements?

garrettmflynn:
Depending on the scope of your suggestions, Figma might be better—as there's less room for interpretation.

Otherwise starting here is fine with me.

CodyCBakerPhD:
What's the link to the new repo?

garrettmflynn:
https://github.com/garrettmflynn/nwb-support-table

CodyCBakerPhD:
Cool, I forked: https://github.com/catalystneuro/nwb-support-table

Mind transferring ownership to CN?

garrettmflynn:
lol I can't now because the name of your fork conflicts with the base repo. Would you want me to change the name or would you like to delete the fork instead?

CodyCBakerPhD:
I just removed it instead - all we need now is to update the table link on this PR and good to go?

garrettmflynn:
Actually it's all good. Just updated to the correct URL

CodyCBakerPhD (approved):

PR #342, opened 2023-09-06
Preload imports on Flask server

### LINKED ISSUE #340: Preload imports on startup from safe endpoint
@garrettmflynn Can you point me to the relevant code on the front-end that launches the first call to the Flask endpoint? (I think it used to have something about 'echo'? Or maybe it's the thing that checks if the backend is connected and running?)

I'd like to add a simple endpoint to trigger various imports just after launch (but as separate hidden operation) to reduce the instantaneous speed of the now locally scoped operations

CodyCBakerPhD:
fix #340 

Current draft for how I imagine this would go - trying to think of a way to test it, but IDK if it's really possible aside from noticing that the 'loading interfaces' page should now be instantaneous

@garrettmflynn This would also trigger on hot reload right?


garrettmflynn:
Looks good. You're correct, this would trigger on hot reload.

I think your "test" is on the mark. Failure is easy to see too—but success requires a bit more. You _could_ send something back to the frontend from the import and check for that

CodyCBakerPhD:
> You could send something back to the frontend from the import and check for that

I technically return `True` from the endpoint success, would that be something you could develop a test for? Or would you prefer a better error/success code (like an integer)

garrettmflynn:
`True` is fine. Then the enpoint will either return that or an inbuilt error. 

Though what is the condition(s) you're actually interested in testing for? Whether the import itself errors?

CodyCBakerPhD:
> Though what is the condition(s) you're actually interested in testing for? Whether the import itself errors?

Yeah, that's as close as we can get. If that sends back `True` it means its cached in memory on the Flask server, so the speed benefit _should_ work indirectly as a consequence

garrettmflynn:
Awesome. If you want to actually write a Python test, you can follow the format noted here: 

https://github.com/NeurodataWithoutBorders/nwb-guide/blob/c5e778ab7cf05f18df3115853aab5c2e86a3273f/pyflask/tests/test_neuroconv.py#L7

You don't need the validate() call since it is sufficient that the endpoint doesn't error

garrettmflynn:
Oh sure. You can check if True is returned from the request—though that would be it

CodyCBakerPhD:
@garrettmflynn Backend status symbol on the app is yellow and dev console outputs

```javascript
Uncaught (in promise) TypeError: Cannot convert undefined or null to object
    at GuidedStructurePage.updated (GuidedStructure.js:80:20)
    at GuidedStructurePage._$didUpdate (reactive-element.ts:1380:10)
    at GuidedStructurePage.performUpdate (reactive-element.ts:1345:12)
    at GuidedStructurePage.scheduleUpdate (reactive-element.ts:1263:17)
    at GuidedStructurePage.__enqueueUpdate (reactive-element.ts:1235:25)
```

Ideas how to fix?

garrettmflynn:
Hmmm how are you getting this? Just ran through with the Tutorial pipeline from scratch and didn't see this.

CodyCBakerPhD:
My workflow

-> `git checkout preload_imports`
-> `git pull`
-> `npm start`
-> backend status light is yellow
-> dev console says API commands were sent to endpoint and look fine
-> navigate any pipeline from scratch up to interface page
-> no interfaces, dev console spits out the above error

Actually, I just checked and I'm seeing this on `main` now as well, perhaps an issue with a recent merge? And one that's Windows specific?

garrettmflynn:
Do you notice anything odd going on with the backend in the main console? 

Otherwise I'll try to see if I can trigger this myself tomorrow.

CodyCBakerPhD:
The only thing in the command prompt out of the ordinary is

```
Skip checkForUpdatesAndNotify because application is not packed
```

no other activity as I go through the first few pages of a pipeline

garrettmflynn:
Alright fixed. There were actually two separate issues here:

1) You were returning after calling `preloadFlaskImports`, interrupting the status update that indicates a successful backend connection.

2) Not having interfaces defined on the global object would result in the error you saw in `GuidedStructure`. I've made sure this is filled in if undefined

CodyCBakerPhD:
Ohhh: https://github.com/NeurodataWithoutBorders/nwb-guide/pull/342/commits/e1f13248d790b8cfef0825e3eb7d74adb944a042#diff-28a5296a524fc5a6da149c021fb048736dd8586329a8d9095bb79c4c78d4f011R112

I see, thanks for debugging my JavaScript!

CodyCBakerPhD:
OK, can you give this a try when you review? Works well on my end

The connect to the backend now takes just a slight moment to go from yellow to green (the time it would have taken on the first screen running a backend operation) and as such if you, for example, navigate to the interfaces screen of a fresh pipeline, it is now instantaneous

garrettmflynn (approved):
Yep, looks good!

Issue #89, opened 2023-03-28
[Mac]: Python server unable to be built with Pyinstaller


garrettmflynn:
Attempting to run `python-onefile-build` fails to output anything on Mac.

Below is the output to the terminal until an exit code is thrown:

```
1495 INFO: PyInstaller: 4.7
1495 INFO: Python: 3.9.7 (conda)
1528 INFO: Platform: macOS-10.16-x86_64-i386-64bit
1529 INFO: wrote /Users/garrettflynn/Documents/Github/nwb-guide/app.spec
1561 INFO: UPX is not available.
1562 INFO: Removing temporary files and cleaning cache in /Users/garrettflynn/Library/Application Support/pyinstaller
1570 INFO: Extending PYTHONPATH with paths
['/Users/garrettflynn/Documents/Github/nwb-guide/pyflask']
2821 INFO: checking Analysis
2821 INFO: Building Analysis because Analysis-00.toc is non existent
2821 INFO: Initializing module dependency graph...
2825 INFO: Caching module graph hooks...
2834 WARNING: Several hooks defined for module 'numpy'. Please take care they do not conflict.
2846 INFO: Analyzing base_library.zip ...
5322 INFO: Processing pre-find module path hook distutils from '/opt/anaconda3/envs/env-electron-python/lib/python3.9/site-packages/PyInstaller/hooks/pre_find_module_path/hook-distutils.py'.
5455 INFO: distutils: retargeting to non-venv dir '/opt/anaconda3/envs/env-electron-python/lib/python3.9'
8663 INFO: Caching module dependency graph...
8815 INFO: running Analysis Analysis-00.toc
8827 INFO: Analyzing /Users/garrettflynn/Documents/Github/nwb-guide/pyflask/app.py
13879 INFO: Processing pre-safe import module hook six.moves from '/opt/anaconda3/envs/env-electron-python/lib/python3.9/site-packages/PyInstaller/hooks/pre_safe_import_module/hook-six.moves.py'.
14659 INFO: Processing pre-safe import module hook urllib3.packages.six.moves from '/opt/anaconda3/envs/env-electron-python/lib/python3.9/site-packages/PyInstaller/hooks/pre_safe_import_module/hook-urllib3.packages.six.moves.py'.
19687 INFO: Processing pre-find module path hook site from '/opt/anaconda3/envs/env-electron-python/lib/python3.9/site-packages/PyInstaller/hooks/pre_find_module_path/hook-site.py'.
[...]

PR #346, opened 2023-09-06
Trying to fix Windows and Linux distributable 

CodyCBakerPhD:
Trying to fix windows side of current distributable errors related to Pydantic


CodyCBakerPhD:
@garrettmflynn 3 tests are passing according to log output for Windows here: https://github.com/NeurodataWithoutBorders/nwb-guide/actions/runs/6102263434/job/16560353693?pr=346

But CI itself is failing and says there are 'Errors' without reasons or tracebacks, any idea how to interpret this?

garrettmflynn:
This looks like it's fine. The executable test suite finishes without issue, causing the `.exe` to close—though this action seems to be something that registers as an error for the workflow. 

The only outstanding issue is that an Error Code 1 is thrown by the latter. Not sure why or how we could figure that out—but I'll look into it

CodyCBakerPhD:
> This looks like it's fine. The executable test suite finishes without issue, causing the .exe to close—though this action seems to be something that registers as an error for the workflow.
>
> The only outstanding issue is that an Error Code 1 is thrown by the latter. Not sure why or how we could figure that out—but I'll look into it

Fantastic! Exciting.

Yeah if you can figure out how to make the CI give the glorious green check that would be great, then I'll move onto the other platforms

garrettmflynn:
Found a solution on this issue: https://github.com/open-cli-tools/concurrently/issues/187

Turns out this is a common problem, I'd tried using the `--success` flag originally but it turns out `first` means the first to _exit_ when I thought it meant the first process in the list of those to run

garrettmflynn:
@CodyCBakerPhD Worked for both Windows and Linux! Unfortunately, we're running into the same Mac issue that @rly  originally solved in #89...

CodyCBakerPhD:
> @CodyCBakerPhD Worked for both Windows and Linux! Unfortunately, we're running into the same Mac issue that @rly originally solved in https://github.com/NeurodataWithoutBorders/nwb-guide/issues/89...

Awesome, thanks a bunch! We can tackle Mac in a follow-up then

rly (approved):

PR #347, opened 2023-09-07
Remove all internal files / folders related to a conversion

### LINKED ISSUE #344: [Bug?] Deletion of pipeline does not clear folder contents in `stub`
I didn't check the other content folders, but the following workflow can reproduce this

-> open the stub folder for the GUIDE
-> delete existing tutorial pipeline
-> folder and its contents are still present
-> so if I regenerate a new tutorial pipeline, it still has all the previous contents

These should be cleaning up on deletion, right?

garrettmflynn:
fix #344


rly:
Looks good and works fine locally on my end.

rly (approved):

PR #349, opened 2023-09-07
Ensure pages always save once

garrettmflynn:
This PR ensures that pages always save once, populating the required data structures for progress through Guided Mode.

This became apparent when trying to refresh the File Metadata page when running straight through a fresh Tutorial pipeline, which resulted in an error because one of the global metadata properties was missing.

Moving forward, this allows us to move certain operations from the `onNext` callback into the `beforeSave` such that they are only run when data actually changes.


CodyCBakerPhD:
One conflict here - also how would you recommend I test this change?

CodyCBakerPhD:
Yup, that one specific issue on the file metadata page appears to be resolved with this

CodyCBakerPhD (approved):

PR #350, opened 2023-09-07
Handle Files and Folders Together

garrettmflynn:
This PR allows you to handle files and folders in the same input.

Appropriate Stories have been created for testing purposes.


CodyCBakerPhD:
The stories look good

In a follow-up I'd like to see this implemented for the DANDI upload standalone page (as well as the NWB Inspector page, separate PRs for each)

The logic would be as follows:

if multiple files are selected, a new folder is created somewhere in the app background (similar to the stub folder) and symlinks are made in that folder that point to the files selected - then, the folder-based Flask call takes that temporary folder as input to perform the upload or inspection

CodyCBakerPhD (approved):

PR #360, opened 2023-09-13
Initialize without activate call

### LINKED ISSUE #354: [Bug - Mac Release on M1 and Windows] Distributable not able to run on main (9/11/2023)
Original comment chain starting at https://github.com/NeurodataWithoutBorders/nwb-guide/pull/337#issuecomment-1711919404

Just reproduced it on the DANDI Hub, the one shared computing environment we all have access to; I suggest we start there and see if fixing that fixes the other platforms

I also wasn't able to run dev mode either on the Hub - I'm pretty sure both used to work, up to those previous backend distribution issues, right?

To be clear, workflow is

1) fresh clone of `nwb-guide`
2a) `npm start` to ensure dev mode is working
2b) might need to do `npm ci` if `npm start` doesn't work due to 'electron-vite' not being found
3) `npm run build:<platform>`
4a) Windows: run setup `.exe`, follow instructions, then try to run application from the installed location
4b) Mac: `.dmg` to App folder, try to run from there
4c) Linux: Configure `.AppImage` to run as executable, then try to run it (on the Hub, do so from the remote desktop)

Might be related to issues seen by @alessandratrapani on ubuntu?

garrettmflynn:
Attempting to fix #354 by initializing the window regardless of platform

### DIFF
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -383,7 +383,7 @@ function makeSingleInstance() {
   else app.on("second-instance", () => restoreWindow());
 }
 
-if (process.platform === 'darwin') initialize();
+initialize();
 
 app.on('activate', () => {
   if (BrowserWindow.getAllWindows().length === 0) initialize()


CodyCBakerPhD:
Voila, frontend launches now

So glad it was something that simple

I'll merge this in to get the ball rolling

I do however still have a problem now with the Windows dist in that it won't activate the backend; can investigate in follow-up

Don't know if it's related but I do now still see the good old message

image

garrettmflynn:
Argh. Usually that means the path to the executable is incorrect.

Thankfully we have great error-checking between the frontend and electron now—so we can log everything to see if this is the case.

PR #361, opened 2023-09-13
Lower version pin on pydantic

CodyCBakerPhD:
Totally my bad for not catching this earlier

Now that it's not a `pip` dependency the `pip` resolution detector doesn't complain about an incompatibility; DANDI relies on `pydantic < 2.0` for the time being

Hopeful this will perhaps resolve the recursions depth issue too

### DIFF
--- a/environments/environment-Linux.yml
+++ b/environments/environment-Linux.yml
@@ -9,7 +9,7 @@ dependencies:
   - numcodecs = 0.11.0
   # install these from conda-forge so that dependent packages get included in the distributable
   - jsonschema = 4.18.0  # installs jsonschema-specifications
-  - pydantic[email] = 2.0.2  # installs email-validator
+  - pydantic[email] = 1.10.12  # installs email-validator
   - pip
   - pip:
       - chardet == 5.1.0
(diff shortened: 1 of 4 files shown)


CodyCBakerPhD:
IDK - how do you guys feel about just storing a different `.spec` file for each supported platform/architecture, having `npm` commands for regenerating those spec files, and adjusting the existing pyinstaller commands to just run on those `.spec` files?

Personally I think it makes the setup much more explicit

CodyCBakerPhD:
@rly @garrettmflynn

How do you feel about just storing a different `.spec` file for each supported platform/architecture, having `npm` commands for regenerating those spec files, and adjusting the existing pyinstaller commands to just run on those `.spec` files?

Personally I think it makes the setup much more explicit

garrettmflynn:
Based on our discussion today, I think this is an appropriate setup.

CodyCBakerPhD:
No longer needed

environments/linux-env-fix.yml
CodyCBakerPhD, line 1: Removing this unused file

Though the act of exact version pinning on the Python side is something we should consider

CodyCBakerPhD:
@rly @garrettmflynn Ready for review - this is still relevant though minor issue since the pip dependency resolver will force install of correct version for the environments on main

garrettmflynn (approved):
Yep looks good

rly (approved):

rly:
Looks good to me too

PR #363, opened 2023-09-13
Alternative lower pydantic pin

CodyCBakerPhD:
Might replace #361, just seeing if backpeddling the conda-forge jump fixes anything


CodyCBakerPhD:
Nope

PR #376, opened 2023-09-18
Allow direct updates to deep results when no nested forms exist

garrettmflynn:
This PR ensures that updates to the results of the JSONSchemaForm object propagate for deep fields, even when not nested in another form. 

This can be tested by creating a new pipeline and attempting to fill in the Source Data manually. Previously, this would not update correctly.


garrettmflynn:
An automated test was also added

CodyCBakerPhD:
@garrettmflynn Oh my, you are quite right - I've never tried that myself...

After following your suggestion for testing on this PR, by manually typing the path into the pop-up file explorer on Windows, the box disappeared

image

No obvious errors from any console

CodyCBakerPhD:
In all honesty however, I don't see this as a huge problem if this is the only place it is an issue. People will generally be navigating and selecting graphically or dragging and dropping

garrettmflynn:
@CodyCBakerPhD Hmm this doesn't happen on my system. Can you tell me your exact workflow?

garrettmflynn:
Oh weird, looks like a windows-only issue

garrettmflynn:
@CodyCBakerPhD Alright I fixed this issue—which actually uncovered a few others.

Now the solution for these direct update issues is solved in a much more elegant fashion

CodyCBakerPhD:
Ty for the tests!

CodyCBakerPhD (approved):

PR #377, opened 2023-09-18
Cosmetic updates from User Test 1

garrettmflynn:
This PR implements a set of cosmetic updates from User Test 1—specifically the inclusion of subtitles to indicate page purpose.

PR #374, opened 2023-09-18
Allow Non-Compliant JSON (e.g. NaN values)

garrettmflynn:
fix https://github.com/catalystneuro/neuroconv/issues/515

This PR will prepare us to accept `NaN` values passed back as metadata from the Ophys interfaces.

Issue #311, opened 2023-08-21
[Cody's Windows JS Error Collection]


CodyCBakerPhD:
I'll just start a little collection of screenshots of these since they do not occur within our standard loggers or dev consoles; the window itself isn't even copy/pastable

CodyCBakerPhD:
Running `npm start` in dev mode from the neurosift branch

Everything worked fine - *there was no problem with the app* - I'm just leaving this here as a record in case we ever see it again

Upon closing down the application (top-right X button on Windows)

![image](https://github.com/NeurodataWithoutBorders/nwb-guide/assets/51133164/29938eb4-087d-4861-a8ed-f134b68e9386)

CodyCBakerPhD:
Selecting 'OK' actually triggers follow-up window errors too, slightly different

![image](https://github.com/NeurodataWithoutBorders/nwb-guide/assets/51133164/2786cea8-df42-4821-a23c-04ad3b8b7a0d)

CodyCBakerPhD:
For provenance, the first time I saw something like this was in https://github.com/NeurodataWithoutBorders/nwb-guide/pull/283#issuecomment-1675979061

![image](https://github.com/NeurodataWithoutBorders/nwb-guide/assets/51133164/524e9911-8942-4f2c-9779-d46a65c8f121)

garrettmflynn:
These are errors thrown by the main Electron process. The Object ones occur when trying to send messages to the browser window that has been destroyed—an issue that I believe was fixed in a recent PR (which I'll link here momentarily) by keeping better track of the open state.

garrettmflynn:
I'm assuming you got these errors after merging? I'll have to do a bit more digging if that's the case to see what messages are being thrown after window closure.

CodyCBakerPhD:
Huh, new one today

This one showed up - not immediately - but rather when I first woke my computer up (I usually just put it to sleep over night)

I had closed the app earlier the night before

![image](https://github.com/NeurodataWithoutBorders/nwb-guide/assets/51133164/36f35377-653a-4afd-85ab-1f4e66a8055d)

garrettmflynn:
Hmmm this looks like the Python server didn't close correctly, threw a message when you opened your computer, and an attempt was made to send this to a non-existent BrowserWindow object.

[...]

PR #384, opened 2023-09-20
Add single-plane Bruker Interface

garrettmflynn:
This is a useful test for #393 and the first ophys interface we can test.


CodyCBakerPhD:
Oh, was the discussion/changes of the JSON handling only relevant for rendering of the metadata forms for ophys?

If so I really think we should add metadata control for ophys in a standalone PR and leave these interface introductions to be standalone additions with their own debugs since there's a lot we would want to change about the metadata forms independent of each format

CodyCBakerPhD:
Even with the changes, I was unable to proceed past the metadata page

```javascript
Uncaught (in promise) Error: Request to convert failed: None is not of type 'number'

Failed validating 'type' in schema['properties']['Ophys']['properties']['ImagingPlane']['items']['properties']['excitation_lambda']:
    {'description': 'Excitation wavelength in nm.', 'type': 'number'}

On instance['Ophys']['ImagingPlane'][0]['excitation_lambda']:
    None
    at run (utils.js:37:33)
    at async GuidedMetadataPage.runConversions (Page.js:167:28)
    at async GuidedFooter.onNext [as __onNext] (GuidedMetadata.js:43:27)
```

So I say let's save the metadata exposure for a follow-up and just use the internally generated default values inside the API functions for now, which means we probably don't need to merge #374 or use it as the base here

garrettmflynn:
Ah, I see what's going on. Since I'm passing the object back to the Flask server. using `JSON.stringify`, those NaN values are being converted to `null` (`None` in Python).

Your suggestion makes sense. It's similar to how we're not handling Ecephys metadata at this point as well.

garrettmflynn:
Alright I was able to run a conversion using the changes made in #393, essentially implementing what you were thinking might be useful in Neuroconv itself (as a layer above PyNWB / NWB Schema).

garrettmflynn:
Thought I did switch the base. Does it look different on your end?

CodyCBakerPhD:
> Thought I did switch the base. Does it look different on your end?

The base is as you indicate, but I wasn't sure if I should hit the 'update branch' button given this branch also started from a different base originally

garrettmflynn:
Oh sure. I'd done that manually to—but wouldn't hurt to confirm

CodyCBakerPhD:
When launching the app to test this, saw https://github.com/NeurodataWithoutBorders/nwb-guide/issues/311#issuecomment-1735995141

Though I suspect it may be residual from another PR such as #392 

Not originally detected because I used hot reloading on an already launched app; this only shows when I closed by testing #392 and reopened after switching the branch

CodyCBakerPhD:
This actually does propagate to an issue where the backend does not connect anymore

```javascript
TypeError: paths.subfolders.stubs is not iterable (cannot read property undefined)
    at Object.<anonymous> (D:\GitHub\nwb-guide\build\main\main.js:15954:37)
    at Module.<anonymous> (node:internal/modules/cjs/loader:1271:14)
    at Object..js (node:internal/modules/cjs/loader:1326:10)
    at Module.load (node:internal/modules/cjs/loader:1126:32)
    at node:internal/modules/cjs/loader:967:12
    at Function._load (node:electron/js2c/asar_bundle:2:13377)
    at loadApplicationPackage (D:\GitHub\nwb-guide\node_modules\electron\dist\resources\default_app.asar\main.js:121:16)
    at Object.<anonymous> (D:\GitHub\nwb-guide\node_modules\electron\dist\resources\default_app.asar\main.js:233:9)
    at Module.<anonymous> (node:internal/modules/cjs/loader:1271:14)
    at Object..js (node:internal/modules/cjs/loader:1326:10)
Killing all previous processes
```

CodyCBakerPhD:
Which is also an issue on `main` and so blocking testing of everything else

Would you like me to raise an issue for it?

garrettmflynn:
Just forced a push to `main`. It was from #377, I remember making the correct change, but must not have saved properly before pushing.

CodyCBakerPhD:
OK working well now

Just leaving as a note, the dual-channel example (NCCR62_2023_07_06_IntoTheVoid_t_series_Dual_color-000) fails unless you know the stream name for each optic channel, but that's hard to tell ATM

Main solution will be to expose the 'simple' mode of interface selection, which is actually to use the subconverters that handle this automatically

CodyCBakerPhD:
Can you add a source data story so we can keep track of how the stream name and other future options look?

garrettmflynn:
We may want to add a trigger to regenerate the Source Data stories whenever the guidedGlobalMetadata file changes, since all I have to do is manually run `python generateInterfaceSchema.py`. I thought of this since we missed doing this with Spike2 before.

What do you think?

CodyCBakerPhD:
> We may want to add a trigger to regenerate the Source Data stories whenever the guidedGlobalMetadata file changes, since all I have to do is manually run python generateInterfaceSchema.py. I thought of this since we missed doing this with Spike2 before.

It would be nice if not too much trouble - what kind of trigger were you thinking of?

Various CI around the ecosystem have a step called `assess-file-changes` and other dependent jobs after that to send logic back to the display to notify a specific file was changed as a reminder; auto running the generator, commiting, and pushing back to the repo is also possible but much more of a headache to setup

CodyCBakerPhD (approved):

garrettmflynn:
Yeah just that. Detect changes to that file, run the generation script if so, then push back to the repo.

If it's a hassle, there's no real need though. Manual is fine—we just have to remember to re-sync every so often.

CodyCBakerPhD:
> If it's a hassle, there's no real need though. Manual is fine—we just have to remember to re-sync every so often.

Pushing back would be a hassle

Setting up a reminder based on file changes is pretty easy, I can show you how to set it up tomorrow

PR #386, opened 2023-09-20
Fix error in updating project name, edit error msg

### LINKED ISSUE #385: Updating project name results in error
When I open an existing project and change the name of the project, a red box pops up that says "ReferenceError: hasEntry is not defined". Same if I create a new project and use the left sidebar to go back to "Project details" and change the name of the project.

Also, if the project name has not been updated, a green box pops up that says "No changes made to dataset name" that is too fast to read. I think this pop up is not necessary.

OS: Mac M1 using the latest commit on main

I know what to change and will submit a PR.

rly:
Fix #385.

- Returning to "Project details" page and:
	- not changing the name no longer results in green popup box that says "No changes made to dataset name"
	- changing the name to an allowed name no longer results in green popup box that says "Dataset name updated"
	- changing the name to one that is already taken no longer results in ReferenceError (it will now result in error that says "An existing project already exists with that name. Please choose a different name.")
- Changed "progress file" -> "project" in related error messages

### DIFF
--- a/src/renderer/src/progress/index.js
+++ b/src/renderer/src/progress/index.js
@@ -58,8 +58,6 @@ export const save = (page, overrides = {}) => {
 export const getEntries = () => {
     if (fs && !fs.existsSync(guidedProgressFilePath)) fs.mkdirSync(guidedProgressFilePath, { recursive: true }); //Check if progress folder exists. If not, create it.
     const progressFiles = fs ? fs.readdirSync(guidedProgressFilePath) : Object.keys(localStorage);
-    console.log(progressFiles)
-
     return progressFiles.filter((path) => path.slice(-5) === ".json");
 };
 
--- a/src/renderer/src/progress/index.js
+++ b/src/renderer/src/progress/index.js
@@ -14,7 +14,7 @@ import { merge } from "../stories/pages/utils.js";
 import { updateAppProgress, updateFile } from "./update.js";
 import { updateURLParams } from "../../utils/url.js";
 
-import * as operations from './operations'
+import * as operations from "./operations";
 
 export * from "./update";
 
@@ -124,7 +124,7 @@ export const remove = async (name) => {
         focusCancel: true,
     });
 
-    if (result.isConfirmed) return operations.remove(name)
+    if (result.isConfirmed) return operations.remove(name);
 
     return false;
 };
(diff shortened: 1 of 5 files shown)


garrettmflynn:
> changing the name to an allowed name no longer results in green popup box that says "Dataset name updated"
IMO, unlike when you first set a project's name or keep it the same, I'd want to be told that I've changed my project's name—in case that was not intended, or simply as a confirmation that my intended action was successful. 

Aside from reducing the complexity, is there a reason you wanted this removed?

garrettmflynn:
Otherwise this works great! Thanks for getting in there and fixing it

rly:
> > changing the name to an allowed name no longer results in green popup box that says "Dataset name updated"
> 
> IMO, unlike when you first set a project's name or keep it the same, I'd want to be told that I've changed my project's name—in case that was not intended, or simply as a confirmation that my intended action was successful.
> 
> Aside from reducing the complexity, is there a reason you wanted this removed?

IMO the confirmation is unnecessary and adds visual clutter. The button to proceed says "Save and Continue" so as a user, I would assume that if I changed the name, that it was saved. Also, the other pages, like the Subject Metadata Table do not provide popups that a value has changed.

CodyCBakerPhD:
Does it simply change the folder name, or does it create a new folder after renaming? I think we may have been concerned with data loss originally

CodyCBakerPhD:
BTW at risk of sounding repetitive, is there any way either of you can think of for adding a test of this behavior?

rly:
> Does it simply change the folder name, or does it create a new folder after renaming? I think we may have been concerned with data loss originally

If the app has file system access, it uses [fs.renameSync](https://nodejs.org/api/fs.html#fsrenamesyncoldpath-newpath) to rename the folder. Otherwise, it creates a new local storage entry, copies the contents of the old entry into it, and removes the old entry.

> BTW at risk of sounding repetitive, is there any way either of you can think of for adding a test of this behavior?

I'll look into this.

garrettmflynn:
@rly Just added a test. Confirmed on `main` that this would otherwise fail

garrettmflynn:
> IMO the confirmation is unnecessary and adds visual clutter. The button to proceed says "Save and Continue" so as a user, I would assume that if I changed the name, that it was saved. Also, the other pages, like the Subject Metadata Table do not provide popups that a value has changed.

This is a fair point. We should be striving for consistency here. While `name` is an essential identifier for the user when revisiting a pipeline, we aren't indicating when, for instance, a user changes a `subject_id` to something else.

If you prefer it gone, then I trust your judgment.

garrettmflynn:
Adding @CodyCBakerPhD as a reviewer because of the tests

CodyCBakerPhD (approved):

PR #389, opened 2023-09-21
Encode Global Settings File

### LINKED ISSUE #369: [Discussion] Encryption of DANDI API keys?
We are currently hiding the main page display of the API keys using the black dots convention, which is good because then someone can show or record their use of the guide without revealing their API key publicly

But I just noticed that we're dumping the DANDI API keys to the config.json file in the home directory for the GUIDE

As such, anyone could go into that file and they would display; or anyone could obtain these if they accidentally shared the config file

Should we be instead locking these keys away using encryption/password protection?

garrettmflynn:
This PR encrypts / decrypts strings in the global settings file.

fix #369


CodyCBakerPhD:
The values of the API keys in the `config.json` text file appear the same as on `main` for me 🤔

I even deleted the file and saved a fresh one through the app

garrettmflynn:
Whoops! Was just logging the encoded values for testing, never actually saved them...

CodyCBakerPhD:
One small conflict now

garrettmflynn:
Should be good to go!

CodyCBakerPhD:
@garrettmflynn The new config file contents look good, but I can't navigate away from the settings page in the app after resetting the values

The popup

image

blocks me even if I keep hitting 'save and continue'

garrettmflynn:
Good catch. Fixed

CodyCBakerPhD:
OK looks good now 👍

CodyCBakerPhD (approved):

PR #390, opened 2023-09-22
Update Path Expansion Page

### LINKED ISSUE #381: Update Path Expansion Page for Improved Usability
During user tests, it was difficult to explain the Locate Data (Path Expansion) page. To fix this issue, we propose the following changes to that page:

1. Have the description box expanded by default
2. Include one full example on the page, including (i) what the files look like in the system and (ii) what the resulting f-string looks like. Also provide a list of special tokens (e.g. `subject_id`) to use in the f-string
3. When selecting the base directory, populate initially with a deep file path to provide a template for users to start replacing things
4. Live update the results of the f-string search when the user updates the input. This will involve (i) parsing all the entities in the base directory, then (ii) filtering them based on matches in real-time.

garrettmflynn:
fix #381 

- A complete example is contained in the information toggle previously defined with an external link
- The information toggle is automatically open when the user firsts visits the page (i.e. if Yes / No has not been selected yet)
- A box shows beneath each `format_string_path` input with the current source file matches, while the full metadata is printed to the console

In attempting to implement all the suggestions, I decided it wouldn't be worthwhile to approach (3) and parts of (4) without further discussion.
- Auto-populating the `format_string_path` input is too unpredictable what the user might want (e.g. for the tutorial pipeline, I don't have enough knowledge to select a Phy folder or a SpikeGLX-related file if both exist at the lowest level of the `base_directory`)
- Listing out all the entities in the base directory will almost always be a lot of files / folders—too many to show nicely in a naive way.

Are there alternative approaches that y'all would suggest or something that I haven't implemented here that you'd like to see still?


garrettmflynn:
Based on our discussion today, we can wait for an implementation of (3) based on #397. This PR should be good for review.

CodyCBakerPhD (approved):

CodyCBakerPhD:
I like. Will get @bendichter's feedback on this during that one on one alpha test

PR #391, opened 2023-09-25
Audit NPM Packages

garrettmflynn:
This PR removes as many security vulnerabilities possible based on the `npm audit` command. 

From 32 vulnerabilities (10 moderate, 16 high, 6 critical) using `npm ci` to 18 vulnerabilities (1 low, 12 high, 5 critical)

To achieve a reduction of our vulnerabilities, Electron was upgraded to the latest version and our required node version was bumped (or really explicitly set) to v18, where the `fetch` API is native (removing the need for `axios`.

Remaining vulnerabilities are largely a result of our reliance on `fomantic-ui` from the original SODA application, which there are no ready fixes for. 

Should be possible to remove this dependency by manually adding styles that are similar to those in the `fomantic-ui` library. Though this may take some time. Would we want to do this? Chromatic should be very useful to detect changes.


CodyCBakerPhD:
Had to accept new baselines, but everything looks equivalent to me

Merge if ready

CodyCBakerPhD (approved):

Issue #176, opened 2023-05-15
[Ophys Support] Ophys metadata


CodyCBakerPhD:
Ecephys is the initial early focus of the NWB GUIDE, but ophys will soon follow

First step will be to adopt and polish the NeuroConv ophys metadata and implementation here, similar to what's going on with the ecephys metadata in #161

garrettmflynn:
Completed with #505

PR #393, opened 2023-09-26
Python-side fix for expecting NaN in JSON

garrettmflynn:
This PR relies on Python instead of JavaScript to handle expected NaN values in the JSON metadata (unlike #374). 

Instead of allowing non-compliant JSON on the browser (which doesn't work when passing the object back to Python), this approach replaces `None` values where numbers are expected with `NaN` on the Python side, while converting `NaN` values to `None` before passing to JavaScript.

Since this is primarily testing Ophys interfaces (#177), the `Ophys` metadata field is ignored in this PR so we can work on those interactions in a follow-up (related to #176).

### DIFF
--- a/pyflask/manageNeuroconv/manage_neuroconv.py
+++ b/pyflask/manageNeuroconv/manage_neuroconv.py
@@ -2,6 +2,7 @@
 import os
 import json
 import math
+import copy
 from datetime import datetime
 from typing import Dict, Optional  # , List, Union # TODO: figure way to add these back in without importing other class
 from shutil import rmtree, copytree
@@ -57,9 +58,6 @@ def resolve_references(schema, root_schema=None):
 
 
 def replace_none_with_nan(json_object, json_schema):
-    import math
-    import copy
-
     """
     Recursively search a JSON object and replace None values with NaN where appropriate.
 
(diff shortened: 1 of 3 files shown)


pyflask/manageNeuroconv/manage_neuroconv.py
```diff
@@ -12,6 +13,83 @@
...
+    """
+    Recursively resolve references in a JSON schema based on the root schema.
+
+    Args:
+        schema (dict): The JSON schema to resolve.
+        root_schema (dict): The root JSON schema.
+
+    Returns:
+        dict: The resolved JSON schema.
+    """
```
CodyCBakerPhD, line 41:
```suggestion
    """
    Recursively resolve references in a JSON schema based on the root schema.

    Args:
        schema (dict): The JSON schema to resolve.
        root_schema (dict): The root JSON schema.

    Returns:
        dict: The resolved JSON schema.
    """
    from jsonschema import RefResolver
```

docstring won't get set unless it's the first string after the `def`

pyflask/manageNeuroconv/manage_neuroconv.py
```diff
@@ -1,6 +1,7 @@
 """Collection of utility functions used by the NeuroConv Flask API."""
 import os
 import json
+import math
```
CodyCBakerPhD, line 4:
```suggestion
import math
import copy
```

CodyCBakerPhD (approved):

CodyCBakerPhD:
Much less hacky than `eval`; nice solution

CodyCBakerPhD (approved):

PR #398, opened 2023-09-28
Add Hdf5ImagingInterface

garrettmflynn:
This PR adds the Hdf5ImagingInterface and was successful on the test data. 

This interface has a lot of optional metadata that we may choose to hide:
Screenshot 2023-09-28 at 1 13 32 PM


CodyCBakerPhD:
This is one of those super generic ones, like having a BinaryRecordingInterface - it doesn't correspond to actual output from any microscope AFAIK, always some kind of post-processing of it

Focusing on non generic proprietary formats for the time being, so I saw we close this one for now then come back at a future point

PR #399, opened 2023-09-28
Add MicroManagerTiffImagingInterface

garrettmflynn:
This PR adds the MicroManagerTiffImagingInterface. The first pass through this interface, however, reveals an issue where **JavaScript does not maintain float values with trailing zeros** so the value of `imaging_rate` in JSON is `20.0` (float), which gets converted to `20` in JS (number) and passed back to Python as `20` (int).

Do we need to correct for int-like floats like we do for expected NaN values?

## Error
```

[2023-09-28 13:17:03,870] ERROR in app: Exception on /neuroconv/convert [POST]
Traceback (most recent call last):
  File "/Users/garrettflynn/Documents/Github/nwb-guide/pyflask/apis/neuroconv.py", line 82, in post
    return convert_to_nwb(neuroconv_api.payload)
  File "/Users/garrettflynn/Documents/Github/nwb-guide/pyflask/manageNeuroconv/manage_neuroconv.py", line 383, in convert_to_nwb
    converter.run_conversion(
  File "/opt/anaconda3/envs/nwb-guide/lib/python3.9/site-packages/neuroconv/nwbconverter.py", line 160, in run_conversion
    self.add_to_nwbfile(nwbfile_out, metadata, conversion_options)
  File "/opt/anaconda3/envs/nwb-guide/lib/python3.9/site-packages/neuroconv/nwbconverter.py", line 114, in add_to_nwbfile
    data_interface.add_to_nwbfile(
  File "/opt/anaconda3/envs/nwb-guide/lib/python3.9/site-packages/neuroconv/datainterfaces/ophys/baseimagingextractorinterface.py", line 118, in add_to_nwbfile
    add_imaging(
  File "/opt/anaconda3/envs/nwb-guide/lib/python3.9/site-packages/neuroconv/tools/roiextractors/roiextractors.py", line 492, in add_imaging
    add_photon_series(
  File "/opt/anaconda3/envs/nwb-guide/lib/python3.9/site-packages/neuroconv/tools/roiextractors/roiextractors.py", line 364, in add_photon_series
    nwbfile = add_imaging_plane(nwbfile=nwbfile, metadata=metadata_copy, imaging_plane_index=photon_series_index)
  File "/opt/anaconda3/envs/nwb-guide/lib/python3.9/site-packages/neuroconv/tools/roiextractors/roiextractors.py", line 251, in add_imaging_plane
    imaging_plane = _create_imaging_plane_from_metadata(
[...]

PR #405, opened 2023-09-28
Coerce All Numbers to Floats to Maintain Consistency Between JS, JSON Schema, and Python

garrettmflynn:
fix #399 and #400

### DIFF
--- a/pyflask/manageNeuroconv/manage_neuroconv.py
+++ b/pyflask/manageNeuroconv/manage_neuroconv.py
@@ -69,22 +69,24 @@ def replace_none_with_nan(json_object, json_schema):
         dict: The modified JSON object with None values replaced by NaN.
     """
 
-    def replace_none_recursive(obj, schema):
+    def coerce_schema_compliance_recursive(obj, schema):
         if isinstance(obj, dict):
             for key, value in obj.items():
                 if key in schema.get("properties", {}):
                     prop_schema = schema["properties"][key]
                     if prop_schema.get("type") == "number" and value is None:
-                        obj[key] = math.nan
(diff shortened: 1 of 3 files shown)


pyflask/manageNeuroconv/manage_neuroconv.py
```diff
@@ -69,22 +69,24 @@ def replace_none_with_nan(json_object, json_schema):
...
+    def coerce_schema_compliance_recursive(obj, schema):
         if isinstance(obj, dict):
             for key, value in obj.items():
                 if key in schema.get("properties", {}):
                     prop_schema = schema["properties"][key]
                     if prop_schema.get("type") == "number" and value is None:
-                        obj[key] = math.nan
+                        obj[key] = math.nan # Turn None into NaN if a number is expected (JavaScript JSON.stringify turns NaN into None)
+                    elif prop_schema.get("type") == "number" and isinstance(value, int):
+                        obj[key] = float(value) # Turn integer into float if an integer is expected (JavaScript / JSON Schema coerces floats with trailing zeros to integers)
```
CodyCBakerPhD: Should the comment be 'if a float is expected'?

Also is it possible to auto-detect from the presence of a `.` in the stringified number whether it ought to be float or int?

garrettmflynn: Yes my bad. I'd written this before I looked up that `number` is equivalent to `float` in JSON Schema.

Theoretically yes—but that'd require us to roll our own JSON parser in addition to storing this information together with the parsed value (as JS has no way to differentiate `20` and `20.0`) until it's re-stringified for storage / use by Python.

CodyCBakerPhD (approved):

Issue #373, opened 2023-09-18
[Post-v1.0] Custom data interfaces as plugins


bendichter:
I wanted to create a ticket to discuss plugins. This is not for the MVP, since there are several features that need to take higher priority, but I wanted to start discussing how it would be implemented. The idea would be that a developer could define a custom interface, and in the GUIDE UI, a user could find that interface and add it to their app. This could be a simple as copying/pasting a GitHub repo URL.

This would potentially provide two big features:

1. Provide GUIDE functionality for users that have custom code in their conversion pipelines. We now have 35 different labs that we have built pipelines for and every single one has required at least a little custom code, making it impossible for them to use GUIDE to do their entire conversion. It can in most cases do the most involved parts, so GUIDE is still valuable for these groups, but it would be better if we could easily add the custom interfaces to GUIDE so that they could use these pipelines without needing to interact with Python at all. We could create a screen where a user can input the URL of a repo. Then we could pip install that repo and have some way to indicate where the interfaces are (maybe as an entry point?). Then the metadata forms might be tricky. We would need to either 1. automatically generate all of the metadata forms from the json schema, 2. manually define these forms within the plugin repo, 3. not allow any altering of metadata, or 4. only allow altering of metadata through a precisely structured yaml file.

2. Reduce the size of the initial download and duration of initial installation by moving some less-commonly-used interfaces to plugins.
The installation package is currently very large and takes a long time to install. Something that might help would be to remove some of the larger and/or less commonly used interfaces from the default package, and allow users to install them as needed. This would presumably speed up the initial installation, and would reduce the size of the application on the client's computer.

This feature could start as a simple unchecked GitHub repo. Then we could add features for automatically testing the interface to ensure it uses the proper classes and methods, etc., and finally we could create a registry of repos so they are easy to explore and discover within the GUIDE app itself.

CodyCBakerPhD:
This has been on our radar in the background

> We would need to either 1. automatically generate all of the metadata forms from the json schema, 2. manually define these forms within the plugin repo, 3. not allow any altering of metadata, or 4. only allow altering of metadata through a precisely structured yaml file.

I think this is exactly the workflow we could use

1.1: try to automatically generate a form based on JSON schema using default components. If it fails for certain reasons (unknown/unsupported components for a field) then proceed to

1.2: search for, or ahead of (1.1), import custom components or pages in some structured way

[...]

PR #407, opened 2023-09-28
Add SimaSegmentationInterface

garrettmflynn:
This PR adds the SimaSegmentationInterface. However, there are a few blockers:
1. I didn't notice proper test files—or at least the ones to use were unclear
2. I need to install `sima` and `dill` to use this interface, which we can't currently do for users in production (#300, https://github.com/NeurodataWithoutBorders/nwb-guide/issues/373#issuecomment-1723695159).


CodyCBakerPhD:
You are correct on this one - I had not been aware of this format previously, it does not seem to have tests anywhere, and I have no idea what project it was historically from

Letting @pauladkisson know about this too

For now, yeah I'd say let's just close this one for time being until we figure this one out

pauladkisson:
Thanks for the heads up!  I'll take a look at SIMA when I get a chance to see what we should do with it.

PR #409, opened 2023-09-29
Try new docs theme

CodyCBakerPhD:
Hoping this is all it takes to use the more modern docs theme


CodyCBakerPhD:
@garrettmflynn could you try stretching the window containing the table on this PR? The new theme should allow it to look better

CodyCBakerPhD:
Both vertically and horizontally might be worth trying, this new theme leaves a lot of free space

garrettmflynn:
Width restriction is built into the theme—but I did add about 20vh more height to the table: https://nwb-guide.readthedocs.io/en/try_new_docs_theme/format_support.html (may still be building)

Setting width anything other than `100%` results in a smaller table OR hidden overflow that cannot be scrolled.

CodyCBakerPhD:
Dang 🤔 I guess we just have to shrink the table itself a little bit then? https://github.com/catalystneuro/format-support-table/issues/9

Even zooming out on the new theme page doesn't quite fit the whole thing vertically

Not a huge deal at the end of the day, just feels like it ought to be fixable

garrettmflynn (approved):
Looks really good!

PR #412, opened 2023-09-29
Add Axona Recording Interface

garrettmflynn:
This PR adds the AxonaRecordingInterface and AxonaLFPDataInterface, both of which were successful on the test data.


CodyCBakerPhD:
@garrettmflynn Can you split the AxonaLFP one into a separate PR? The recording looks ok but there's some issues with the LFP output; it might not be an interface we want to allow for a while

CodyCBakerPhD:
@garrettmflynn Ping to split off the LFP interface to a separate PR; if this one is just the basic RecordingInterface then we can get this in as is

CodyCBakerPhD (approved):

PR #416, opened 2023-10-02
Fix @electron/remote

garrettmflynn:
This PR fixes the issue we had resolving `@electron/remote` in production because it was contained in `devDependencies`, which are not carried over.


CodyCBakerPhD:
Worked on local build

CodyCBakerPhD (approved):

PR #417, opened 2023-10-02
Add non-dev install instructions to README

CodyCBakerPhD:
From first alpha test, adding some install instructions when not going the dev install route


garrettmflynn (approved):
Looks good! Just fixed a typo

PR #420, opened 2023-10-03
Add electron as a main dependency

alejoe91:
To fix blank screen on Ubuntu


CodyCBakerPhD (approved):

CodyCBakerPhD:
The problem here was the screen would go blank on linux after 1 second of loading on the main page

PR #423, opened 2023-10-06
Move ubuntu installation note

CodyCBakerPhD:
I suspect this may have caused a small delay in reading the installation instructions

Moving further below since they would only encounter this after attempting the main way and then coming back looking for info


garrettmflynn (approved):
Ah perfect. Good to have this top-level

PR #430, opened 2023-10-07
[Bug] Use dandi config in all inspection

CodyCBakerPhD:
This could also be exposed to the payload as a future feature if people want to design their own configs, but that is extremely rare

### DIFF
--- a/pyflask/manageNeuroconv/manage_neuroconv.py
+++ b/pyflask/manageNeuroconv/manage_neuroconv.py
@@ -17,7 +17,7 @@ announcer = MessageAnnouncer()
 def replace_nan_with_none(data):
     if isinstance(data, dict):
         # If it's a dictionary, iterate over its items and replace NaN values with None
-        return {key: replace_nan_with_none(value) for key, value in data.items()}
+        return {key: replace_nan_with_none(value) for key, value in data.items()}x
     elif isinstance(data, list):
         # If it's a list, iterate over its elements and replace NaN values with None
         return [replace_nan_with_none(item) for item in data]
@@ -536,6 +536,7 @@ def inspect_nwb_file(payload):
                         "check_description",
                         "check_data_orientation",
                     ],  # TODO: remove when metadata control is exposed
+                    config="dandi",
                     **payload,
                 )
             ),
@@ -556,6 +557,7 @@ def inspect_nwb_file(payload):
                         "check_description",
                         "check_data_orientation",
                     ],  # TODO: remove when metadata control is exposed
+                    config="dandi",
                     **payload,
                 )
             ),
@@ -564,7 +566,7 @@ def inspect_nwb_file(payload):
     )
 
 
-def inspect_nwb_folder(payload):
+def _inspect_nwb_folder(payload):
     from nwbinspector import inspect_all
     from nwbinspector.nwbinspector import InspectorOutputJSONEncoder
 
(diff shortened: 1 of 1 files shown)


pyflask/manageNeuroconv/manage_neuroconv.py
```diff
@@ -564,7 +566,7 @@ def inspect_nwb_file(payload):
     )
 
 
-def inspect_nwb_folder(payload):
+def _inspect_nwb_folder(payload):
```
garrettmflynn, line 569: Why are you adding the `_` before this function name? While it's used in another function here, it's also exported and used in `neuroconv.py`

pyflask/manageNeuroconv/manage_neuroconv.py
```diff
@@ -17,7 +17,7 @@
 def replace_nan_with_none(data):
     if isinstance(data, dict):
         # If it's a dictionary, iterate over its items and replace NaN values with None
-        return {key: replace_nan_with_none(value) for key, value in data.items()}
+        return {key: replace_nan_with_none(value) for key, value in data.items()}x
```
garrettmflynn, line 20: Looks like an extra character was added here

CodyCBakerPhD, line 20:
```suggestion
        return {key: replace_nan_with_none(value) for key, value in data.items()}
```

pyflask/manageNeuroconv/manage_neuroconv.py
CodyCBakerPhD: Python convention in both cases; to indicate they are marked as 'private' functions not necessarily intended to be used directly by the API endpoints or anything outside the project

Especially for 'inspect_nwb_folder' to avoid any confusion with public nwbinspector methods, which follow a similar name pattern

garrettmflynn: If that's the case, would you like me to refactor the `neuroconv/inspect` route so it isn't using this directly? Because at present, this isn't _really_ private. Or is the intention (rather than actual use) sufficient?

CodyCBakerPhD: It's not a requirement, or a big deal - I just noticed it when I was poking around for this PR and changed it by convention while I was there

CodyCBakerPhD: Though I did not notice it is also used in another endpoint; which means it's not effectively private

Restored that now; keeping the aggregation one private though

pyflask/manageNeuroconv/manage_neuroconv.py
```diff
@@ -528,40 +528,18 @@ def inspect_nwb_file(payload):
...
+                "check_description",
+                "check_data_orientation",
+            ],  # TODO: remove when metadata control is exposed
+            config="dandi",
+            **payload,
         )
     )
 
-
-def inspect_nwb_file(payload):
```
CodyCBakerPhD, line 547: This functions appeared to have been duplicated

pyflask/manageNeuroconv/manage_neuroconv.py
garrettmflynn: Perfect thank you!

garrettmflynn (approved):
This looks good to me now!

PR #434, opened 2023-10-07
Add admin note

CodyCBakerPhD:
Windows users should be made aware of this early

@garrettmflynn How hard do you think it might be to have a specially tailored notification to inform them specifically of this when such an event happens?

Also, is there anything we can do on the electron side to perhaps make it run that way from the get-go (i.e., admin mode is default when double-clicking on it)

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,8 @@ Begin by downloading the latest release from the [GitHub release page](https://g
 
 Run the `setup.exe` file and follow all instructions.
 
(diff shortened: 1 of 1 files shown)


garrettmflynn:
Yes there's an option in `electron-builder` to force opening in Admin mode: https://www.electron.build/configuration/win.html#WindowsConfiguration-requestedExecutionLevel.

Do you want me to quickly implement this instead of merging this documentation update?

CodyCBakerPhD:
> Do you want me to quickly implement this instead of merging this documentation update?

Oh, sure if it's that easy!

PR #438, opened 2023-10-11
User Test #3 (Alessio) Updates

### LINKED ISSUE #429: [Bug] Staging key should not show up as required (or red banner) if not specified in settings
It's becoming obvious that the majority of people will not have their separate staging accounts to try DANDI upload outside of the main archive...

For alpha tests we are happy to provide temporary creds to our own (which are then reset for security) just to debug issues, but I suppose for the tutorial itself we might want to emphasize something about the difference between servers

garrettmflynn:
Includes several other cosmetic fixes that were relevant to proper rendering / navigation of the pages and how we communicate these things to the user.

fix #429 

- ”guided mode” url did not go to the correct location
- Tried updating the base folder in the Locate Data page. Dynamic prediction of path expander did not update correctly w.r.t. Changes to base directory
- “2 errors on this table” thrown twice when Age is wrong
- No feedback on hitting save button on metadata page
- Neurosift unable to connect to stub files because the preview folder doesn’t appear to be created at the home level
- Tried to click the top ‘-’ buttons where one is colored green and other is white (counter indicator through steps of subworkflow)
- Unsure what cleanup button meant on DANDI upload page
- Conversions page says “Continue Converting” instead of “Edit pipeline”
- The severity of warnings vs. errors is never explicitly addressed.
- Search bar is not sticky to the top of the modal—so you can’t search if you’re deep in the list
- Bottom two buttons on first page had terrible overlap with the three panels behind
- Not clear what the top-level buttons do on the Tutorial page
- It’s unclear that you don’t need the Main API key at the same time as the staging API (for the tutorial)
- cannot see the whole error for generating dataset because it is clipped


CodyCBakerPhD:
Clicking 'Generate Data Structure' on this page

image

leads to

```javascript
GuidedSubjects.js:42 Uncaught (in promise) TypeError: Cannot convert undefined or null to object
    at GuidedSubjectsPage.beforeTransition (GuidedSubjects.js:42:35)
    at Main.to (Page.js:64:14)
    at GuidedFooter.to (GuidedFooter.js:51:54)
    at GuidedFooter.onNext [as __onNext] (GuidedFooter.js:32:29)
    at EventPart.handleEvent (lit-html.ts:2018:29)
```

CodyCBakerPhD:
Also, I'm still able to insert some initial values to the age field, and if they are invalid, then I remove them, they stay invalid

image

garrettmflynn:
Fixed those validation issues!

CodyCBakerPhD:
> Tried to click the top ‘-’ buttons where one is colored green and other is white (counter indicator through steps of subworkflow)

Are these supposed to be clickable now?

garrettmflynn:
IMO the users should use the navigation sidebar to handle page jumps. So I tried to make clearer that these are _not_ clickable by removing the hover pointer (i.e. the cursor no longer reacts when it is placed over the '-').

Do you think we should allow users to click these other elements ('-', section flow headers, etc.)?

CodyCBakerPhD:
> It’s unclear that you don’t need the Main API key at the same time as the staging API (for the tutorial)

Looks like this is only being countered via documentation, but I'm not sure the way it is now completely resolves all confusion about which API key is needed

But it's a lower priority thing so maybe it can go to the backburner



> Do you think we should allow users to click these other elements ('-', section flow headers, etc.)?

I do - people had the intuition that it was clickable even without the hover, as their eyes were more immediately drawn to that than to the side bar

CodyCBakerPhD:
OK went through everything and this is definitely a big improvement

The only way to test some of these things (that were specific to Alessio) would be to do a follow up with him

If you want to move clickability of pill boxes to a separate PR that would be fine by me, just let me know

garrettmflynn:
Alright, the capsules are now clickable. 

> Looks like this is only being countered via documentation, but I'm not sure the way it is now completely resolves all confusion about which API key is needed

Do you have any suggestions for a better fix? I can implement something else quickly—otherwise I think it's fine to put this into a separate issue.

CodyCBakerPhD:
> Do you have any suggestions for a better fix? I can implement something else quickly—otherwise I think it's fine to put this into a separate issue.

Alas I do not, otherwise I'd mention it  🤔 Fine to raise as a separate issue if you can make that

CodyCBakerPhD (approved):

CodyCBakerPhD:
Can confirm capsules are reactive and... slightly pleasurable to be able to click after all this time. Especially on that first batch of 5 pages, I can see why people intuited that as a form of page navigation

PR #439, opened 2023-10-11
Require Admin Privileges on Windows

garrettmflynn:
This PR ensures that the GUIDE is run in Admin Mode on Windows distributions, which is required for operating on the user's filesystem.


CodyCBakerPhD:
Works like a charm!

image

And was able to interact with local file system without error (and without manually launching as admin from the get-go)

CodyCBakerPhD (approved):

PR #444, opened 2023-10-16
Add Axona LFP Interface

garrettmflynn:
Split off from #412 to add the AxonaLFPDataInterface, which was successful on the test data.


CodyCBakerPhD:
Decided not to support for time being given how add-hoc the format on NeuroConv is

Added easy to click link to format support page in case anyone ever wants to request it in the future

PR #445, opened 2023-10-16
Add Neuralynx Recording

garrettmflynn:
This PR adds NeuralynxRecordingInterface, which was successful on the test data for Cheetah_v5.6.3.


CodyCBakerPhD:
Bizarre issue for me; when I follow the Folder selector that appears on the source data page, and select my folder of neuralynx data, I get

```javascript
Uncaught (in promise) Error: <h5 style="margin-bottom: 0;">Incorrect filesystem object</h5> <small>Please provide a <b>directory</b> instead.</small>
    at #onThrow (FileSystemSelector.js:104:15)
    at #checkType (FileSystemSelector.js:131:26)
    at Array.forEach (<anonymous>)
    at #handleFiles (FileSystemSelector.js:137:53)
    at FilesystemSelector.selectFormat (FileSystemSelector.js:165:30)
```

garrettmflynn:
Oh right, I was using a heuristic to detect files vs. directories that might be passed using drag-and-drop (i.e. not controlled natively)—but that was just checking whether there are `.` characters in the path.

Added a direct check when Electron (Node) libraries are accessible.

CodyCBakerPhD:
Note to self: electrode table is a bit messy, need to clean up all those excessive/uninformative columns

Otherwise looks good!

CodyCBakerPhD (approved):

PR #446, opened 2023-10-16
Add AbfInterface

garrettmflynn:
This PR adds the AbfInterface to the GUIDE, which passes on the test data. 

> **Note:** For the electron file selector, it seems that you aren't allowed to multi-select symlinks (at least symlinked files)—so you have to drag-and-drop the test data into the FileSystemSelector component if you'd like to test more than one


CodyCBakerPhD:
The source data page here needs work and even from the get-go sends us down the fairly time consuming road of figuring out how to handle the icephys metadata, so let's table that for the time being

garrettmflynn:
Would it be possible to simply exclude the `icephys_metadata` field and allow this interface—or does this still leave too much complexity / too little control for the user?

CodyCBakerPhD:
As indicated in the conversion gallery https://neuroconv.readthedocs.io/en/main/conversion_examples_gallery/recording/abf.html

this metadata is quite important

Icephys in general is much less about the time series data itself and more about all the annotations and structure that goes into describing it

PR #447, opened 2023-10-16
Fix Config Loading (NWB Inspector)

garrettmflynn:
This PR ensures that the DANDI configuration is directly passed to the NWB Inspector, instead of the related "dandi" key. 

This fixes the following error when attempting to run the Inspector: 
Screenshot 2023-10-16 at 12 44 52 PM


CodyCBakerPhD:
@garrettmflynn Oh geez, sorry about that. 100% my fault for forgetting the API for those downstream ones don't take the simple string

CodyCBakerPhD (approved):

PR #450, opened 2023-10-16
Add DeepLabCut Interface

garrettmflynn:
This PR adds the DeepLabCutInterface to the GUIDE, which passes on the test data.


CodyCBakerPhD:
Looks OK - there is an improvement similar to stream name enum that can be made eventually to help identify available subjects

@garrettmflynn However, chromatic currently complaining about something in the books: https://github.com/NeurodataWithoutBorders/nwb-guide/actions/runs/6551463289/job/17792680103

Might also be from one of the previous merges (I haven't been running every time to save space on captures)

garrettmflynn:
Just updated the stories. I'm assuming that's why this had an issue.

If there's ever a merge conflict with the Source Data stories file, I just accept `main` and re-generate before merging the PR itself.

PR #451, opened 2023-10-16
Add SLEAP Interface

garrettmflynn:
This PR adds SLEAPInterface to the GUIDE, which passes on the test data.


CodyCBakerPhD:
Looks good 👍 The frames per second source argument might be OK to remove in the future but will need to confirm

CodyCBakerPhD (approved):

PR #455, opened 2023-10-16
Add a loose requirement mode

### LINKED ISSUE #425: [Feature] Indicate future requirements/suggestions earlier on global pages
Many alpha testers voiced that they would prefer to be warned on the global metadata pages what will eventually be flagged as recommended by the NWB Inspector

Ideally, people would fill out as much information as possible under assumption that everything on that page is valuable - but this nonetheless caused confusion among a number of testers

garrettmflynn:
fix #425

### DIFF
--- a/src/renderer/src/stories/JSONSchemaForm.js
+++ b/src/renderer/src/stories/JSONSchemaForm.js
@@ -116,7 +116,7 @@ hr {
   :host([requirementmode="loose"]) .required label:after {
     color: gray;
   }
-  
+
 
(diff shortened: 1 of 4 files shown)


CodyCBakerPhD:
Similar to the explanation of red vs. yellow boxes, do you think we should have any part on the first pages of guided mode that explains what the gray asterisk means?

garrettmflynn:
Good idea. Just added that explanation

CodyCBakerPhD (approved):

PR #456, opened 2023-10-17
Add tooltips using Tippy.js

### LINKED ISSUE #427: [Feature] Make info icons display content on click as well as hover
Hover is the current default behavior for info icons, but a lot of people assumed clicking on it would activate, and didn't keep the cursor on the element long enough to see the trigger

garrettmflynn:
fix #427

Rather than opening on click in addition to hover, I've just used Tippy.js—which is much more responsive. Let me know if this works for you!


CodyCBakerPhD:
The hover here does indeed pop up super quickly, and is nicely formatted. We can give this a go and see what users think!

CodyCBakerPhD (approved):

PR #460, opened 2023-10-17
Add Suite2pSegmentationInterface

garrettmflynn:
This PR adds the Suite2pSegmentationInterface to the allowed interfaces. Though this currently fails on the test data with the following error when attempting to create a stub preview:

Screenshot 2023-10-17 at 10 05 20 AM

@CodyCBakerPhD I tried updating my test data—but this didn't change anything. Looks like this has to do with something in ROIExtractors?


weiglszonja:
hmm, weird. and this is something that only occurred with the suite2p and not with the other segmentation interfaces in stud mode?

garrettmflynn:
Yeah you got it. The rest were tested within the same hour, aside from Caiman (which was approved a couple weeks ago)

weiglszonja:
I figured it out, before https://github.com/catalystneuro/roiextractors/pull/242 the extractor didn't have an _image_masks attribute rather override the base `get_roi_image_masks()` to return the image masks. When we're using `FrameSliceSegmentationExtractor` for stubbing however, in retrieving the image masks we fall back to the base `get_roi_image_masks()` which expects the `_image_masks` attribute.

https://github.com/catalystneuro/roiextractors/blob/88372b5cc44fca1812755ef414d90af45163a331/src/roiextractors/segmentationextractor.py#L133

garrettmflynn:
Just tested. This works now!

CodyCBakerPhD:
Can you remove the `combined` and `plane_no` arguments? They've been hard deprecated for a while, just looks like we forgot to remove them https://github.com/catalystneuro/neuroconv/blob/main/src/neuroconv/datainterfaces/ophys/suite2p/suite2pdatainterface.py#L65-L66

garrettmflynn:
Updated!

CodyCBakerPhD:
How interesting - we've been seeing some daily failures on the PyFlask builds but they haven't shown up on a PR trigger until now

Could you try looking into the source issue and fix on another PR? I'll do a bit of investigation on my own too

PR #464, opened 2023-10-17
Add MiniscopeBehaviorInterface

garrettmflynn:
This PR adds the MiniscopeBehaviorInterface, which will run a full conversion on the test data.


CodyCBakerPhD:
While the specific error here is due to selecting the incorrect folder path, I had thought we fixed the clipping of the message in Alessios review PR

image

Doesn't need to be fixed here, just wanted to let you know it still happens

CodyCBakerPhD:
Also, food for thought - we might want to have a different endpoint for running the inspector on guided mode vs. standalone, and suppress the check about relative video paths in guided mode since user can't change anything about it

Otherwise interface works fine

garrettmflynn:
> While the specific error here is due to selecting the incorrect folder path, I had thought we fixed the clipping of the message in Alessios review PR

Thanks for the heads up!  I noticed this as well and pushed a quick fix to `main`.
Screenshot 2023-10-17 at 4 06 36 PM

This was a persistent issue here because the long file path wasn't wrapping.

PR #470, opened 2023-10-23
Update API Settings 

### LINKED ISSUE #469: Make staging key in settings not strictly required
I'd suggest separating the DANDI keys into different tabs on the settings, and also having only the main key be required (i.e., show up as red status when missing)

The staging, for the majority of actual use cases beyond the alpha testing and optional tutorial completion, will mainly be for developers (also requires different permissions on DANDI side, apparently) - so I'd either not show a status requirement or if you need to, show it as yellow instead of red

garrettmflynn:
fix #469

Kept in the same structure for easier validation of their existence—though can still move the staging key around if you'd like. What header should we put that under?

### DIFF
--- a/schemas/json/dandi/global.json
+++ b/schemas/json/dandi/global.json
@@ -6,12 +6,12 @@
                 "main_api_key": {
                     "type": "string",
                     "format": "password",
-                    "description": "From the <a href='https://dandiarchive.org' target='_blank'>main archive</a>"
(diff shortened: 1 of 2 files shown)


CodyCBakerPhD:
This already looks better

> What header should we put that under?

I would have said 'Developer' perhaps, but not sure right now.

Let's see how this plays out for the next round of tests (we'll provide the API keys and dandi sets though of course)

PR #467, opened 2023-10-18
Allow selection of converters with SpikeGLXConverter Support

garrettmflynn:
fix #174

This PR adds sub-converters to the options available to the user. 

There's currently a blocking error when attempting to instantiate the custom converter on the GUIDE backend:
```
index.js:33 [main-process]: [2023-10-18 11:40:09,245] ERROR in app: Exception on /neuroconv/metadata [POST]
Traceback (most recent call last):
  File "/opt/anaconda3/envs/nwb-guide/lib/python3.9/site-packages/flask/app.py", line 1484, in full_dispatch_request
    rv = self.dispatch_request()
  File "/opt/anaconda3/envs/nwb-guide/lib/python3.9/site-packages/flask/app.py", line 1469, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)
  File "/opt/anaconda3/envs/nwb-guide/lib/python3.9/site-packages/flask_restx/api.py", line 404, in wrapper
    resp = resource(*args, **kwargs)
  File "/opt/anaconda3/envs/nwb-guide/lib/python3.9/site-packages/flask/views.py", line 109, in view
    return current_app.ensure_sync(self.dispatch_request)(**kwargs)
  File "/opt/anaconda3/envs/nwb-guide/lib/python3.9/site-packages/flask_restx/resource.py", line 46, in dispatch_request
    resp = meth(*args, **kwargs)
  File "/Users/garrettflynn/Documents/Github/nwb-guide/pyflask/apis/neuroconv.py", line 84, in post
    return get_metadata_schema(neuroconv_api.payload.get("source_data"), neuroconv_api.payload.get("interfaces"))
  File "/Users/garrettflynn/Documents/Github/nwb-guide/pyflask/manageNeuroconv/manage_neuroconv.py", line 240, in get_metadata_schema
    converter = instantiate_custom_converter(source_data, interfaces)
  File "/Users/garrettflynn/Documents/Github/nwb-guide/pyflask/manageNeuroconv/manage_neuroconv.py", line 203, in instantiate_custom_converter
    return CustomNWBConverter(source_data)
  File "/opt/anaconda3/envs/nwb-guide/lib/python3.9/site-packages/neuroconv/nwbconverter.py", line 65, in __init__
    self.data_interface_objects = {
  File "/opt/anaconda3/envs/nwb-guide/lib/python3.9/site-packages/neuroconv/nwbconverter.py", line 66, in <dictcomp>
[...]

PR #472, opened 2023-10-23
Update headers for search categories

garrettmflynn:
Addresses (ii) discussed in #467

Since these headers are simple markers less important than the interfaces / converters themselves, I opted for _smaller_ text while boosting the contrast in different ways beyond the list items.

## Before
Screenshot 2023-10-23 at 9 45 54 AM

## After
Screenshot 2023-10-23 at 9 45 36 AM


garrettmflynn:
Yep! Just added

CodyCBakerPhD (approved):

PR #474, opened 2023-10-26
Update Documentation for Poster

garrettmflynn:
This PR ensures that the user-focused installation instructions are accessible on the Read the Docs page instead of the README file.

https://nwb-guide.readthedocs.io/en/update-docs/installation.html


README.md
```diff
@@ -24,82 +28,71 @@ NWB GUIDE is a desktop app that provides a no-code user interface for converting
...
 ## Developer Installation
+### Clone the Repo
 
 Start by cloning the repository
 
 ```
 git clone https://github.com/catalystneuro/nwb-guide
 ```
 
+### Install Pythoh Dependencies
```
CodyCBakerPhD, line 43:
```suggestion
### Install Python Dependencies
```

CodyCBakerPhD:
My 2 cents - I'd say just go all in and move all the stuff from the README into official documentation, except the 'about' section with the demo

That way the readme is just a minimal landing page and we're not splitting documentation across locations

garrettmflynn:
Ah sure. I thought we might want to keep the points referencing the repo itself (e.g. Developer Setup, Repo Structure) directly where the "user" (developers in this case) would actually apply the knowledge gained. 

Otherwise, we're referencing some abstract operations in our Read the Docs that are, in my opinion, best understood as repo-dependent bits of documentation. Though I agree that if we're considering both Read the Docs and our GitHub page as websites, it's best to only have one—though the `README` text will uniquely be top-level in each developer's source code.

Can you con

garrettmflynn:
In other words, it's safe to assume only developers will reference the GitHub / README—so there may be a case for splitting user-focused and developer-focused (at least that which should be near at hand when starting) documentation

CodyCBakerPhD:
I would leave an informative sentence and link to the RTD in the README to resolve any such confusion (the GitHub page also links to it at the very top)

All of our other repos in the ecosystem follow this approach, as it allows us to expand the developer section at a larger scale over time

So yeah I'd go ahead and move the dev instructions to RTD

CodyCBakerPhD (approved):

PR #476, opened 2023-10-26
Validate on input delay

garrettmflynn:
This PR allows forms to validate on input delay (currently 1s) instead of waiting for the user to click off the input—as some don't realize to do this.

## Alpha Test Issues
1. Would be nice to be able to get feedback on the validity of your metadata before clicking out of the input but still when you’re done typing (e.g. if the user doesn’t change their input in 2s)

### DIFF
--- a/src/renderer/src/stories/JSONSchemaInput.js
+++ b/src/renderer/src/stories/JSONSchemaInput.js
@@ -120,13 +120,36 @@ export class JSONSchemaInput extends LitElement {
 
     getElement = () => this.shadowRoot.querySelector(".schema-input");
 
-    #updateData = (path, value) => {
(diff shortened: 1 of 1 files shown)


CodyCBakerPhD:
Works fine on the main metadata pages, but does not seem to apply to entering fields to the hands on table - so this is definetely fine as a first step

CodyCBakerPhD (approved):

garrettmflynn:
Unfortunately, HOT handles validation internally—and the same way as we were before (i.e. when the user unfocused the element). Since there are many types of cell editors, I haven't been able to find a uniform way to register user inputs for each cell. 

As such, forcing HOT validation to occur after a timeout doesn't seem like it's feasible in the short-term—though we may dig deeper in the future and/or implement this feature in our custom table Web Components.

PR #479, opened 2023-10-26
Update Schema Handling

garrettmflynn:
This PR updates how schema are handled, allowing for simple additions to existing schema across their usage in the GUIDE.

## Alpha Test Issues
- Exclude cluster groups has no description
- Stimulus notes description is the same as Data Collection - where does this come from - it seems incorrect
- Would be nice to have example keywords


schemas/json/base_metadata_schema.json
```diff
@@ -102,7 +102,7 @@
         },
         "stimulus_notes": {
           "type": "string",
-          "description": "Notes about data collection and analysis."
+          "description": "Notes about stimuli, such as how and where presented."
```
CodyCBakerPhD, line 105: But we maintain a separate base here? I see, if we need to make changes specific to GUIDE that wouldn't apply to NeuroConv

CodyCBakerPhD (approved):

PR #480, opened 2023-10-27
Update DANDI API Workflow

garrettmflynn:
This PR updates the DANDI API key workflow in alignment with the recent alpha tests. Instead of redirecting to the settings page, a modal provides the specific API key for the user to fill out without leaving the page. This input is validated and updates the global metadata. 

An additional check for whether the API key provided is _actually authorized_ has also been added, ensuring users will provide an API key that works—and which will fail and prompt them if it doesn't.

## Alpha Test Issues
- No links to DANDI
- Hard to get back to your conversion quickly
- DANDI API key validation is not informative
- Should be extremely clear that you actually have to create a dandiset (staging for now) for the conversion to upload there


CodyCBakerPhD (approved):

CodyCBakerPhD:
Yes, very cool! Glad it was that easy

PR #483, opened 2023-10-30
Global metadata on demand

garrettmflynn:
This PR allows users to pull up global metadata on demand and edit those properties. The Subject, Source Data, and File Metadata pages all have a new button near the header to toggle Global Metadata editing.

Additional related fixes were included, such as:
- Maintain form state on updates (i.e. accordions will stay open when updated with new globals)
- Do not allow users to specify an empty value if there are globals. Auto-fill with global when cleared—as this is how the property will be interpreted (and rendered on page reload)
- Ensure proper deletion of properties that are set to an empty string / undefined (as previously these would not clear...)

## User Testing Issues
- Made all single-session fixes to experimenter name but did not change the global level
- Missed the idea of the earlier global metadata pages (did not read the description, or was not clear?)
- would be nice to be able to specify the interface kwargs (Source Data Page) at a global / subject-specific level.
- Did not fill out global subject metadata on the Subject Table however, such as species and age [commented on how this was annoying to specify for each session]
- unclear how to triage inspector report feedback (e.g. global metadata is usually the best way to go for recurring issues)

### DIFF
--- a/schemas/base-metadata.schema.ts
+++ b/schemas/base-metadata.schema.ts
@@ -4,3 +4,30 @@ baseMetadataSchema.properties.Subject.properties.weight.unit = 'kg' // Add unit 
 
 
 export default baseMetadataSchema
+
+
+export const instanceSpecificFields = {
+    Subject: ["weight", "subject_id", "age", "date_of_birth", "age__reference"],
+    NWBFile: [
+        "session_id",
+        "session_start_time",
+        "identifier",
+        "data_collection",
+        "notes",
+        "pharmacolocy",
+        "session_description",
+        "slices",
+        "source_script",
+        "source_script_file_name",
+    ],
+};
+
+
+const globalSchema = structuredClone(baseMetadataSchema);
+Object.entries(globalSchema.properties).forEach(([globalProp, schema]) => {
+    instanceSpecificFields[globalProp]?.forEach((prop) =>  delete schema.properties[prop]);
+});
+
+export {
+    globalSchema
+}
--- a/schemas/json/base_metadata_schema.json
+++ b/schemas/json/base_metadata_schema.json
@@ -1,8 +1,6 @@
 {
   "$schema": "http://json-schema.org/draft-07/schema#",
   "$id": "base_metafile.schema.json",
-  "title": "Base schema for the metafile",
-  "description": "Base schema for the metafile",
   "version": "0.1.0",
   "type": "object",
   "required": ["NWBFile"],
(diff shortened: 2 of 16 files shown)


CodyCBakerPhD:
@garrettmflynn small conflict here

garrettmflynn:
Alright this is properly merged now

CodyCBakerPhD:
This design wasn't very intuitive for me personally

I went to the file metadata page and started to enter a couple manual values for a field that should have been globally defined, then I clicked the new button and this is what it said

image

Next I manually navigated back to the global metadata page for NWB file metadata, added the field there, hit 'save and continue', and went back to the session level pages, which now have the global value

But when I still click on the new 'Edit Global Metadata' button I see the same screen with no values to update, and pressing update didn't seem to change anything on any session pages (even after I manually altered some of them to new values)

Maybe this is due to a bug, but can you describe what the exact workflow for the usage of this button is supposed to be? We had discussed it simply being a link back to the global metadata page (to raise user awareness of the early step that they only realize is useful later on in the process)

garrettmflynn:
There were a few bugs—so I've fixed those now :)

Generally, though, I see where the confusion lies. What I was describing was editing the **global metadata data structure** through a button that opens all / specific properties on demand. As such, the workflow is to (1) press the button to open all the global metadata on that page, (2) update any properties you'd like, (3) press the Update button, (4) re-render the form with the new globals filled in. I've went through this many times for the latest fixes—so it should just work now.

We'd also described a pop-up when entering the global metadata pages so that users were forced to explicitly decide to add this metadata or not early in the pipeline—though I have not added this because of the availability of the global data through these buttons.

CodyCBakerPhD:
Oh, I see - I'll give it another fresh try then

Some conflicts now though

CodyCBakerPhD:
Also looks like tests are failing after the resolution

garrettmflynn:
Got it my bad. Should be fixed now

CodyCBakerPhD:
Ahh, yes... This is much better when it's working lol

Looks great!

CodyCBakerPhD (approved):

CodyCBakerPhD:
The only thing I'd change, which I leave as a simple follow-up, would be to remove the button from the 'source data' page since it has no relevance to that specific step

Attentive users will also have their attention drawn to the new button when it then appears on the metadata page right after thatt too

garrettmflynn:
One of the drivers of this PR in the first place was Felix's suggestion to include a way to specify global kwargs for the interfaces—so wouldn't we want to keep this for Source Data?

CodyCBakerPhD:
Ohhh I did not notice it was global source data on that page

image

In that case it's just the name of the button that needs to change, it still says 'metadata'

image

Also might as well skip any/all interface banners that don't have extra arguments (that result in that 'no additional properties to render' text)

Issue #482, opened 2023-10-30
Separate session management from global subject table


CodyCBakerPhD:
Also resolve that outstanding issue with adding new sessions to the pipeline from Bens

garrettmflynn:
Just to provide some more details for my sake here:
1. We are splitting the Subject Table into two tables: a Subject Metadata table (with Subject ID as a main column) and a Session ID table where each Session ID is associated with one Subject ID (constrained to exising IDs).
2. This helps people think in terms of standard input cells whenever they use our tables (rather than the CSV-like list editor).
3. When changing the Subject ID, it should auto-update on the second table

CodyCBakerPhD:
While the current method of manually adding more sessions could be improved in separate ways, I'd say this separation is no longer needed

PR #484, opened 2023-10-30
Allow editing sessions on new rows

### LINKED ISSUE #212: HOT New Rows Don't Allow for Editing Custom Renderer / Editor Columns (e.g. sessions)
When attempting to edit the `sessions` property on a new row of the Subjects table, the HOT instance will freeze. This isn't the case for existing table entries—and, as far as I can tell, rows are generated in the same manner in either case.

I've narrowed the problem to when the custom ArrayEditor class in `hot.js` returns a value from the `getValue` method, after which I haven't been able to print out this value.

While this would be nice to show for the upcoming NWB GUIDE demo video, it's easy enough to skip around this interaction because I can still add a subject to the existing row.

## To Reproduce
1. Go to `guided/subjects`
2. Right-click on the table and add a new row
3. Type into the `sessions` box in that new row
4. See that the table has frozen and no longer supports user interactions

garrettmflynn:
fix #212 and #482 

This PR fixes the issue where you can't edit sessions on new rows.

Though slightly different than #482, the root problem of this issue is solved. @bendichter Would we still like the Subject table to be separated into two tables, where one handles subject metadata and the other tracks sessions? The specifications for this split are still a bit unclear to me—so I thought I'd bring it up before attempting such a refactor.


CodyCBakerPhD:
@garrettmflynn Tests failing here too

garrettmflynn:
Looks to be fixed after merging with `main`. Doing that again now

CodyCBakerPhD (approved):

PR #486, opened 2023-10-31
Fallback to n_jobs=1 if NWB Inspector fails

garrettmflynn:
This PR is a quick fix for https://github.com/NeurodataWithoutBorders/nwbinspector/issues/415 where the GUIDE will fall back to `n_jobs=1` if using multiple jobs fails.

### DIFF
--- a/pyflask/manageNeuroconv/manage_neuroconv.py
+++ b/pyflask/manageNeuroconv/manage_neuroconv.py
@@ -593,17 +593,21 @@ def inspect_nwb_folder(payload):
     from nwbinspector import inspect_all, load_config
     from nwbinspector.nwbinspector import InspectorOutputJSONEncoder
 
-    messages = list(
-        inspect_all(
-            n_jobs=-2,  # uses number of CPU - 1
-            ignore=[
-                "check_description",
-                "check_data_orientation",
-            ],  # TODO: remove when metadata control is exposed
-            config=load_config(filepath_or_keyword="dandi"),
-            **payload,
-        )
+    kwargs = dict(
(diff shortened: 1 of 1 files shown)


pyflask/manageNeuroconv/manage_neuroconv.py
```diff
@@ -593,18 +593,22 @@ def inspect_nwb_folder(payload):
...
+            "check_description",
+            "check_data_orientation",
+        ],  # TODO: remove when metadata control is exposed
+        config=load_config(filepath_or_keyword="dandi"),
+        **payload,
     )
 
+    try:
+        messages = list(inspect_all(**kwargs))
+    except:
```
CodyCBakerPhD, line 608: Please don't use bare `except` - add a match to the exact type that occurs with the pickling problem if you can

garrettmflynn: Updated to check for a `PicklingError`. Is it alright to bypass anything from pickle as an error with multithreading?

CodyCBakerPhD: > Updated to check for a PicklingError. Is it alright to bypass anything from pickle as an error with multithreading?

Probably OK, but I always prefer to be as explicit as possible; you can add a second check `assert "some expected part of the specific message for this specific isssue" in str(e)` to be extra sure it's the problem you see

garrettmflynn: Okay done.

CodyCBakerPhD (approved):

PR #489, opened 2023-10-31
Add MiniscopeConverter

garrettmflynn:
This PR adds the MiniscopeConverter, which currently has an error from the NWB Inspector but can complete a full conversion despite this.

Screenshot 2023-10-31 at 10 56 48 AM


CodyCBakerPhD:
Indeed files look fine, but Inspector message is actually a PyNWB validation error. Also the CLI report shows more info about this issue

image

The validation error is a known and has been raised upstream for a while. It might actually prevent DANDI upload from within the app, not sure since DANDI uses a slightly different strategy for assessing that aspect

CodyCBakerPhD:
Realized the base of this too late

Issue #153, opened 2023-05-02
Cannot paste sessions using Handsontable


garrettmflynn:
Since the Table component (which wraps Handsontable) requires a custom Editor for the sessions property, it disables the correct paste behavior.

CodyCBakerPhD:
Summarized and moved to https://github.com/NeurodataWithoutBorders/nwb-guide/issues/846

Re-open again whenever this starts active development

PR #491, opened 2023-10-31
Move Subject Table Validation Inside + Allow Copy-Pasting Sessions (when empty)

garrettmflynn:
This PR moves all of the subject table validation steps inside the Table itself—requiring that Subject ID and Sessions are both provided to validate. 

A fix for copy-pasting is also provided, though this only works when the Subject cell is empty (because of an internal HOT issue #153)


CodyCBakerPhD:
As discussed, attempt to fix the issue where I add a new subject but use an existing ID, have it not override all the previous work

Perhaps automatically append some index/format to the new duplicated ID, similar to what happens when you copy+paste a file in the same directory on most OS?

garrettmflynn:
Fixed! I'm hesitant to add anything to the ID automatically, as this'll likely only happen for naive copy/paste operations—where I'd rather let the user know they need to input their own unique value.

Additionally, this'll be managed internally for the next iteration of this (i.e. two tables). So will go away as a possible problem soon.

CodyCBakerPhD:
@garrettmflynn Small conflict now

garrettmflynn:
Fixed!

CodyCBakerPhD:
OK sure, an example like this is self evident

image

CodyCBakerPhD (approved):

CodyCBakerPhD:
@garrettmflynn One question arose w.r.t. button placement in Chromatic (may have occurred in previous PR without me realizing until now)

If unrelated to changes of this PR we can go ahead and merge as-is and follow up with a specific fix for that

garrettmflynn:
You're right, this is a mistake that was on main. I've pushed there directly—so it should be gone.

CodyCBakerPhD:
@garrettmflynn Cool, just a small conflict here now

CodyCBakerPhD (approved):

PR #492, opened 2023-11-01
Edit list order + values

garrettmflynn:
This PR allows users to edit the order and values of list items generally (e.g. Experimenters, Video File Paths, etc.)

## User Test Issues
- No edit button for experimenters; couldn’t easily change order either


CodyCBakerPhD:
As discussed, backspacing all text from box selection should delete the entry instead of attempt to populate default

garrettmflynn:
Updated to delete the entry if the value is cleared. Also added a placeholder for where the item will move to if dragged

CodyCBakerPhD:
Feels much better now, thanks

CodyCBakerPhD (approved):

PR #496, opened 2023-11-02
Allow Structured Forms for Array Fields

### LINKED ISSUE #432: [Feature] Swap experimenter entries to structured form
Several users mentioned that it was mildly annoying to have to type out an experimenter name as a single string only to be informed of its incorrect form right after typing; initial structuring to constrain to the form could be useful

@garrettmflynn How hard would this be?

For example,

```
1. [Lastname], [Firstname], [Middle name or initial]
2.  ...
[Add experimenter]
```

garrettmflynn:
fix #432 

This PR allows for the rendering of object arrays as nested forms, as well as the conversion of this result into a string.

### DIFF
--- a/pyflask/app.py
+++ b/pyflask/app.py
@@ -95,6 +95,16 @@ def send_preview(path):
     return send_from_directory(STUB_SAVE_FOLDER_PATH, path)
 
 
+@app.route("/cpus")
+def get_cpu_count():
+    from psutil import cpu_count
+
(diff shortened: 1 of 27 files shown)


CodyCBakerPhD:
Feels great! Some tests failing though

garrettmflynn:
Fixed!

CodyCBakerPhD (approved):

PR #499, opened 2023-11-06
Use latest dandi dev branch

CodyCBakerPhD:
Thoroughly testing latest distributable on Windows and Mac - I too run into the seg fault error on my Windows device via the distributable (so at least we have a way to test it now...), even after we built in the override

I suspect maybe the older DANDI version is to blame - attempting a rebuild from this branch


garrettmflynn (approved):
Ah interesting. Let's see if it works! 

We did see this issue on Ben's Mac too—so we'll just want to remember to bump all the environment files if this works.

CodyCBakerPhD:
Yeah I'll be testing on my Mac here soon too

CodyCBakerPhD:
@garrettmflynn Do you think we should add a note somewhere on the upload pages that if the user encounters an error with the jobs selector at `-1`, they should try `1`?

If so feel free to add that on this PR (I'm not sure of the best way to do so)

nwb-guide.spec
```diff
@@ -11,6 +11,12 @@ datas = [('./paths.config.json', '.'), ('./package.json', '.')]
 binaries = []
 hiddenimports = ['scipy._distributor_init', 'scipy._lib.messagestream', 'scipy._lib._ccallback', 'scipy._lib._testutils', 'email_validator']
 datas += collect_data_files('jsonschema_specifications')
+tmp_ret = collect_all('dandi')
```
CodyCBakerPhD, line 14: I have absolutely no idea where or how the package was getting included previously, but this actually makes it show up in the pyflask collection

CodyCBakerPhD:
@garrettmflynn I can confirm that the v0.11 releases created from this branch both reproduce the issue when jobs is `-1` and upload without error when set to `1`

nwb-guide.spec
```diff
@@ -19,6 +25,8 @@ tmp_ret = collect_all('pynwb')
 datas += tmp_ret[0]; binaries += tmp_ret[1]; hiddenimports += tmp_ret[2]
 tmp_ret = collect_all('hdmf')
 datas += tmp_ret[0]; binaries += tmp_ret[1]; hiddenimports += tmp_ret[2]
+tmp_ret = collect_all('hdmf_zarr')
+datas += tmp_ret[0]; binaries += tmp_ret[1]; hiddenimports += tmp_ret[2]
```
CodyCBakerPhD, line 29: Including this to be compatible as of latest merges on NeuroConv `main`

garrettmflynn:
Just added the note about using a value of `1`. Sounds like this is good to go!

On my end, I've had issues building locally (the app fails on open with an error), so trying to figure that out. Might be something funky with my JS environment

garrettmflynn:
Just worked when using v0.0.11 from Github on my Mac (Intel) :)

environments/environment-Windows.yml
```diff
@@ -18,5 +18,6 @@ dependencies:
       - flask-cors === 3.0.10
       - flask_restx == 1.1.0
       - neuroconv @ git+https://github.com/catalystneuro/neuroconv.git@main#neuroconv[full]
+      - dandi @ git+https://github.com/dandi/dandi-cli.git@master
```
CodyCBakerPhD, line 21:
```suggestion
      - dandi >= 0.57.0
```

DANDI just cut a new release lol

PR #505, opened 2023-11-09
Ophys Metadata

garrettmflynn:
This PR enables the Ophys Metadata and allow for editing table properties correctly to update the File Metadata structure passed to the server.

Ongoing work on this PR will include the following updates:
1. Inter-table updates for valid device / imaging plane selection.
2. Updating NeuroConv-internal representations (similar to Ecephys)
3. Ensure we actually want to represent all of the metadata
    - As per a recent discussion, the PlaneSegmentationTable could have tens of thousands of rows (ROIs). We could render a subset of the table and otherwise allow for CSV import / export using BasicTable.


CodyCBakerPhD:
This is a great first draft!

Starting with ImagingInterface metadata

- disable `manifold`, `unit`, and `conversion` in `ImagingPlane` section (in general, drop anything whose description includes the phrase 'DEPRECATED: ...')
- drop `format`, `starting_frame`, `starting_time`+`rate` (will be reintroduced in temporal alignment feature; also don't want user overriding known `rate`), `bits_per_pixel`, `dimension` (this can be set automatically, don't want user to provide something that conflicts), `control`, `control_description`, and `device` (confused why this one is even in the schema for the 2P series lol - it's in the ImagingPlane link after all...) from `TwoPhotonSeries` section
- reorgnize the `TwoPhotonSeries` columns to interrelate a bit more; suggestion, `name` -> `description` (one caveat - sometimes this is a big JSON dictionary with acquisition settings so it doesn't render super well - see the ScanImage testing data for an example) -> `scan_line_rate` -> `field_of_view` -> `unit` (see next comment) -> `conversion` -> `offset`
- One idea would be to ask if the user knows the `unit` of the 2P data (yes or no), which is quite uncommon for raw microscopy. If they say 'no' then drop `unit` (set to `n.a.` auomatically) as well as `conversion` and `offset` since they have no meaning in the metadata any more
- reorganize the `ImagingPlane` columns similarly, at least starting with `name` -> `description` -> `Device` -> `OpticChannel`

I love the nested table for OpticChannel BTW, we should think of ways to utilize that in other metadata places

garrettmflynn:
@CodyCBakerPhD Just flagging that this is ready to move onto the SegmentationInterface metadata. Let me know what sort of changes you'd like to be made there!

CodyCBakerPhD:
Status update: awaiting patternProperty fixes on segmentation things over on NeuroConv

CodyCBakerPhD:
@garrettmflynn Lots of merge conflicts here now

garrettmflynn:
From our meeting today, it seems that we are relying on `additionalProperties` to handle some fields like `ImageSegmentation`. This is fine if we don't expect users to add their own additional properties beyond the defaults—however, it becomes somewhat problematic if they are since I can't know the format of what they'll put down beforehand.

Is it possible to move these to clearly defined `patternProperties` or do we need to have a way to handle these arbitrary additional properties?

garrettmflynn:
@CodyCBakerPhD Now might be a good time for another review of the expected behaviors that are missing here. 

I've made it so that you can add / remove / edit pattern properties. Hopefully these interactions are seamless now. 

There are still a few items on the docket:
1. Warnings before deleting a property that has interactions (e.g. devices)
2. Invalidation based on interactions (same as above)
3. Selecting "I don't know" instead of auto-filling values (e.g. NaN for emission_lambda)

Anything else you see? Additional details about the above might also be helpful before I dig in—otherwise, I'll start approaching anything at hand related to these desired behaviors / interactions.

CodyCBakerPhD:
Seeing the following on the standalone DANDI upload page

image

CodyCBakerPhD:
Also the standalone Inspection page

image

Confirmed both of these issues are not present on `main`

CodyCBakerPhD:
Also can't create a new pipeline to test this out

```javascript
Uncaught (in promise) Error: 4 JSON Schema errors on this form.
    at JSONSchemaForm.throw (JSONSchemaForm.js:387:15)
    at JSONSchemaForm.validate (JSONSchemaForm.js:441:28)
    at async GuidedFooter.onNext [as __onNext] (GuidedNewDatasetInfo.js:44:13)
```

preceded by

```javascript
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading '0')
    at #render (JSONSchemaInput.js:438:108)
    at JSONSchemaInput.render (JSONSchemaInput.js:208:35)
    at JSONSchemaInput.update (lit-element.ts:160:24)
    at JSONSchemaInput.performUpdate (reactive-element.ts:1329:14)
    at JSONSchemaInput.scheduleUpdate (reactive-element.ts:1261:17)
    at JSONSchemaInput.__enqueueUpdate (reactive-element.ts:1233:25)
```

garrettmflynn:
Thanks for flagging these issues! I previously added some logic to handle additional properties when present, but didn't fully account for the fact that I need to render those that _do_ exist, but _do not_ need to render any header / list / button of additional properties if they don't (since we can't add properties that we don't know the structure of). I'd been explicitly setting `additionalProperties: false` on the schema I knew that we didn't want additional properties to be shown—though I obviously missed a few.

Based on these limitations, I've taken a better strategy that avoids rendering the additional properties group if none exist. on the initial set of property values provided to the JSONSchemaForm instance. 

Additionally, you couldn't create a new pipeline for a similar reason: I'd set `additionalProperties: false` on that schema but continued to pass an object that _did_ have additional properties (i.e. the full project metadata object properties) that were previously unrendered. Because of these changes, I'm now removing additional properties if they are explicitly not allowed.

CodyCBakerPhD:
Now getting this on the file metadata page (currently testing with only a single imaging interface selected)

```javascript
TypeError: Cannot read properties of undefined (reading 'optical_channel')
    at preprocessMetadataSchema (base-metadata.schema.ts:100:43)
    at createForm (GuidedMetadata.js:176:21)
    at utils.js:68:73
    at Array.map (<anonymous>)
    at utils.js:68:45
    at Array.map (<anonymous>)
    at mapSessions (utils.js:67:10)
    at GuidedMetadataPage.mapSessions (Page.js:116:63)
    at GuidedMetadataPage.render (GuidedMetadata.js:257:27)
    at GuidedMetadataPage.update (lit-element.ts:160:24)
```

garrettmflynn:
Looking into it. Can you let me know the specific interface and source data?

CodyCBakerPhD:
In this case, both ScanImage 'volume' .tif and Bruker single plane (the only into the void one that is valid for it without needing the stream name)

garrettmflynn:
Got it. I was replacing `$ref` values inside the JSON Schema form but didn't realize that these would be inconsistently used in the metadata given back from the converter (i.e. imagingPlane.items is declared using a reference with ScanImage, but not Caiman).

I've made it so that these references are resolved before any processing by the `base-metadata.schema.ts` transformation function.

CodyCBakerPhD:
@garrettmflynn More conflicts now

garrettmflynn:
@CodyCBakerPhD Updated the NaN representation. Let me know if this now works on your end.

garrettmflynn:
@CodyCBakerPhD Updated based on our discussion to show the Plane Metadata tables as soon as possible.

CodyCBakerPhD:
@garrettmflynn One small conflict present ATM and then I'll take another look

CodyCBakerPhD:
Looking much better now!

One small detail, IDK how easy it is to improve but for validation of the excitation lambda, it's only appearing on the outer page not the inner one

image

image

Minor detail: I also had to double-click to open the inner window, not single click (did this change from a previous state? Or am I imagining things? Either way, the instruction to 'click to view cell' should probably be 'double-click to view cell' in that case)

CodyCBakerPhD:
Also, there was a really good popup info instruction for inputting NaN for unknown values for the emission field, but I can't find a similar instruction for the excitation lambda

CodyCBakerPhD:
I would remove conversion + offset + unit from the TwoPhotonSeries metadata for now to simplify the view of it; there's so many properties on it, and these are never specified in practice

CodyCBakerPhD:
Also remove 'starting_frame' and 'format', these aren't allowed through NeuroConv/DANDI effectively

CodyCBakerPhD:
Also if you could remove the 'starting_time' from TwoPhotonSeries, as that will be handled in the #566 and similar follow-ups

garrettmflynn:
Got it. Looks like the `propsToIgnore` variable slipped through without being applied to the nested tables.

Just fixed that. You can refer to that for the expected properties to be ignored.

CodyCBakerPhD:
With a simple ScanImage pipeline applied to the single.tif file, the preview succeeds but gives report that renders as

image

Opened #568 that would make it easier to debug where this came from

I think it might be the 'dimension' field?

CodyCBakerPhD:
On a simple pipeline testing only Suite2P, the TwoPhotonSeries section of the metadata shows up even though it is empty

image

Would be better to not show this in this case

CodyCBakerPhD:
Also if you could remove starting_time and rate from fluoresence/DfoverF (any RoiResponseSeries field, basically); again, temporal alignment feature is what controls this, I don't even know if specifying the metadata would change the conversion

CodyCBakerPhD:
The segmentation metadata looks a lot better now!

I don't think we need to show the property_key, but not 100% sure; would it be possible to 'hide' it but have it still be there?

Could you instead replace that column on the flat table with the name (value) of the plane segmentation (contained in the image segmentations object) that matches the key from the patternProperty (this is the intended usage of these keys from the programmatic perspective; but in the GUIDE it would make more sense to 'follow' those keys to their values in some sense)

Specifically, in the example below

image


the first column would be 'Plane Segmentation' and the value for each row would be 'PlaneSegmentationChan1Plane0' as seen from expanding the image segmentation

image

CodyCBakerPhD:
Also, I think it might be more convenient to expand the property of the image segmentation object similar to what you did for the flourescence fields

Basically, each row would be a different 'additional property' and the total table would appear as a concatenation of all the popups from pressing 'edit' on the current layout

garrettmflynn:
> One small detail, IDK how easy it is to improve but for validation of the excitation lambda, it's only appearing on the outer page not the inner one

Fixed this.

> Minor detail: I also had to double-click to open the inner window, not single click (did this change from a previous state? Or am I imagining things? Either way, the instruction to 'click to view cell' should probably be 'double-click to view cell' in that case)

We talked about this during our last meeting. All other cells require a double-click to edit, so that convention is maintained here. I've updated the text attached to this.



Will start working on your other suggestions now.

CodyCBakerPhD:
@garrettmflynn backend tests failing now on this PR

Also just discovered I could press enter while typing in a value to the emission lambda field to cause this:

image

which is maybe fine in an of itself (pending another PR where I think the 'enter' functionality was requested to "enter" information to a cell; from Olivers user test I believe?) but then I can backspace and click off to get this:

which by all accounts looks the same as just typing `NaN` then clicking off

CodyCBakerPhD:
Possibly related; clicking into the description field accidentally then clicking off leads to this

image

CodyCBakerPhD:
Everything is starting to feel a bit better now; minor change requested to the order of columns on the ImagingPlane

Starting from 'location'

... -> location -> reference frame -> imaging rate -> grid spacing -> grid spacing unit -> origin coords -> origin coords unit

then that part should be good to call done

I still see an empty table for TwoPhotonSeries even when running a strictly Suite2p conversion though

image

CodyCBakerPhD:
Also from what we talked about; replacing the 'property key' column of all segmentation RoiResponseSeries to be the name (value in ImageSegmentation list of the same key) of the PlaneSegmentation its ROIs refer to

While you're at it, ideal column order for RoiResponseSeries is

Name -> description -> comments -> control -> control description -> resolution

For now, suppressing conversion/offset/unit for that since typically unrealistic to have scientific units for that

Also for now, suppressing starting_time and rate fields since that belongs to the time alignment PR

Also we talking about pulling out the table of the ImageSegmentation into the field instead of it being the 'add property' / 'edit' buttons

garrettmflynn:
Thanks for the updates! Gonna start tackling all this tomorrow, didn't get a chance to catch all of the last round of updates while organizing the tqdm repo.

CodyCBakerPhD:
@garrettmflynn Small conflict

garrettmflynn:
The behavior discussed in https://github.com/NeurodataWithoutBorders/nwb-guide/pull/505#issuecomment-1896641850 is quite complicated. I think we should discuss how we'll want to move forward with this tomorrow, just the ensure I've got the right idea.

Otherwise, the only untackled comments are the ScanImage issue (partially dependent on #569), pressing enter on the NaN cells, and the failing backend tests.

garrettmflynn:
> I think it might be the 'dimension' field?
Actually, it looks like this issue is because of `origin_coords` and `grid_spacing`?
```
{
  "header": {
    "Timestamp": "2024-01-23 15:36:14.295038-08:00",
    "Platform": "macOS-13.5-arm64-arm-64bit",
    "NWBInspector_version": "0.4.31"
  },
  "messages": [
    {
      "message": "incorrect shape - expected '[[2], [3]]', got '(0,)'",
      "importance": "PYNWB_VALIDATION",
      "severity": "LOW",
      "check_function_name": "ImagingPlane/origin_coords",
      "object_type": null,
      "object_name": null,
      "location": "general/optophysiology/ImagingPlane/origin_coords"
    },
    {
      "message": "incorrect shape - expected '[[2], [3]]', got '(0,)'",
      "importance": "PYNWB_VALIDATION",
      "severity": "LOW",
      "check_function_name": "ImagingPlane/grid_spacing",
      "object_type": null,
      "object_name": null,
      "location": "general/optophysiology/ImagingPlane/grid_spacing"
    },
    {
      "message": "Experimenter is missing.",
      "importance": "BEST_PRACTICE_SUGGESTION",
      "severity": "LOW",
      "check_function_name": "check_experimenter_exists",
      "object_type": "NWBFile",
      "object_name": "root",
      "location": "/"
    },
    {
      "message": "Experiment description is missing.",
      "importance": "BEST_PRACTICE_SUGGESTION",
      "severity": "LOW",
      "check_function_name": "check_experiment_description",
      "object_type": "NWBFile",
      "object_name": "root",
      "location": "/"
    },
    {
      "message": "Metadata /general/institution is missing.",
      "importance": "BEST_PRACTICE_SUGGESTION",
      "severity": "LOW",
      "check_function_name": "check_institution",
      "object_type": "NWBFile",
      "object_name": "root",
      "location": "/"
    },
    {
      "message": "Metadata /general/keywords is missing.",
      "importance": "BEST_PRACTICE_SUGGESTION",
      "severity": "LOW",
      "check_function_name": "check_keywords",
      "object_type": "NWBFile",
      "object_name": "root",
      "location": "/"
    }
  ]
}
```

CodyCBakerPhD:
Trying to add some ophys pipelines to the testing YAML but the generation button has disappeared - any ideas?

garrettmflynn:
@CodyCBakerPhD Fixed the ScanImage array auto-fill issue!

garrettmflynn:
Starting with this PR, there's a more effective way to request access to form elements (`getFormElement` vs. `getInput`, `getForm`, and `getTable` separately). One of the old methods was used to inject the pipeline generation button, so that's why it didn't show up.

CodyCBakerPhD:
Almost done!

Reminder to

- remove conversion/offset/unit from RoiResponseSeries
- remove display of property_key from all the tables
- reorder name -> description on SegmentationImages

CodyCBakerPhD:
Oh, I guess I did just notice the 'Dynamic Fields' checkbox. Is that necessary since the tabs themselves are expandable?

garrettmflynn:
@CodyCBakerPhD Alright all the updates discussed earlier today should be completed.

CodyCBakerPhD:
OK, getting pathological with my metadata overrides now

Odd behavior observed in any segmentation pipeline (this PR can now autogenerate suite2p and caiman for easy testing): go to the Fluorescence level and change `resolution` from `-1` to any other number - I then get informed that it must be a float not an int, which I then correct to be a float, but still get the error on attempting to generate preview - then trying to restore it to the default of `-1` continues to cause the error beyond that

CodyCBakerPhD:
Though I notice I'm actually getting this error now even when I don't modify the `resolution` field at all; seems that manually modifying any field on the table will trigger it - are you able to reproduce?

CodyCBakerPhD:
Also on the current auto-generated ScanImage pipeline, I get

```javascript
base-metadata.schema.ts:95 Uncaught (in promise) TypeError: Cannot set properties of undefined (setting 'order')
    at preprocessMetadataSchema (base-metadata.schema.ts:95:27)
    at createForm (GuidedMetadata.js:187:32)
    at utils.js:69:73
    at Array.map (<anonymous>)
    at utils.js:69:45
    at Array.map (<anonymous>)
    at mapSessions (utils.js:68:10)
    at GuidedMetadataPage.mapSessions (Page.js:117:63)
    at GuidedMetadataPage.render (GuidedMetadata.js:444:27)
    at GuidedMetadataPage.update (lit-element.ts:160:24)
```

on the file metadata page

CodyCBakerPhD:
Very weird issue possible related. Steps to reproduce

1) autogenerate fresh testing pipelines
2) start with ScanImage, go to the file metadata page if you can (+/- the error above if you see it, either way proceed to 3)
3) exit out and go to suite2p or caiman
4) proceed to file metadata page
5) exit/save and go back to scanimage pipeline
6) I see the suite2p/caiman metadata page on the scanimage pipeline now

garrettmflynn:
> Also on the current auto-generated ScanImage pipeline, I get
> 
> ```js
> base-metadata.schema.ts:95 Uncaught (in promise) TypeError: Cannot set properties of undefined (setting 'order')
>     at preprocessMetadataSchema (base-metadata.schema.ts:95:27)
>     at createForm (GuidedMetadata.js:187:32)
>     at utils.js:69:73
>     at Array.map (<anonymous>)
>     at utils.js:69:45
>     at Array.map (<anonymous>)
>     at mapSessions (utils.js:68:10)
>     at GuidedMetadataPage.mapSessions (Page.js:117:63)
>     at GuidedMetadataPage.render (GuidedMetadata.js:444:27)
>     at GuidedMetadataPage.update (lit-element.ts:160:24)
> ```
> 
> on the file metadata page

Oddly, this example has the `items` property defined as `undefined` on the `ImagingPlane` or `TwoPhotonSeries` schema. Should this be passed like this from NeuroConv?

garrettmflynn:
Hmm this appears to be consistent now, though I'd never seen it before.

garrettmflynn:
Oh looks like the `definitions` resolution is not happening correctly anymore. One sec

garrettmflynn:
> Odd behavior observed in any segmentation pipeline (this PR can now autogenerate suite2p and caiman for easy testing): go to the Fluorescence level and change resolution from -1 to any other number - I then get informed that it must be a float not an int, which I then correct to be a float, but still get the error on attempting to generate preview - then trying to restore it to the default of -1 continues to cause the error beyond that

So the core problem here is that JavaScript does not have the ability to maintain floats that are just integers (i.e. `-1.0` is equivalent to `-1`, and will be sent to Python as such). 

When you haven't manually changed the value, there will be nothing sent to Python for that property—allowing the default to take effect. However, manually providing `-1` (or `-1.0`) will result in a `-1` integer registered in Python.

We will probably have to coerce to a float similar to our previous NaN parsing on the Python end.

garrettmflynn:
Can you check if this still happens? https://github.com/NeurodataWithoutBorders/nwb-guide/pull/505#issuecomment-1910505757 

I wasn't able to get this to replicate now. Originally, I wasn't able to get _anything_ to render on ScanImage because of the missing items before, so that might be why it didn't update the previous render.

CodyCBakerPhD:
> Can you check if this still happens? https://github.com/NeurodataWithoutBorders/nwb-guide/pull/505#issuecomment-1910505757

Ah, apparently not


> So the core problem here is that JavaScript does not have the ability to maintain floats that are just integers (i.e. -1.0 is equivalent to -1, and will be sent to Python as such).
> 
> When you haven't manually changed the value, there will be nothing sent to Python for that property—allowing the default to take effect. However, manually providing -1 (or -1.0) will result in a -1 integer registered in Python.
>
> We will probably have to coerce to a float similar to our previous NaN parsing on the Python end.

Right, that all makes sense. Casting would be fine (the default should also be -1.0 in that case anyway); checking the NeuroConv hardcoded schemas though, I don't actually see exposure of the `resolution` argument (again, one of those I don't think majority of people would ever willingly specify, or have even indicated knowing except in special cases) so if it's easier for you at the moment I say we just suppress it from all the tables

CodyCBakerPhD:
I found a couple unrelated bugs while thoroughly testing this; they can be solved in follow-ups

I think this is just about ready to merge once the resolution thing is dealt with

garrettmflynn:
Figured out the root of the `resolution` issue. I have been coercing to floats for some time now, though I wasn't searching through the `patternProperties` for values to coerce. This is now implemented!

CodyCBakerPhD (approved):

PR #507, opened 2023-11-12
Minor text fixes for developer guide

rly:
Super minor text fixes in the developer guide:

- Use the NeurodataWithoutBorders org github link for `git clone`
- Change the formatting of the OS-specific instructions for installing the Python env so that the instructions to activate the conda environment do not appear to be just for Linux
- Fix incorrect path to `**src/renderer/assets**`
- Wrap long text
- And I guess pre-commit wants to remove some empty lines from other files...


garrettmflynn (approved):
Looks great. Thank you for passing through this!

PR #508, opened 2023-11-12
Try to left align text of navbar links

rly:
The text on the left nav bar is not left-aligned. This PR adjusts the margins of the SVGs to correct for that. I assume there is a reason the icons are different sizes. I think it does look better at these sizes, so I did not touch that.

Before:
Screenshot 2023-11-12 at 12 56 37 AM

After:
Screenshot 2023-11-12 at 12 53 25 AM

It's not pixel perfect if you zoom in - I'm not sure why - but it does look better to me now.


rly:
PS This felt too small of an issue to create an issue ticket for, but let me know if that would be preferred.

garrettmflynn:
I also tried a CSS-based solution—but this is the best approach given the inconsistent widths of the icons themselves. 

When setting things to a uniform 25px outside of this PR, this is what it would look like:
Screenshot 2023-11-12 at 8 00 30 AM

Not great balance between the original SODA icons and the new ones for the GUIDE. Could change them for new [Material Icons](https://fonts.google.com/icons?selected=Material+Symbols+Outlined:save_as:FILL@0;wght@400;GRAD@0;opsz@48&icon.query=create), but that's a broader discussion.

garrettmflynn (approved):
This is great. Thank you @rly!

PR #514, opened 2023-11-12
Dynamically populate hidden imports rather than use the PyInstaller CLI

garrettmflynn:
Fixes issues building and running the GUIDE distributable on M2 because of a missing `scipy` DLL invoked in the import test on initialization. This implements a more robust way to catch all DLLs in `scipy` so we aren't surprised in the future.


CodyCBakerPhD (approved):

CodyCBakerPhD:
Everything appears to work well with this

PR #519, opened 2023-11-17
[DO NOT MERGE] Linux branch

CodyCBakerPhD:
# DO NOT MERGE

Opening this draft PR to make it easy to keep up to date with main


garrettmflynn:
As discussed by @rly in the Slack, we may want to merge this into `main` for the sake of simplicity if this works for Mac and Windows—even if we have a slightly larger distributable size.

@CodyCBakerPhD Is there a way that you can access the builds created here and see whether they are even of a different size than the `main` branch?

CodyCBakerPhD:
> As discussed by @rly in the Slack, we may want to merge this into main for the sake of simplicity if this works for Mac and Windows—even if we have a slightly larger distributable size.
>
> @CodyCBakerPhD Is there a way that you can access the builds created here and see whether they are even of a different size than the main branch?

The reason this has been on a side branch all this time is because we _cannot_ successfully build any distributable from this branch

https://github.com/NeurodataWithoutBorders/nwb-guide/actions/runs/6946096739/job/18896843132

https://github.com/NeurodataWithoutBorders/nwb-guide/actions/runs/6946098457/job/18896849262

rly:
Ah, I see. For reference, the error from electron-builder is 
```
  ⨯ Package "electron" is only allowed in "devDependencies". Please remove it from the "dependencies" section in your package.json.
```

Let's keep the current installation method for Linux then, until we find a better solution.

CodyCBakerPhD:
@garrettmflynn Could you submit a review with a 'requested change' somewhere here to make this that much harder (nigh impossible) to accidentally merge?

garrettmflynn (changes requested):
Adding this to block accidental merges. This PR should be kept separate for the convenience of Linux users.

CodyCBakerPhD:
replaced by #670

PR #524, opened 2023-11-21
Remove Global Pages

garrettmflynn:
This PR removes the initial Global Metadata pages in favor of pop-ups on an as-needed basis.

### DIFF
--- a/src/renderer/src/stories/pages/Page.js
+++ b/src/renderer/src/stories/pages/Page.js
@@ -170,7 +170,7 @@ export class Page extends LitElement {
         Object.assign(element.style, {
             textAlign: "left",
             display: "block",
-        })
+        });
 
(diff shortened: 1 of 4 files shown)


CodyCBakerPhD:
Finding a global metadata bug on this PR (unsure if it's technically on main too)

If I fill in multiple fields on the subject metadata table (now my first exposure in the pipeline to the notion of a 'global' value) they do not propagate down to the session level pages

(notably the subject species + age + sex, which is my usual workflow for speeding through the tutorial)

However, they _all_ show back up again if I just go to the 'Global Metadata' specification on the session-level page and hit 'update'

garrettmflynn:
Great catch. Fixed so that this works even though `globalState.project.Subject` is not reliably available.

CodyCBakerPhD (approved):

PR #525, opened 2023-11-21
Remove preview output location option

garrettmflynn:
Based on user feedback, this PR removes the option to change the preview output location.


CodyCBakerPhD:
LGTM - only bug found was mentioned on upstream branch #524 

In general I believe this should be fine as a general practice (storing previews under the GUIDE home directory)

Most stub files are ~KB-MB in scale and if it becomes a problem for someone (they raise an issue about it filling up their home folder) we can add a simple button in settings to clear it, or just document instructions for how to find the folder outside the App

CodyCBakerPhD (approved):

PR #526, opened 2023-11-23
Replace Desktop Build Infrastructure with CLI Tool

garrettmflynn:
This PR introduces a CLI tool I've been working on called [commoners](https://github.com/garrettmflynn/commoners) that would let us build applications like the NWB GUIDE on the same infrastructure without copying code into different repos.

With this PR, files that are not related directly to our frontend or Flask server (e.g. Electron main and preload processes that manage windows and servers) have been removed since their features are now handled by the CLI. While analogous code is still included in the final build, it's worth noting that the final `.dmg` for Mac is nearly half the size of the current `main` branch build (395 MB vs. 685 MB).

As I've been working on this tool on my own time, I've been motivated by how it might help CN and other collaborators produce apps like the GUIDE faster, without the pains we faced when determining the parts of SODA to include or remove. 

Using [commoners](https://github.com/garrettmflynn/commoners) for the GUIDE would be a great test case to run the CLI through its paces—though I also understand it may be out of scope. 

Let me know your thoughts!


garrettmflynn:
If it makes more sense, I can also roll back to before the code organization changes (https://github.com/NeurodataWithoutBorders/nwb-guide/pull/526/commits/c70390bf3efa7f21f1eb3a452d214bfe045ffd96) so the changelog is simplified.

CodyCBakerPhD:
@garrettmflynn Conflicts now

CodyCBakerPhD:
I do worry that doing this significant of a swap to core build structure could introduce bugs our limited testing suite will not detect

To assuage those concerns, I'd rigorously test some distributed releases (built by CI) - however, I can't run the actions for that on this branch since the deploy commands have changed (from `deploy:<platform>` to just `deploy`) so that's a bit of a chicken and egg problem

@garrettmflynn could you fork the repo, modify that workflow on your forks `main`, and then point me to the released versions built using this PR?

CodyCBakerPhD:
Also, what guarantees can you offer that this build structure will receive the same support that more centralized projects like `electron-builder` (which tons of people use) recieve?

Even if it's a true statement that this new package is better in some respects, it's a large undertaking to pin ourselves to that important of a dependency and so seems riskier to me long-term rather than letting a popular upstream package handle this kind of thing, especially a brand new one with only one contributor (even if it's you, who I trust quite a bit) and no peer-reviewed code (not that I'd be able to help review it; would want an expert on this kind of thing to do that)

In particular, this statement

> Using [commoners](https://github.com/garrettmflynn/commoners) for the GUIDE would be a great test case to run the CLI through its paces

Being the first major package to use a new build tool scares me; what if various major bugs are discovered after merging this PR?

Would we hold up GUIDE development/releases in order to fix those issues? How distracting would that be to your time?

Other questions...

Would it be easy to roll back to electron?

Would it be feasible to keep both build commands exposed in the `package.json` for easy fallback in case of failure in `commoners`?

Are there any changes to the GUIDE repo that are incompatible between the two build strategies?

garrettmflynn:
Sounds good, I'll prepare that fork as soon as we determine this is appropriate in the first place.

To address your concerns:

> Also, what guarantees can you offer that this build structure will receive the same support that more centralized projects like electron-builder (which tons of people use) recieve?

This tool uses `electron-builder` (API vs. CLI) internally and exposes the full configuration object on config.electron.build. As such, it should be thought of as *extending* rather than replacing such a tool.

At present, I haven't garnered any dedicated funding for `commoners`—though I'd like to through future projects at CN or other channels. It'd be up to discussions here at CN to determine if it's justified to let me charge for GUIDE-specific improvements to this infrastructure (e.g. integration costs such as this, fixing issues present in either the current `main` or the `commoners` version, etc.) if we move forward, although I see this as being useful for my own projects and don't intend to drop it anytime soon.

More than anything, `commoners` gives people looking at our project a unified conceptual structure to understand how it works the way it does (i.e. we're packaging our web app as a desktop application with a single Python service and a handful of Electron plugins to handle file association, etc.) and a predictable way to extend it in a way that allows for encapsulating related behaviors (e.g. add a service, plugin, etc.). While the current GUIDE repo is fairly organized, it isn't necessarily _simple_—or possible to simplify—because of the inherent challenges involved in configuring `electron`, `electron-builder`, and `electron-vite`, when the interesting parts of our code are actually located on the frontend and Python. 

Thinking about the future of this project where we might want to have additional developers/maintainers, I'd rather have the nitty-gritty details of deploying for desktop handled abstractly rather than having to sift through a large codebase that handles everything itself.

> Would it be easy to roll back to electron? Would it be feasible to keep both build commands exposed in the package.json for easy fallback in case of failure in commoners? Are there any changes to the GUIDE repo that are incompatible between the two build strategies?

Like `electron-builder`, `electron` is still used under the hood. `commoners` provides a default configuration (e.g. `main.js`, `preload.js`, `notarize.js`, etc.) that works for most use-cases, which can be extended through the `plugins` field of the configuration file.   I'd call these possible "changes to the GUIDE repo that are incompatible between the two build strategies" —although my goal is to remove these limitations, if found, so they aren't fundamental to the tool itself.

Because of the above, maintaining the `main.js` and related files that have been superceded would have to be done separately and redundantly.

> Being the first major package to use a new build tool scares me; what if various major bugs are discovered after merging this PR? Would we hold up GUIDE development/releases in order to fix those issues? How distracting would that be to your time?

Overall, our requirements for the Electron backend are fairly minimal, easily tested, and already encapsulated in the `commoners` plugin scheme (e.g. file dialogs, file association, launching the backend, etc.). As such, I don't anticipate too many fundamental issues between what we want and the behaviors of `commoners`.

If any issues are found, they'd hold across all apps created with this tool, making them highly important (to me personally) to fix. GUIDE development on related fixes / feature would be held up until the next version of the `commoners` library if they're integral to the library itself (which, as mentioned above, I'm trying to minimize)—though I'm not against a fast release schedule based on the needs of the GUIDE.

As a related note, we can lock the version of `commoners` to, say, `0.0.40` instead of `^0.0.40` on the GUIDE to avoid any possible issues introduced in later versions. 

<hr>

I know this is a lot of information, so let me know if it's helpful or hasn't directly addressed your initial questions.

garrettmflynn:
Yes, you could think of it as a wrapper for `electron` that provides a default configuration that works for most use-cases and simplified advanced features like the addition of arbitrary local backends.

> however, if commoners were to theoretically encounter a critical issue, we could fallback immediately to electron-builder

This, however, isn't exactly accurate. As mentioned right after the quoted portion, we'd have to maintain our current `main.ts` and related files independently in order to use `electron` + `electron-builder` in its raw form.

Though you can [configure `electron-builder` in the configuration file](https://github.com/NeurodataWithoutBorders/nwb-guide/blob/0d1d9da34c535ab510d97b2b38740c991261e292/commoners.config.ts#L120) if that's all you're worried about.

CodyCBakerPhD:
@garrettmflynn Are we good to close?

garrettmflynn:
Yeah looks like we'll stick with our current solution.

PR #530, opened 2023-11-28
Auto generate tutorial data

CodyCBakerPhD:
@garrettmflynn Need your help integrating this with front-end

In the meantime, I'll continue debugging and extend it to multi-subject, multi-session, and multi-probe


CodyCBakerPhD:
@garrettmflynn Might also need some help getting the test to trigger the endpoint properly, I keep getting a 415 error locally

garrettmflynn:
Getting an error about `ModuleNotFoundError: No module named 'sklearn'`. Are there additional dependencies here that weren't in the repo before?

CodyCBakerPhD:
Ahhhh, yes sorry - I should have added them to the conda environment files

Through trial and error it will be whatever is needed to succeed with these imports: https://github.com/NeurodataWithoutBorders/nwb-guide/pull/530/files#diff-bfe22e422f7a3669b04d7ddfc8b73491eda63907f0a4da680bb6494f98f15d69R771-R773

garrettmflynn:
Alright, I've updated this PR so that you can generate test data (and, at the same time, a dataset that replicates this data) on the Settings page. I've pre-emptively removed the Tutorial page since this replaces the data generation features there—though we don't currently have any textual material that replaces the original tutorial.

@CodyCBakerPhD Any suggestions before we can merge this in?

src/renderer/src/stories/pages/settings/SettingsPage.js
```diff
@@ -134,6 +141,42 @@ export class SettingsPage extends Page {
...
 
+    generateTestData = async () => {
+        await run(
+            "generate",
+            {
+                output_path: dataOutputPath,
+            },
+            {
+                title: "Generating test data",
+                html: "<small>This will take ~1min to complete.</small>",
```
CodyCBakerPhD, line 152:
```suggestion
                html: "<small>This will take several minutes to complete.</small>",
```

CodyCBakerPhD:
> @CodyCBakerPhD Any suggestions before we merge this in?

Only to have a navigation popup button next to the 'delete' one that leads the user to the generated dataset so that they can see what it looks like after it completes

CodyCBakerPhD:
Also, this PR removes autogeneration of the 'tutorial pipeline' in anticipation of the step-by-step screenshot approach approach, right?

CodyCBakerPhD:
Did you run through a full pipeline using the generated data?

I have an issue with the Phy folders specifically; they formed invalid symlinks for me

CodyCBakerPhD:
Specifically leading to 

```javascript
Uncaught (in promise) Error: ENOENT: no such file or directory, stat '~\NWB_GUIDE\test-data\data\phy\phy_example_0'
    at statSync (node:fs:1615:3)
    at Object.<anonymous> (node:electron/js2c/asar_bundle:2:4557)
    at Object.statSync (D:\GitHub\nwb-guide\node_modules\graceful-fs\polyfills.js:319:16)
    at #checkType (FileSystemSelector.js:134:38)
    at Array.forEach (<anonymous>)
    at #handleFiles (FileSystemSelector.js:142:53)
    at FilesystemSelector.selectFormat (FileSystemSelector.js:168:30)
```

in the folder selector

garrettmflynn:
That symlink issue should be fixed. My bad!

I've added that extra interaction as well. The "show in folder" command actually completely stalls my Finder on Mac, which is super weird. I'll see if I can find a way to avoid that

CodyCBakerPhD:
The delete command triggers a non-meaningful error when it tries to delete the target of the symlinks after the first time it was deleted (I assume that's what's going on); is there a way to 'safely' do that?

I also pushed a fix to the channel lengths on SpikeGLX, can you also run through a full test of a pipeline that uses these data?

garrettmflynn:
Can't get past the SourceData page with this error: 
Screenshot 2024-02-02 at 3 10 47 PM

garrettmflynn:
> The delete command triggers a non-meaningful error when it tries to delete the target of the symlinks after the first time it was deleted (I assume that's what's going on); is there a way to 'safely' do that?

Can you clarify what's triggering this error? Are you generating the dataset, deleting, regenerating, then deleting again? This is just removing the dataset output folder and shouldn't let you press the button without that folder existing.

garrettmflynn:
I've also found a fix for the lack of responsiveness when opening the folder location on Mac. Does this make things faster on your side as well?

garrettmflynn:
> Can't get past the SourceData page with this error: Screenshot 2024-02-02 at 3 10 47 PM

Fixed this. Turns out that if I don't delete the data (only the dataset), the previously generated files will not be overwritten?

pyflask/manageNeuroconv/manage_neuroconv.py
```diff
@@ -699,3 +698,106 @@ def inspect_multiple_filesystem_objects(paths):
...
+
+    base_path = Path(output_path)
+    spikeglx_output_folder = base_path / "spikeglx"
+    phy_output_folder = base_path / "phy"
+
+    recording, sorting = generate_ground_truth_recording(
+        durations=[5.0], sampling_frequency=30_000.0, num_channels=385, num_units=200, seed=0
+    )
+    artificial_lf_band = bandpass_filter(recording=recording, freq_min=300, freq_max=6000)
+    waveform_extractor = extract_waveforms(recording=recording, sorting=sorting, mode="memory")
```
CodyCBakerPhD, line 781: Moving discussion from Slack and looking for advice from @alejoe91

We're trying to approximate synthetic AP/LF bands and their associated spiking activity + waveforms (which get written as Phy)

(i) what filter should we use (using bandpass ATM; if so, what frequency range would you recommend) to approximate LF - also, how would you recommend downsampling?

(ii) for doing `recording.astype("int16")`, do you think I should do that before or after `extract_waveforms`?

CodyCBakerPhD:
> I've also found a fix for the lack of responsiveness when opening the folder location on Mac. Does this make things faster on your side as well?

Mine no longer even opens at all; if possible, could you use the previous method if its Windows and your current method if its Mac?



> Can you clarify what's triggering this error?

It was the kind of windows JS popup message that I can't copy/paste; said something about not being able to remove the resources because it didn't exist; I haven't seen it as of the last few commits

garrettmflynn:
Got it. I've ensured that the directories are only deleted if they exist, so that should solve the latter problem.

As for it not working on Windows, that's quite weird. Can you confirm that starting GUIDE again from the command line doesn't fix the issue? Since the changes were partially implemented on the main Electron process, the app needs to be restarted (which I also didn't notice and thought the system was broken after the initial implementation).

pyflask/manageNeuroconv/manage_neuroconv.py
CodyCBakerPhD: Moved to new issue #583 ; I've got something that at least looks not terrible for the moment

CodyCBakerPhD:
> As for it not working on Windows, that's quite weird. Can you confirm that starting GUIDE again from the command line doesn't fix the issue? Since the changes were partially implemented on the main Electron process, the app needs to be restarted (which I also didn't notice and thought the system was broken after the initial implementation).

That was it; fully restarting the app fixed the issue

CodyCBakerPhD:
@garrettmflynn This is now working well for me; can you review, try it completely, and approve if happy with it?

garrettmflynn (approved):
Works great! Let's get this in.

Issue #540, opened 2023-12-13
[Bug] Successful DANDI upload with multiple jobs spawns hanging processes


CodyCBakerPhD:
Another odd but reproducible issue on certain systems is that after a successful upload to the archive using the NWB GUIDE App, there are several rogue Python processes spawned in the background that take about ~5 minutes to timeout

In those 5 minutes, any attempt to close and relaunch the application will stall out. After the 5 minutes are up, the app can be used again without issue

The actual underlying issue remains a mystery (possibly swapping away from `joblib` as per previous discussions with Yarik may magically fix the problem?) but in the meantime there are two paths of action to resolve

(a) just wait 5 minutes; as long as this issue is easy to find when/if any user encounters the error, the problem will eventually fix itself

(b) forcing single job usage (not the current default as of 12/13/2023) also fixed the issue by not spawning the bad processes to begin with; @garrettmflynn suggested we patch this in as the new default just to be safe, which will also help to avoid those previous seg fault errors sporadically observed by users

CodyCBakerPhD:
Action item here is to just swap DANDI code to not use joblib

CodyCBakerPhD:
Will keep this in mind, but the current solution isn't too bad. Depending on system and internet, it's possible that a single job can move as quickly as multiple if the bandwidth is the bottleneck

PR #541, opened 2023-12-13
Make default upload parallel params 1

CodyCBakerPhD:
alleviates problem of #540 as well as joblib seg faults when running the GUIDE with default values, but doesn't actually solve underlying problem (which is unclear what the cause even is at this point...)

### DIFF
--- a/schemas/dandi-upload.schema.ts
+++ b/schemas/dandi-upload.schema.ts
@@ -14,7 +14,7 @@ Object.assign(idSchema, {
 })
 
 const additionalSettings = schema.properties.additional_settings.properties as any
-additionalSettings.number_of_jobs.description = additionalSettings.number_of_threads.description = '<b>⚠️</b> Setting this value has resulted in unpreditable behaviors (runaway processes, segmentation errors, etc.) on several systems</small>'
+additionalSettings.number_of_jobs.description = additionalSettings.number_of_threads.description = '<b>⚠️</b> Setting this value to more than 1 has resulted in unpredictable behaviors (runaway processes, segmentation faults, etc.) on several systems</small>'
 
 
 const setReady: any = {}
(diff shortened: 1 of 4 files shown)


garrettmflynn:
For this to truly take effect, we'd still want to (1) change the default on the Python end, as well as (2) stop populating the actual value of the form to the maximum value when the CPU information is returned from the server.

I'll handle this quick

garrettmflynn:
Should be good to go. Also added a caution message so users understand the consequences of bumping this value.

schemas/dandi-upload.schema.ts
```diff
@@ -6,12 +6,17 @@ import { isStaging } from '../src/renderer/src/stories/pages/uploads/utils'
...
 
+
 const schema = structuredClone(upload)
 const idSchema = schema.properties.dandiset as any
 Object.assign(idSchema, {
     strict: false
 })
 
+const additionalSettings = schema.properties.additional_settings.properties as any
+additionalSettings.number_of_jobs.description = additionalSettings.number_of_threads.description = '<b>⚠️</b> Setting this value has resulted in unpreditable behaviors (runaway processes, segmentation errors, etc.) on several systems</small>'
```
CodyCBakerPhD, line 17:
```suggestion
additionalSettings.number_of_jobs.description = additionalSettings.number_of_threads.description = '<b>⚠️</b> Setting this value to more than 1 has resulted in unpredictable behaviors (runaway processes, segmentation faults, etc.) on several systems</small>'
```

CodyCBakerPhD:
@garrettmflynn Small adjustment to displayed warning text, otherwise LGTM (but you need to approve since it's technically my PR lol)

garrettmflynn (approved):
Awesome. Thanks for updating.

PR #542, opened 2023-12-18
Ephys Metadata

garrettmflynn:
This PR re-enables Ephys metadata, integrating the changes to Ophys metadata (#505) and the last several months of updates.

To run this, make sure your environment is using the correct NeuroConv dev branch (https://github.com/catalystneuro/neuroconv/pull/447), which exposes the `data_type` property that corresponds to the expected return type of the table properties. Otherwise, all table properties will default to string types.


garrettmflynn:
@rly Just FYI tested this today and it looks like there's a blocker to convert the Ecephys metadata.

In addition, the relevant NeuroConv branch doesn't support OpenEphys yet

CodyCBakerPhD:
So, IDK. The more I think about this the more I think it might be best to just do a custom GUIDE-side operation on each recording interface similar to how the ophys metadata breaks up different planes. NeuroConv has a low-code method of attaching properties and doesn't really think of them as tables until they are in the NWB file, so trying to redesign that approach is going to be fairly difficult (especially with schematic validation since this isn't counted as 'metadata')

Probably best to do on a separate/fresh PR that runs on `main` of NeuroConv, but what I would do is generate a separate HandsOnTable for every recording interface that has columns for each property key (and rows for every channel) of the attached `recording_extractor` for each interface (there will always be exactly one extractor per interface, but could be multiple interfaces per converter)

CodyCBakerPhD:
As far as the table validation goes, that would then give us total control over the structure to enforce on the table, largely inferred from the `dtype` of the numpy array of that property (which they all should be I believe), which should make it easier to customize here

garrettmflynn:
@CodyCBakerPhD Would you mind reiterating the conclusions we reached in today's meeting so I can push forward with a new PR confidently?

CodyCBakerPhD:
@garrettmflynn separate hands on table for each recording interface; then a static non-interactive table that combines all of them to represent the way the actual NWB electrodes table will appear

Although, I think the neurosift preview after the metadata page should effectively handle that part; if that's OK with @rly then we can just have the individual tables exposed on the file metadata page and defer the 'unification' to the NeuroConv write tools + neurosift rendering on that small output

CodyCBakerPhD:
The tables will also need to support copy/pasting and the classic 'drag to copy/paste a repeated value' to be really usable if they don't do that already

CodyCBakerPhD:
replaced by #586

PR #547, opened 2023-12-28
Add group labels to sidebar

garrettmflynn:
Updated based on @oruebel User Test comments.

### DIFF
--- a/src/renderer/src/pages.js
+++ b/src/renderer/src/pages.js
@@ -104,6 +104,7 @@ const pages = {
     conversion: new GuidedHomePage({
         label: "Conversions",
         icon: guidedIcon,
+        group: 'Workflow',
         pages: {
             start: new GuidedStartPage({
                 label: "Start",
@@ -176,26 +177,32 @@ const pages = {
     inspect: new InspectPage({
         label: "Inspect",
         icon: inspectIcon,
+        group: 'Workflow'
     }),
     preview: new PreviewPage({
         label: "Neurosift",
         icon: neurosiftIcon,
+        group: 'Workflow',
     }),
     uploads: new UploadsPage({
         label: "Uploads",
         icon: uploadIcon,
+        group: 'Workflow'
     }),
     tutorial: new TutorialPage({
         label: "Tutorial",
         icon: tutorialIcon,
+        group: 'Documentation'
     }),
     docs: new DocumentationPage({
         label: "Documentation",
         icon: documentationIcon,
+        group: 'Documentation'
     }),
     contact: new ContactPage({
         label: "Contact Us",
         icon: contactIcon,
+        group: 'Documentation'
     }),
     settings: new SettingsPage({
         label: "Settings",
--- a/src/renderer/src/pages.js
+++ b/src/renderer/src/pages.js
@@ -104,7 +104,7 @@ const pages = {
     conversion: new GuidedHomePage({
         label: "Conversions",
         icon: guidedIcon,
-        group: 'Workflow',
+        group: "Workflow",
         pages: {
             start: new GuidedStartPage({
                 label: "Start",
@@ -177,32 +177,32 @@ const pages = {
     inspect: new InspectPage({
         label: "Inspect",
         icon: inspectIcon,
-        group: 'Workflow'
+        group: "Workflow",
     }),
     preview: new PreviewPage({
         label: "Neurosift",
         icon: neurosiftIcon,
-        group: 'Workflow',
+        group: "Workflow",
     }),
     uploads: new UploadsPage({
         label: "Uploads",
         icon: uploadIcon,
-        group: 'Workflow'
+        group: "Workflow",
     }),
     tutorial: new TutorialPage({
         label: "Tutorial",
         icon: tutorialIcon,
-        group: 'Documentation'
+        group: "Documentation",
     }),
     docs: new DocumentationPage({
         label: "Documentation",
         icon: documentationIcon,
-        group: 'Documentation'
+        group: "Documentation",
     }),
     contact: new ContactPage({
         label: "Contact Us",
         icon: contactIcon,
-        group: 'Documentation'
+        group: "Documentation",
     }),
     settings: new SettingsPage({
         label: "Settings",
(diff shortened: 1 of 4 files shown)


src/renderer/src/pages.js
```diff
@@ -176,26 +178,32 @@ const pages = {
...
     }),
     docs: new DocumentationPage({
         label: "Documentation",
         icon: documentationIcon,
+        group: "Documentation",
     }),
     contact: new ContactPage({
         label: "Contact Us",
         icon: contactIcon,
+        group: "Documentation",
```
CodyCBakerPhD, line 206:
```suggestion
        group: "Help",
    }),
    docs: new DocumentationPage({
        label: "Documentation",
        icon: documentationIcon,
        group: "Help",
    }),
    contact: new ContactPage({
        label: "Contact Us",
        icon: contactIcon,
        group: "Help",
```

Minor naming suggestion; on the new tab, it feels a little redundant to have a listing "Documentation/Documentation"

Or an alternative improvement would be to call the second documentation something else, but I have fewer ideas about that

CodyCBakerPhD:
The overall look is great and does help visually navigate that ever growing sidebar

My only major concern with this PR is the initial landing page when first launching the app - it's just a blank white screen - and a very long one at that (notice the size of the scroll bar in the screenshot below)

image

If you want to minimize information display initially presented to the user after launch, I'd at least add the GUIDE logo, or maybe a collage of all the logos, just so it doesn't feel so empty (and/or 'glitchy' - I think most users would expect _something_ to be in the main visual space of an app they just launched - a blank white space is reminiscent of that linux bug where it just fails to render anything that should be there)

If you don't care about minimizing initial display of info I'd say just start off on the first tab (Conversions) 

Also, having the 'settings' pinned as a kind of frozen tab leads to a weird visual effect where it looks like the 'Documentation' (which I'm above proposing rename to 'Help') group has only two tabs, showing the actual third only after scrolling, which by all appearances should have scrolled down past the 'settings'

So I'd propose putting 'Settings' under a group 'Configuration' there at the bottom

src/renderer/src/pages.js
rly: I agree about changing "Documentation" to something else. How about "Resources" or "Reference"?

garrettmflynn:
Alright, I've updated the labels for each of the groups and fixed the initialization so the home page is reached on first load.

src/renderer/src/pages.js
garrettmflynn: Just updated with latest push! Though I like "Resources" more than "Help". 

@CodyCBakerPhD What do you think?

rly:
The scrollbar not going down to Settings is good, but I think it is still visually confusing in the default smallish window size. The style is too similar to the menu items above it. And the "Configuration" header for a one item menu seems unnecessary. How about removing "Configuration" and either A) changing the background color of the Settings button or B) using a thicker, darker line to help distinguish the Settings button from the rest of the menu?

src/renderer/src/pages.js
CodyCBakerPhD: Sounds good

CodyCBakerPhD:
A darker background color would be sufficient for me

The extra 'Configuration' group is what I was imagining if and only if the group itself is becomes unpinned

But if you want to keep it pinned, then yeah, no need for a group label but does need a visual distinguisher from the other sections of the tab

garrettmflynn:
Got it. I've removed the header.

Since we don't use a darker gray elsewhere in the application, I've darkened the borders separating the main sidebar body from the header and footer. Is this sufficient?

rly:
It looks all right to me. Thanks for the adjustments. Is there a reason the default window size is the size that it is? If we don't plan to add new items to the menu, it would be nice if the default window height were large enough that a scroll bar in the menu is not necessary by default. Maybe 800px instead?

https://github.com/NeurodataWithoutBorders/nwb-guide/blob/main/src/main/main.ts#L234-L238 

I see that a 1366x768 screen resolution is [still fairly common](https://gs.statcounter.com/screen-resolution-stats/desktop/worldwide), but I would guess that it is rare among computers that would be working with converting and processing neurophysiology data.

CodyCBakerPhD:
Certainly looks better, but I'd also be in favor of Ryans suggestion if it is possible to force minimal size of window to be large enough to fit the 7 tabs of the app; maybe circle back around to the better solution when/if we ever add more workflows

On that note, a mild inconsistency I just noticed; one group is 'Workflow' (singular), other is 'Resources' (plural) - should we make it 'Workflows' to be consistent (and since there is more than one workflow)?

garrettmflynn:
Got it. Added those changes!

CodyCBakerPhD:
While the default window size now starts off as Ryan requested, I am able to manually shrink the window to the previous default size, which is just below where it needs to be to prevent this cutoff

If it's possible to constrain the absolute minimal size to above that cutoff that would be great, but let me know if that's too difficult to force

garrettmflynn:
If it doesn’t interfere with functionality, shouldn’t we provide the
flexibility to adjust the window size based on user requirements? While the
sidebar does render differently when smaller, it still functions as usual.

Otherwise I can set the minHeight rather than the height directly on the
browser window instance.

CodyCBakerPhD:
> If it doesn’t interfere with functionality, shouldn’t we provide the flexibility to adjust the window size based on user requirements?

I'm not saying freeze the entire window size - users can definitely adjust as they desire - simply bump the amount of 'smallest it can be' ever so slightly because the current minimum is just tantalizingly below what would be necessary to avoid the complexity of any scrolling on that subwindow altogether

garrettmflynn:
Got it, makes sense. I wasn’t aware how close the current minimum was to
our desired one.

I’ll make the change when I’m back at my laptop.

garrettmflynn:
Updated. Didn't catch that SODA initialized the browse window at the minimum size (since they had the same numbers declared twice). I've pulled the `height` and `minHeight` values out into a single variable (same for width) so it's more obvious and accomplishes your desired behavior.

CodyCBakerPhD:
LGTM

CodyCBakerPhD (approved):

PR #549, opened 2023-12-28
Automatic chromatic deployment on PR approval

CodyCBakerPhD:
Only running chromatic at the end of a PR (instead of every push) has really saved us a lot on monthly snapshots. Mildly annoying to have to trigger it manually, so trying out this trigger method on submission of an approved PR review


garrettmflynn (approved):
Great idea. Let's give it a go.

garrettmflynn:
@CodyCBakerPhD Looks like the workflow is invalid: https://github.com/NeurodataWithoutBorders/nwb-guide/actions/runs/7357710172

PR #452, opened 2023-10-16
Add VideoInterface

garrettmflynn:
This PR adds VideoInterface to the GUIDE, which passes on the test data for **single video files**.

However, we need to allow for specifying temporal alignment for multiple video files.

@CodyCBakerPhD How would you like to approach this?

PR #550, opened 2024-01-01
Allow Relative Video / Audio File Paths during NeuroConv Conversion

garrettmflynn:
This is a follow-up PR to #452 that will allow us to determine how to allow relative video / audio file paths without breaking the ability for NeuroConv to discover those files in the first place.


CodyCBakerPhD:
replaced by #833 which will simply suppress this inspector message in guided mode (and if anyone asks about it, standalone mode ought to be run in this case after the user does `dandi organize`)

PR #551, opened 2024-01-03
Remove abbreviations

garrettmflynn:
This PR removes abbreviations introduced on the `main` branch that don't provide enough clarity for developers without significant JavaScript experience.

There were all that I could find on a first pass. If additional abbreviations are identified or found to not be replaced here, we should make them explicit moving forward.

### DIFF
--- a/src/renderer/src/stories/Accordion.js
+++ b/src/renderer/src/stories/Accordion.js
@@ -196,19 +196,19 @@ export class Accordion extends LitElement {
     }
 
     setStatus = (status) => {
-        const el = this.shadowRoot.getElementById("dropdown");
-        el.classList.remove("error", "warning", "valid");
-        el.classList.add(status);
+        const dropdownElement = this.shadowRoot.getElementById("dropdown");
+        dropdownElement.classList.remove("error", "warning", "valid");
+        dropdownElement.classList.add(status);
         this.status = status;
     };
 
     onClick = () => {}; // Set by the user
 
-    #updateClass = (name, el, force) => {
-        if (force === undefined) el.classList.toggle(name);
+    #updateClass = (name, element, force) => {
+        if (force === undefined) element.classList.toggle(name);
         else {
-            if (force) el.classList.remove(name);
-            else el.classList.add(name);
+            if (force) element.classList.remove(name);
+            else element.classList.add(name);
         }
     };
 
--- a/src/renderer/src/stories/Accordion.js
+++ b/src/renderer/src/stories/Accordion.js
@@ -238,11 +238,7 @@ export class Accordion extends LitElement {
         else if (this.disabled) return false; // Force closed if disabled
         return state;
     };
-
-    updated() {
-        console.log(this)
-    }
-
+    
     render() {
         const isToggleable = this.content && this.toggleable;
 
--- a/src/renderer/src/stories/Accordion.js
+++ b/src/renderer/src/stories/Accordion.js
@@ -238,7 +238,7 @@ export class Accordion extends LitElement {
         else if (this.disabled) return false; // Force closed if disabled
         return state;
     };
-    
+
     render() {
         const isToggleable = this.content && this.toggleable;
 
--- a/src/renderer/src/stories/Accordion.js
+++ b/src/renderer/src/stories/Accordion.js
@@ -239,6 +239,10 @@ export class Accordion extends LitElement {
         return state;
     };
 
+    updated() {
+        console.log(this)
+    }
+
     render() {
         const isToggleable = this.content && this.toggleable;
 
--- a/src/renderer/src/stories/DandiResults.js
+++ b/src/renderer/src/stories/DandiResults.js
@@ -62,11 +62,11 @@ export class DandiResults extends LitElement {
                         const { message } = await fetch(
                             `http://api.crossref.org/works${new URL(identifier).pathname}`
                         ).then((res) => res.json());
-                        li.innerHTML = `${message.author.map(({ family, given }) => `${family}, ${given[0]}.`).join(", ")} (${
-                            message.created["date-parts"][0][0]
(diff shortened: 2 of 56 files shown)


src/renderer/src/stories/Accordion.js
```diff
@@ -196,19 +196,19 @@ export class Accordion extends LitElement {
     }
 
     setStatus = (status) => {
-        const el = this.shadowRoot.getElementById("dropdown");
-        el.classList.remove("error", "warning", "valid");
-        el.classList.add(status);
+        const dropdownElement = this.shadowRoot.getElementById("dropdown");
```
CodyCBakerPhD, line 199: Thanks for making this really explicit!

src/renderer/src/stories/DandiResults.js
```diff
@@ -52,21 +52,21 @@ export class DandiResults extends LitElement {
 
         const publicationEl = this.shadowRoot.querySelector(`.publication`);
         publicationEl.innerHTML = "";
-        const publications = (info.relatedResource ?? []).filter((o) => o.relation === "dcite:IsDescribedBy");
+        const publications = (info.relatedResource ?? []).filter(({ relation }) => relation === "dcite:IsDescribedBy");
```
CodyCBakerPhD, line 55: This one doesn't quite look equivalent syntactically - maybe it's OK though?

garrettmflynn: Yeah it's equivalent. The new version is destructuring the item and filtering by the `relation` property, *without* retaining a reference to the entire item in the `filter` function. Though `filter` returns the whole item if the function evaluates to `true` in either case.

CodyCBakerPhD: Oh I see now, fancy curly braces - neat

src/renderer/src/stories/InstanceManager.js
```diff
@@ -186,8 +186,8 @@ export class InstanceManager extends LitElement {
             const input = this.shadowRoot.querySelector("#new-info input");
             input.focus();
 
-            const mousePress = (e) => {
-                if (!e.composedPath().includes(newInfoDiv)) {
+            const mousePress = (mousePressEvent) => {
```
CodyCBakerPhD, line 189: What would you think about 'clickEvent' here?

Some implicit bias from neurodata experience led to a first reading as 'mouse (species) lever press', which are a type of common LabeledEvent

src/renderer/src/stories/InstanceManager.js
```diff
@@ -428,18 +430,22 @@ export class InstanceManager extends LitElement {
                     <div class="controls">
                         <span id="selectedName"></span>
                         <div>
-                            ${this.controls.map((o) => {
+                            ${this.controls.map(({ name, icon, primary, onClick }) => {
```
CodyCBakerPhD, line 433: Explicitly unpackaging contents really helps readability here, thanks!

src/renderer/src/stories/JSONSchemaForm.js
```diff
@@ -441,31 +441,10 @@ export class JSONSchemaForm extends LitElement {
...
-        } catch (e) {
-            this.throw(e.message);
+        } catch (error) {
+            this.throw(error.message);
         }
 
-        // NOTE: Ensure user is aware of any warnings before moving on
-        // const activeWarnings = Array.from(this.shadowRoot.querySelectorAll('.warnings')).map(input => Array.from(input.children)).filter(input => input.length)
-
-        // if (this.#nWarnings) {
```
CodyCBakerPhD, line 451: Yep, this is a good time to clean any commented code out as well

src/renderer/src/stories/Search.js
```diff
@@ -371,7 +371,7 @@ export class Search extends LitElement {
 
                     return li;
                 })
-                .filter((el) => el);
+                .filter((itemEl) => itemEl);
```
CodyCBakerPhD, line 374: `itemElement`? If that is any more informative than just `item` or just `element`

src/renderer/src/stories/SimpleTable.js
```diff
@@ -209,7 +209,7 @@ export class SimpleTable extends LitElement {
         this.addEventListener("copy", (ev) => {
             ev.preventDefault();
             const tsv = Object.values(this.#selected)
-                .map((arr) => arr.map((el) => el.value).join("\t"))
+                .map((arr) => arr.map((inputEl) => inputEl.value).join("\t"))
```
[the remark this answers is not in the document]
CodyCBakerPhD: or just `input`?

CodyCBakerPhD:
Some minor comments scattered across; otherwise looks good. Will give thorough hands on test after those are addressed so ensure nothing is unintentionally broken, otherwise will be good to go

Thanks for taking the time to do this

src/renderer/src/stories/InstanceManager.js
garrettmflynn: Switching to pointerEvent since that's also a type of event in JS.

src/renderer/src/stories/SimpleTable.js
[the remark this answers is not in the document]
garrettmflynn: Argh thanks for catching this. We'll just go with `inputElement`.

src/renderer/src/stories/Search.js
garrettmflynn: It's a `<li>` (i.e. list item) element. I see where you're coming from, HTML probably should've chosen a less ambiguous word lol.

Is `itemElement` fine or would you prefer `listItemElement`?

CodyCBakerPhD: Oh my, if you're saying it's an element of a `LI` -> `ListItem`, not shorthand for `list` itself, then yeah the full `listItemElement` would be the most concise

CodyCBakerPhD (approved):

PR #556, opened 2024-01-08
Try collect all on h5py

CodyCBakerPhD:
Attempting to fix the failing CI

PR #557, opened 2024-01-08
Update PyInstaller to 6.3 to fix missing imports in packaging

### LINKED ISSUE #559: PyFlask failures - January 2024
Test suite applied to built versions of the app on Mac (intel) and ubuntu have been failing for past few days

After analyzing differences between [passing](https://github.com/NeurodataWithoutBorders/nwb-guide/actions/runs/7399504355/job/20131089010) to [failing](https://github.com/NeurodataWithoutBorders/nwb-guide/actions/runs/7409596543/job/20160174711) environments as of 5 days ago, the only difference is boto3 and botocore (the only package we use that cut a new release on Jan 5)

Pinning those in https://github.com/NeurodataWithoutBorders/nwb-guide/actions/runs/7452039831/job/20274447966?pr=558 did not fix the issue however

Core error has something to do with the mysterious `.defs` clib missing from https://github.com/h5py/h5py/blob/master/h5py/h5.pyx#L14C2-L14C21 of `h5py`

Attempts to collect-all on `h5py` and `h5py.defs` have failed: https://github.com/NeurodataWithoutBorders/nwb-guide/pull/557 & https://github.com/NeurodataWithoutBorders/nwb-guide/pull/556

rly:
Fix #559, closes #556, closes #558, closes #560

### DIFF
--- a/environments/environment-Linux.yml
+++ b/environments/environment-Linux.yml
@@ -4,9 +4,9 @@ channels:
   - defaults
 dependencies:
   - python = 3.9.18
-  - PyInstaller = 5.13.0
+  - PyInstaller = 6.3.0
   # pin pyinstaller hooks as temp fix for https://github.com/NeurodataWithoutBorders/nwb-guide/issues/559
-  - pyinstaller-hooks-contrib = 2023.11
+  # - pyinstaller-hooks-contrib = 2023.11
   - nodejs = 18.16.1
   - numcodecs = 0.11.0
   # install these from conda-forge so that dependent packages get included in the distributable
--- a/environments/environment-Linux.yml
+++ b/environments/environment-Linux.yml
@@ -5,6 +5,7 @@ channels:
 dependencies:
   - python = 3.9.18
   - PyInstaller = 5.13.0
+  - pyinstaller-hooks-contrib = 2023.11
   - nodejs = 18.16.1
   - numcodecs = 0.11.0
   # install these from conda-forge so that dependent packages get included in the distributable
--- a/environments/environment-Linux.yml
+++ b/environments/environment-Linux.yml
@@ -5,8 +5,6 @@ channels:
 dependencies:
   - python = 3.9.18
   - PyInstaller = 6.3.0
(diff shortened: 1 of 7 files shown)


environments/environment-Linux.yml
```diff
@@ -4,7 +4,9 @@ channels:
   - defaults
 dependencies:
   - python = 3.9.18
-  - PyInstaller = 5.13.0
+  - PyInstaller = 6.3.0
+  # pin pyinstaller hooks as temp fix for https://github.com/NeurodataWithoutBorders/nwb-guide/issues/559
+  # - pyinstaller-hooks-contrib = 2023.11
```
CodyCBakerPhD, line 9: Looks like 6.3.0 removes the need for the hooks-contrib for Linux ; do you think we should try that for the other platforms?

Also occurred to me that we don't even use the PyInstaller build on Linux (which requires dev mode install anyway0 so we should probably remove that os from the test matrix

CodyCBakerPhD:
@rly CI is passing now! Let me know when you're done cleaning up anything and ready for review

Thanks for tackling this so quickly

environments/environment-Linux.yml
rly: Yeah, it looks like an incompatibility between PyInstaller 5.13.0 and pyinstaller-hooks-contrib 2023.12. I'll make the changes for the other platforms

rly: Yeah I think we should also remove Linux from the test matrix

rly:
This PR now updates PyInstaller from 5.13.0 to the latest version, 6.3.0. This version includes a few breaking changes: https://pyinstaller.org/en/stable/CHANGES.html#id9 .  I don't see anything obvious that would affect GUIDE. @CodyCBakerPhD @garrettmflynn please review these changes to PyInstaller to make sure nothing might break because of them.

rly:
I also removed testing of the PyFlask build and distributable on ubuntu and removed that as a required test for merging PRs.

CodyCBakerPhD (approved):

PR #558, opened 2024-01-08
Try pinning boto

CodyCBakerPhD:
After analyzing differences between [passing](https://github.com/NeurodataWithoutBorders/nwb-guide/actions/runs/7399504355/job/20131089010) to [failing](https://github.com/NeurodataWithoutBorders/nwb-guide/actions/runs/7409596543/job/20160174711) environments as of 5 days ago, the only difference is `boto3` and `botocore` (the only package we use that cut a new release on Jan 5)

Who knows why that would trigger this specific issue, so perhaps a long shot, but trying here


CodyCBakerPhD:
replaced by #557

PR #561, opened 2024-01-11
Fix Bugs with New Application Structure

garrettmflynn:
This PR fixes some bugs introduced in #555 as a result of the new URL structure of the GUIDE. 

Essentially, the correct trigger is used to detect when the user has left their pipeline, and saved Section information is replaced with new path keys so duplicate buttons do not appear on the sidebar of old pipelines.


CodyCBakerPhD:
@garrettmflynn Looks like backend-tests are failing

garrettmflynn:
Got it. Thanks for pinging me

CodyCBakerPhD (approved):

PR #563, opened 2024-01-16
Allow users to generate pipelines from YAML files

### LINKED ISSUE #511: Generate a conversion pipeline for each interface
@CodyCBakerPhD We recently suggested the need to programmatically define the generation of arbitrary numbers of test pipelines—specifically one per interface.

We agreed on a JSON structure that will define the pipeline name, interfaces, and source data—essentially everything that would be included in the pipeline itself but only for one subject.

```json
{
    "pipeline_name": {
        "interfaces": {
            "spikeglx": "SpikeGLXRecordingInterface"
        },
        "source_data": {
            "spikeglx": {
                "file_path": "ephy_testing_data/spikeglx/Noise4Sam_g0/Noise4Sam_g0_imec0/Noise4Sam_g0_t0.imec0.ap.bin"
            }
        }
    }
}
```

garrettmflynn:
This PR allows for users to generate pipelines from a YAML file on the Settings page. Currently, enough metadata is pre-filled to simply click through the conversion without any interactions.

This is the first step to fix #511, where we can expand the test YAML file (`guide_testing_suite.yml`) to account for all interfaces.

Since receiving the test YAML file from @CodyCBakerPhD, I've removed the `base_data_path` property to improve the shareability of the file itself and fixed the `PhySortingInterface` folder path to target the correct subdirectory.


CodyCBakerPhD:
@garrettmflynn This particular feature is not intended to be the generic 'import from YAML' to be exposed to users; that feature would use the NeuroConv specific YAML structure (different from what I shared), and is largely dependent on the custom backend feature to be useful (and possibly other aspects, too) hence why it's not even on the current release roadmap

This particular feature was only supposed to result in a single button on the settings page for developers to point to their local copy of the GIN repos and auto-generate all pre-established pipelines defined in that YAML file (which will grow over time; each entry essentially being a test case)

garrettmflynn:
Got it. Constrained the behavior to load the provided file rather than allow the selection of a custom YAML file.

CodyCBakerPhD:
1) Are the generated pipelines working on your end? @garrettmflynn 

2) Minor, but would it be possible to visually join the button for genreation with the folder path selector to indicate they are related?

image

Perhaps under a 'Developer' testing header like the 'DANDI' one

garrettmflynn:
> Are the generated pipelines working on your end? @garrettmflynn

Ah I see. Phy always was, but the SpikeGLX single-probe example had the wrong interface specified (should be `SpikeGLXRecordingInterface` not `SpikeGLXDataInterface`).

garrettmflynn:
> Minor, but would it be possible to visually join the button for genreation with the folder path selector to indicate they are related?

Updated with this new grouping!

CodyCBakerPhD:
Delightful

CodyCBakerPhD (approved):

CodyCBakerPhD:
Awww, Chromatic build failure: https://github.com/NeurodataWithoutBorders/nwb-guide/actions/runs/7628567110/job/20779934786?pr=563

CodyCBakerPhD (approved):

PR #569, opened 2024-01-22
Download inspector report

garrettmflynn:
This PR allows users to download the NWB Inspector report as either JSON or text after viewing the initial GUIDE version.


CodyCBakerPhD:
This is great!

My one and only qualm is the difference in button text between the pages

Standalone

image

PIpeline

image

garrettmflynn:
Assuming you'd prefer it all written out? In smaller screen sizes, things can get a bit crowded—but that's fine for the sake of clarity.

CodyCBakerPhD:
> Assuming you'd prefer it all written out? In smaller screen sizes, things can get a bit crowded—but that's fine for the sake of clarity.

I think both look fine, it's more about having same style for both - so fine to go with shorter version if you prefer

CodyCBakerPhD (approved):

PR #573, opened 2024-01-29
Alow datetime to have additional information than allowed by the input

### LINKED ISSUE #570: ScanImage session start time display
Generate a fresh ScanImage SinglePlane SingleChannel testing pipeline and navigate to the file metadata page

The session start time field will be correctly yellow (it is found within the ScanImage header) but the value is not displayed in the datetime selector, making it seem as if it is unknown

image

garrettmflynn:
fix #570


CodyCBakerPhD (approved):

CodyCBakerPhD:
Seems to work, will keep an eye out for this from now on

PR #580, opened 2024-02-01
End-to-End Tests on Generated Pipelines using Puppeteer

garrettmflynn:
This PR will implement an end-to-end testing suite that runs through generated and manual pipelines using Puppeteer.

@CodyCBakerPhD Before we can push forward on this reliably, we have to get the GIN testing data downloaded in the Github Actions workflow. I noticed how you're doing this in NeuroConv, but it wasn't clear how to specify where the datasets are downloaded to. For this case, I've made it an assumption that the E2E tests must have the GIN data at `~/NWB_GUIDE/test-data`, as it isn't clear how to pass this path dynamically using `vitest`.

Can you clarify how we'd install the GIN data at this path for Github Actions to use?


CodyCBakerPhD:
> Can you clarify how we'd install the GIN data at this path for Github Actions to use?

Most reliable way would probably be to `cd` to the path you want to save them for `vitetest` to find them

Though, is this just for the tutorial data? Did you want to add cases from the general data based testing suite?

If tutorial only, might be worth finishing up #218 to simplify the issue altogether

garrettmflynn:
You mean this PR? https://github.com/NeurodataWithoutBorders/nwb-guide/pull/530

garrettmflynn:
Either one is fine. Depends on how comprehensive we want to be (e.g. we could quite easily run through all the YAML pipelines automatically).

It would be great if all the tests worked for any user without any configuration (i.e. we have the data included somehow), but there are ways to seamlessly transition between the two states.

CodyCBakerPhD:
Ah, yes, #530

CodyCBakerPhD:
> It would be great if all the tests worked for any user without any configuration (i.e. we have the data included somehow)

That's pretty tough to do in a way that scales well (even in actions we utilize caching) and is really interoperable

Simplest approach is for us to require the data to exist, and simplest approach for data to exist is to use tools specifically built for data transfer

garrettmflynn:
#530 still requires that the testing data is on the user's computer, so we'll still have to install on the CI in a default location.

You're right, though, we will want users to be interacting with a multi-subject, multi-session dataset for the tutorial—so it would be nice to have this merged. I'll use this for the manual pipeline.

garrettmflynn:
> That's pretty tough to do in a way that scales well (even in actions we utilize caching) and is really interoperable

Got it, thank you for the clarification. I'll use the approach you've defined above.

CodyCBakerPhD:
> still requires that the testing data is on the user's computer

What do you mean by 'testing data'? That PR does not require the GIN example data, it generates synthetic files, if those are the ones you're talking about. But since we're the ones making them we have full control where they get saved

garrettmflynn:
Got it. Changed the name of this PR and will keep it in draft mode.

I'll work on #530 until merged, then continue to build out #582.

CodyCBakerPhD:
@garrettmflynn Is coverage possible to track via e2e tests? That could reduce amount of other tests we need to add to increase it

CodyCBakerPhD:
@garrettmflynn Has this been replaced?

garrettmflynn:
Technically no—though we have pretty much decided not to pursue this.

This particular PR prepares the E2E tests to run through _all_ of the generated test pipelines. Something outside the scope of the tutorials, but still possibly useful.

Do we want to approach this or just close this out?

CodyCBakerPhD:
I see

Hm...

I'm not against slower CI for the sake of more thorough testing, but to do so for every generated pipeline may be considered excessive...

Perhaps we just add an ophys one, and then resolve to add new ones for particular generated pipelines that are able to reproduce any bugs that get reported on a rolling basis?

CodyCBakerPhD:
Although, another thing I assume we could do is have a full battery of e2e tests defined, but only run them on say a daily CRON basis so they don't slow down or block PRs, but we still get regular feedback on every generated testing pipeline

garrettmflynn:
Yeah I was thinking this too—even just to have the full suite available for a local test.

Seems like both approaches are useful and on the roadmap. In that case, we should keep this open and I'll adapt this PR for the discussed use-cases.

garrettmflynn:
The general workflow is complete, though there are now a few additional issues that have been uncovered. In summary:
1. Suite2P + Caiman: Fluorescence data is not handled correctly
     - Likely just have to assert that additional properties are okay for this. Also must fix some rendering changes made recently
2. Scanimage: `ValueError: Frame index (0) exceeds number of frames (0)` (full trace below)
3. TDT: Executes fine, but I'd noticed that the Electrodes are not updated after first run if you change the gain—likely because of how arrays are merged and prior values cached

garrettmflynn:
Temporarily fixing ScanImage by switching to `ophys_testing_data/imaging_datasets/ScanImage/sample_scanimage_version_3_8.tiff` instead of `scanimage_20220801_single.tif`

garrettmflynn:
@CodyCBakerPhD So it also looks like tests are failing because both the original E2E tests (now `tutorial.tests.ts`) and the new `pipelines.tests.ts` file rely on an instance of the Electron app opening and this isn't coordinated enough for them both to work.

I can spend some time teasing out the problem and making sure we can run these using the general test command—but I'd thought I'd ask about the general testing approach we'd like to move forward with first.

Locally, I'm running these tests independently and have created new scripts in the `package.json` to handle this (`npm run test:pipelines` and `npm run test:tutorial`). Would we want to break these out into different CI workflows or different steps in the same workflow? Or should we continue optimizing for a single call to `npm run test`?

CodyCBakerPhD:
Sounds like different CI workflows is best then? Most modular and parallel approach

garrettmflynn:
@CodyCBakerPhD Should be working now! Will approach the TDT issue separately

CodyCBakerPhD:
@garrettmflynn CI does not like something in the dependencies ATM - on any platform it seems

garrettmflynn:
Thanks for flagging! Should be fixed.

Also was able to knock out the TDT issue here. Turns out it was more related than I thought

src/renderer/src/stories/JSONSchemaForm.js
```diff
@@ -513,6 +513,14 @@ export class JSONSchemaForm extends LitElement {
                 const resolvedSchema = e.schema; // Get offending schema
 
                 // ------------ Exclude Certain Errors ------------
+                // Allow using null to specify an immutable table
+                if (
+                    e.message.includes("does not meet maximum length") ||
+                    e.message.includes("oes not meet minimum length")
```
CodyCBakerPhD, line 519:
```suggestion
                    e.message.includes("does not meet minimum length")
```

Should this read like this?

Also what kind of element is `e` here?

src/renderer/src/stories/pages/guided-mode/data/GuidedMetadata.js
```diff
@@ -252,14 +252,12 @@ export class GuidedMetadataPage extends ManagedPage {
                     if (path[0] === "Ophys") {
                         const name = path.slice(-1)[0];
 
-                        if (isPatternProperties)
-                            return (schema.minItems = schema.maxItems =
-                                Object.values(resolveFromPath(path, results)).length);
+                        if (isPatternProperties) schema.minItems = schema.maxItems = null; // Do not allow more than on the reuslts
```
CodyCBakerPhD, line 255:
```suggestion
                        if (isPatternProperties) schema.minItems = schema.maxItems = null; // Do not allow more than on the results
```

tests/e2e/pipelines.test.ts
```diff
@@ -0,0 +1,120 @@
...
+            const folderInput = page.form.getFormElement(["developer", "testing_data_folder"])
+            folderInput.updateData(testGINPath)
+
+            const button = folderInput.nextSibling
+            await button.onClick()
+
+            page.save()
+        }, testGINPath)
+
+        // Transiton back to conversions page and count pipelines
```
CodyCBakerPhD, line 58:
```suggestion
        // Transition back to the conversions page and count pipelines
```

CodyCBakerPhD:
Looked over everything, just one small question about the form of an error message and the rest looks good to go!

src/renderer/src/stories/JSONSchemaForm.js
garrettmflynn: Yes thanks for catching this!

`e` is one of the error objects reported by the JSON Schema validation function. This syntax is fairly common for error handing in JavaScript, but we can change `e` to something more descriptive like `jsonSchemaError` if you prefer.

CodyCBakerPhD: Thanks, I'll try changing those on my own soon then

CodyCBakerPhD (approved):

CodyCBakerPhD:
@garrettmflynn Thanks for the hard work on this, this is gonna be great for maintainability!

Issue #588, opened 2024-02-06
Show more readable display names for interfaces


rly:
The interface selector uses cryptic names that look like code:
<img width="1385" alt="Screenshot 2024-02-05 at 10 04 52 PM" src="https://github.com/NeurodataWithoutBorders/nwb-guide/assets/310197/fcc4df2b-02ad-4fe8-9704-a2dd491c48d5">

Ideally, each NeuroConv interface has a class variable `display_name` that is a nice user-friendly display name, e.g., "AlphaOmegaRecordingInterface" -> "AlphaOmega Recording", that can be read by NWB GUIDE and shown in this interface selector and the selected interfaces list. @bendichter is working on adding that class var to each interface.

In the meantime (until the next neuroconv release), we could improve the names with a few simple rules, like remove the word "Interface" and add spaces before certain words. I tried to do this just for kicks and will submit a PR.

CodyCBakerPhD:
@rly I think this has improved - can you check if this can be closed?

PR #589, opened 2024-02-06
Make interface display names more human readable

rly:
Temporary solution for #588 until next NeuroConv release with the display names as class variables for each interface. cc @bendichter 

We don't need to merge this if the NeuroConv release schedule is relatively fast.

Result:
Screenshot 2024-02-05 at 11 44 00 PM

After adding an interface, the key is still the name of the interface:
image

I wasn't sure what the downstream effects of changing that are so I left that as is.


CodyCBakerPhD:
> Temporary solution for https://github.com/NeurodataWithoutBorders/nwb-guide/issues/588 until next NeuroConv release with the display names as class variables for each interface

We don't need a release of NeuroConv to make it work; NWB GUIDE packages the `main` branch

CodyCBakerPhD:
I'm not sure what the "(All Data)" part adds; what are you trying to say with that?

CodyCBakerPhD:
I wouldn't do this on the TypeScript side; I'd modify the values returned by the NeuroConv query on the backend since that's a lot more customizable: https://github.com/NeurodataWithoutBorders/nwb-guide/blob/interface_display_name/pyflask/manageNeuroconv/manage_neuroconv.py#L165-L170

You can also set it up there to retrieve the 'display_name' if set as an attribute on the data interface classes

Did you want to work on that @rly?

garrettmflynn:
Updated to align with NeuroConv and the rest of the GUIDE repo, as `displayName` should've just been used in `GuidedStructure.js` as the `key` value, rather than replacing `key` in the Search and List components.

Note that several converters and interfaces need to be named in NeuroConv still:
Screenshot 2024-02-08 at 2 08 55 PM
Screenshot 2024-02-08 at 2 10 16 PM
Screenshot 2024-02-08 at 2 10 22 PM

CodyCBakerPhD:
LGTM except for failing tests

Fine with getting this through as is and adding more display names on NeuroConv over time

garrettmflynn:
Thanks for flagging that. Tests are passing now!

garrettmflynn:
Just updated so that the `info` attribute is used preferentially to `__doc__`. Also enabled searching by suffix. 

Holding off on the use of `associated_suffixes` in the Data Formats page for a follow-up. That is the desired target location for both the informational and programmatic use of this attribute, right?

CodyCBakerPhD:
> Just updated so that the info attribute is used preferentially to __doc__.

Yeah I thought about just defering to the __doc__ but the downside is our class docstrings have a lot of .rst and sphinx formatting for the autogeneration in the API section. So the `info` is intended to be a shorter, text-only version of that.

> That is the desired target location for both the informational and programmatic use of this attribute, right?

The NeuroConv PR currently has an attribute on every class called `associated_suffixes` which is a tuple of strings. If that PR is merged as is what would be the target location for retrieving the information

CodyCBakerPhD:
@garrettmflynn an e2e conflict on this one

CodyCBakerPhD:
@garrettmflynn https://github.com/catalystneuro/neuroconv/pull/734 is mostly ready to go now; shall we do a pass and see if this is ready for review?

CodyCBakerPhD:
With that PR you can use a new function to get what you need

`get_format_summaries` imported by `from neuroconv import get_format_summaries`

garrettmflynn:
Updated! However, for using `get_format_summaries` to make much sense, I've declared the results as a global variable. Otherwise, I was grabbing the values on an as-needed basis directly from the interface.

CodyCBakerPhD:
@garrettmflynn Still a merge conflict with the e2e tests

garrettmflynn:
Ah missed the merge conflict. Was just trying to nail the test failure down

garrettmflynn:
@CodyCBakerPhD All of these tests are now going to fail because of the reliance on the NeuroConv branch. Previously, I was just using fallbacks if things didn't exist—but that new function can't be handled the same way. 

How do you want to mediate that in the meantime?

garrettmflynn:
I'm fine with the latter if you're fine merging with tests failing 😅

CodyCBakerPhD:
Oh no, not what I meant lol 😆 this should either not be merged until upstream is, or pin to NeuroConv dev branch it's compatible with and immediately update the pins after that merge

garrettmflynn:
Ah gotcha lol

pyflask/manageNeuroconv/manage_neuroconv.py
```diff
@@ -15,9 +15,14 @@
 from sse import MessageAnnouncer
 from .info import GUIDE_ROOT_FOLDER, STUB_SAVE_FOLDER_PATH, CONVERSION_SAVE_FOLDER_PATH
 
+from neuroconv import get_format_summaries
```
CodyCBakerPhD, line 18: Remember, no global imports

garrettmflynn: Noted the reason earlier in [this comment](https://github.com/NeurodataWithoutBorders/nwb-guide/pull/589#issuecomment-1947404689). I'm using `get_format_summaries` globally here, though. If I don't do that, it doesn't make sense to use it at all—as all the interface info can be grabbed from the interface itself when used.

Would reverting to that method be fine, or would you prefer using the helper?

garrettmflynn: Ah I see what you're updated. If you're fine with this "hydration" of a global, that works for me too.

Though wouldn't we want to keep the result persistent so we don't regenerate every time?

CodyCBakerPhD: > Though wouldn't we want to keep the result persistent so we don't regenerate every time?

IDK what you mean

CodyCBakerPhD: Because the global imports are for minimal flask operations. Import of neuroconv and any helpers thereof occurs during `startup` command

What I did is exactly what you had, but safer in that you can now launch the Flask server without importing neuroconv at the same time

If you'd rather just define the global and call the function without the `manageNeuroconv` intermediate, that's perfeclty fine by me, I was just minimally fixing what was already there on this PR

Refer back to that old chain of PRs that implemented all that back when we refactored the PyInstaller stuff (last summer?)

garrettmflynn: Ah, I see. The commit/file I was referring to did not show this was being called in `startup`. 

Thanks for your patience 😅

Reverting to manually grabbing the attributes now.

CodyCBakerPhD:
Yep, this looks great now

Just need a better search functionality in a follow-up to find things like "phy" or "sorted" (as in, "spike sorted" currently only finds "sorting")

CodyCBakerPhD (approved):

PR #614, opened 2024-02-13
Restore automation trigger

CodyCBakerPhD:
It seems the problem with chromatic baselines may actually have been due to the squashing rather than full merge of commits that I've been using on recent PR merges, rather than from the trigger condition

Enabling the full app according to Chromatic instructions appears to have fixed it


garrettmflynn (approved):
Looks good! Thanks

PR #612, opened 2024-02-12
Update issue templates

garrettmflynn:
fix #164

PR #617, opened 2024-02-19
Retry on failure

### LINKED ISSUE #613: E2E Tests are Inconsistent
The new end-to-end testing workflow spontaneously fails on a subset of the CI platforms for no predictable reason. 

E.g. https://github.com/NeurodataWithoutBorders/nwb-guide/actions/runs/7879423399/attempts/1?pr=612 failed on Mac because of a timeout issue on the Subject page, whereas Windows failed because of the inability to locate a certain DANDI asset (whose URL appeared to have an extra colon at the end).

**To Reproduce**
Steps that leads to the issue:

1. Create a new branch from `main`
2. Submit a PR
3. Observe the Actions workflow, specifically the backend tests
4. Repeat backend tests until failure

garrettmflynn:
fix #613

This was suggested when merging #612

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -63,6 +63,9 @@ jobs:
         name: Create and activate environment
         run: mamba env update -n nwb-guide -f ${{ matrix.label }}
 
+      - name: Display conda environment
+        run: conda list
+
       - name: Use Node.js 18
         uses: actions/setup-node@v3
         with:
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -87,7 +87,7 @@ jobs:
           timeout_minutes: 30
           max_attempts: 3
           retry_on: error
-          shell: bash
+          shell: bash {0}
           command: npm run test:coverage
 
       - if: matrix.os == 'ubuntu-latest'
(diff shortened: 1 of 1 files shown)


CodyCBakerPhD:
OK, well the action is working: https://github.com/NeurodataWithoutBorders/nwb-guide/actions/runs/7960795322/job/21737344076?pr=617#step:12:575 -> https://github.com/NeurodataWithoutBorders/nwb-guide/actions/runs/7960795322/job/21737344076?pr=617#step:12:1138

But not helping  alleviate the underlying problem...

garrettmflynn:
That specific run has an issue because of `No module named 'flask'`. Seems like the create and activate environment step didn't run.

I've never noticed this become a problem before. Is there a reason this was excluded?

CodyCBakerPhD:
Hmm... Could it be that the action isn't running in the same shell? (one in which the conda environment is not active?)

Adding a printout of the installed packages to be sure; the cache was restored in the logs I linked so you'd hope so

garrettmflynn:
Ah looks like that's the case. Just updated to the same shell.

Fingers crossed!

garrettmflynn:
Hmm all attempts here don't seem to help. According to the docs, this should default to `bash` on Mac and Linux—and otherwise it seems likely that it'd use the default set at the top of the workflow.

CodyCBakerPhD:
Right, but specifying it as `bash` would be a fresh shell I believe, not the currently active conda environment; maybe add a `conda activate nwb-guide` as first step of the `run` command in the retry?

CodyCBakerPhD:
Tabling for now to see if we actually need this - #621 has been working well so far

CodyCBakerPhD:
Yeah I don't think we actually need this any more

PR #620, opened 2024-02-20
Update search strategy

### LINKED ISSUE #615: Update Search Method
Currently, we are searching for partial matches within the name and keywords properties _anywhere in the string_, something that results in lots of false positives (e.g. searching `phy` returns all interfaces with the `electrophysiology` keyword in addition to the `PhySortingInterface`.

Ideally, searches would be approached in the following manner:
1. Break all strings into their component words
2. Start at the beginning of each word
3. Check if matching
4. Allow non-alphanumeric characters (e.g. `.jpg` to be matched with `jpg`)

garrettmflynn:
fix #615

### DIFF
--- a/src/renderer/src/stories/Search.js
+++ b/src/renderer/src/stories/Search.js
@@ -256,27 +256,31 @@ export class Search extends LitElement {
 
     #sortedCategories = [];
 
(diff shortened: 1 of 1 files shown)


CodyCBakerPhD:
Definitely a bit better! I know search can get very complicated, so without getting to into the weeds of doing faceted search via some external reference database of terms this ought to be good enough for now

CodyCBakerPhD (approved):

PR #623, opened 2024-02-21
Add option to set DANDI_CACHE environment variable to "ignore"

### LINKED ISSUE #619: [Feature]: Add 'ignore cache' as advanced option for DANDI upload

Along with the parallel params in the advanced DANDI upload section, there is also an environment variable mentioned in https://github.com/dandi/dandi-cli/issues/1321#issuecomment-1942757013 `DANDI_CACHE` which can be set to `ignore` to fix certain issues

Would it be possible to add a check box on the GUIDE that then sets that env var to that value if checks and wipes it if not checked?

Yes.

### Code of Conduct

Yes

Yes

garrettmflynn:
fix #619

### DIFF
--- a/pyflask/manageNeuroconv/manage_neuroconv.py
+++ b/pyflask/manageNeuroconv/manage_neuroconv.py
@@ -582,10 +582,17 @@ def upload_folder_to_dandi(
     cleanup: Optional[bool] = None,
     number_of_jobs: Optional[int] = None,
     number_of_threads: Optional[int] = None,
+    ignore_cache: Optional[bool] = None,
 ):
     from neuroconv.tools.data_transfers import automatic_dandi_upload
 
     os.environ["DANDI_API_KEY"] = api_key  # Update API Key
+
+    if ignore_cache:
+        os.environ["DANDI_CACHE"] = "ignore"
+    else:
+        os.environ.pop("DANDI_CACHE", None)
+
 
     return automatic_dandi_upload(
         dandiset_id=dandiset_id,
--- a/pyflask/manageNeuroconv/manage_neuroconv.py
+++ b/pyflask/manageNeuroconv/manage_neuroconv.py
@@ -593,7 +593,6 @@ def upload_folder_to_dandi(
     else:
         os.environ.pop("DANDI_CACHE", None)
 
-
     return automatic_dandi_upload(
         dandiset_id=dandiset_id,
         nwb_folder_path=Path(nwb_folder_path),
--- a/pyflask/manageNeuroconv/manage_neuroconv.py
+++ b/pyflask/manageNeuroconv/manage_neuroconv.py
@@ -616,7 +616,7 @@ def upload_project_to_dandi(
     if ignore_cache:
         os.environ["DANDI_CACHE"] = 'ignore'
     else:
-        os.environ.pop("DANDI_CACHE")
+        os.environ.pop("DANDI_CACHE", None)
 
     return automatic_dandi_upload(
         dandiset_id=dandiset_id,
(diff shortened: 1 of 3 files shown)


garrettmflynn:
@CodyCBakerPhD Hmmm this seems new?

```
Server requires schema version 0.6.6; client only supports 0.6.4.  You may need to upgrade dandi and/or dandischema.
```

pyflask/manageNeuroconv/manage_neuroconv.py
```diff
@@ -582,11 +582,17 @@ def upload_folder_to_dandi(
...
+    ignore_cache: Optional[bool] = None,
 ):
     from neuroconv.tools.data_transfers import automatic_dandi_upload
 
     os.environ["DANDI_API_KEY"] = api_key  # Update API Key
 
+    if ignore_cache:
+        os.environ["DANDI_CACHE"] = "ignore"
+    else:
+        os.environ.pop("DANDI_CACHE", None)
```
CodyCBakerPhD, line 594: Tracing all the way down to the cacher they use, https://github.com/con/fscacher/blob/6ccd52f1845c02cce99cbb4b849c0debf2cdc76a/src/fscacher/cache.py#L23

The proper way I think would be to set this to just an empty string. That's without getting complicated with actually modifying the existence status of the environment variable; i.e., `pop` removes it from the mapping exposed to this Python process but doesn't unset it in the parent

```suggestion
    else:
        os.environ["DANDI_CACHE"] = ""
```

schemas/json/dandi/upload.json
```diff
@@ -20,9 +20,15 @@
                     "min": 1,
                     "default": 1
                 },
+                "ignore_cache": {
+                    "type": "boolean",
+                    "description": "Ignore the cache and re-upload all files.",
```
CodyCBakerPhD, line 25:
```suggestion
                    "description": "Ignore the cache used by DANDI to speed up repeated operations.",
```

Keeping it more generic since I don't actually know every single way they interact with their cache. All I know is it supposedly speeds up certain things. Suspect it has to do mostly with expensive checksum calculations

schemas/json/dandi/upload.json
```diff
@@ -20,9 +20,15 @@
...
+                    "default": false
+                },
                 "cleanup": {
                     "type": "boolean",
-                    "title": "Delete Local Files After Upload",
+                    "title": "Cleanup Local Filesystem",
+                    "description": "Delete Local Files After Upload",
```
CodyCBakerPhD, line 31:
```suggestion
                    "description": "Delete local files after upload",
```

For consistent style

pyflask/manageNeuroconv/manage_neuroconv.py
```diff
@@ -605,13 +611,19 @@ def upload_project_to_dandi(
...
     from neuroconv.tools.data_transfers import automatic_dandi_upload
 
     # CONVERSION_SAVE_FOLDER_PATH.mkdir(exist_ok=True, parents=True)  # Ensure base directory exists
 
     os.environ["DANDI_API_KEY"] = api_key  # Update API Key
 
+    if ignore_cache:
+        os.environ["DANDI_CACHE"] = "ignore"
+    else:
+        os.environ.pop("DANDI_CACHE", None)
```
CodyCBakerPhD, line 625:
```suggestion
        os.environ["DANDI_CACHE"] = ""
```

pyflask/manageNeuroconv/manage_neuroconv.py
```diff
@@ -582,11 +582,17 @@ def upload_folder_to_dandi(
     cleanup: Optional[bool] = None,
     number_of_jobs: Optional[int] = None,
     number_of_threads: Optional[int] = None,
+    ignore_cache: Optional[bool] = None,
```
CodyCBakerPhD, line 585:
```suggestion
    ignore_cache: bool = False,
```

No big reason to have `None` in this case, right?  Since it maps from a checkbox, which is either marked or not

pyflask/manageNeuroconv/manage_neuroconv.py
```diff
@@ -605,13 +611,19 @@ def upload_project_to_dandi(
...
```
CodyCBakerPhD, line 614:
```suggestion
    ignore_cache: bool = False,
```

PR #626, opened 2024-02-27
Update installation instructions

rly:
- Replace hyperlink text "NWB-GUIDE-Setup-vX.Y.Z.exe" (and similar) with the correct filename.
- Add example code to the linux instructions.
- Change Ubuntu -> Linux. It works on Fedora as well. I think we can generalize to all Linux systems and change this if someone reports it not working on their system.
- Update Dev instructions to make clear that activating the env is necessary regardless of which OS you use (currently it looks like it applies only to Linux)


garrettmflynn (approved):
Good catch. Thank you!

CodyCBakerPhD (approved):

PR #627, opened 2024-02-27
Skip DANDI upload to allow tests to pass with broken external services

garrettmflynn:
Skip the DANDI upload part of the E2E tests to avoid failing when external dependencies (e.g. the DANDI server) has an issue.

### DIFF
--- a/.github/workflows/testing-external.yml
+++ b/.github/workflows/testing-external.yml
@@ -1,0 +1,93 @@
+name: External Tests
+on:
+  schedule:
+    - cron: "0 16 * * *"  # Daily at noon EST
+  pull_request:
+
+concurrency:  # Cancel previous workflows on the same pull request
+  group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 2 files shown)


CodyCBakerPhD:
What I'm requesting is that we split the current testing action into two actions; one that runs only the live service tests like DANDI upload, and one that runs only the non-live service tests that we can control

These would show up as two separate CI runs on a PR, and yes the live one would continue to fail until DANDI fixes their end

garrettmflynn:
Sure. Though since this DANDI upload test appears at the very end of the E2E workflow and requires a complete pipeline to finish, would it be sufficient to simply duplicate the `testing.yml` + keep the `DANDI_STAGING_API` that was removed here? 

We could cache a complete pipeline instead, though that'll take a bit longer.

garrettmflynn:
Alright added that extra testing file

CodyCBakerPhD (approved):

PR #629, opened 2024-02-28
Fix Nested Form Cell Error Display

garrettmflynn:
This PR fixes a small error where the nested form cell error is not actually shown.


CodyCBakerPhD:
@garrettmflynn Any way to write a test for this?

CodyCBakerPhD (approved):

Issue #554, opened 2024-01-04
File Metadata Pre-Input Form


garrettmflynn:
As discussed with @CodyCBakerPhD, it would streamline the user workflow to have a popup pre-form for the File Metadata page (which automatically opens on first visit) that allows users to answer a series of simple yes/no questions to filter the possible interactions available in the rest of the form.

As a concrete example, we discussed a sub-series of questions that included "Do you know the Emission Lambda values of your Optical Channels?", "Are all Emission Lambda values the same?", and finally "What is the value of Emission Lambda?".

Paired with #553, this would allow us to populate global values as a result of the user's answer to these questions and/or hide interactions with related inputs so there is less to focus on for the rest of the File Metadata page.

CodyCBakerPhD:
Since this app was so inspired to be the TurboTax for NWB conversions, I'll be noting all kinds of things whilst filing this years taxes

TurboTax uses pre-input forms like this prolifically to simplify the main data entry portions

rly:
I very much agree, especially for NWBFile metadata. Something like:

If the institution name is the same for all of your sessions, what is the institution name?
[Text box]
[Button "Next"] [Button "Let me enter for each session later"]

Same for experimenter name, keywords, related publications, etc.

Same for subject species and sex and maybe some of the other subject ones, but definitely those two.

CodyCBakerPhD:
closed by #633 

We'll continue to refine over time

PR #633, opened 2024-03-06
Pre-Form Basic Implementation

garrettmflynn:
This PR is a first pass at the pre-form page discussed in #554, with relevant interactions across the GUIDE workflow.

This will help constrain the tutorial workflow so users are only presented with the UI they will actually use for a given section.


CodyCBakerPhD:
Instead of checkboxes, can they be yes/no radio selectors? More personable since we're asking questions (again, TurboTax-like)

CodyCBakerPhD:
Other than that, it sure feels really good to not even have to see the locate data page if I say no

Also nice that is shows up grayed out in the main progress tab so I know it 'would' have been there otherwise

garrettmflynn:
> Instead of checkboxes, can they be yes/no radio selectors? More personable since we're asking questions (again, TurboTax-like)

This will be applied globally with https://github.com/NeurodataWithoutBorders/nwb-guide/pull/640

garrettmflynn:
Whoops I forgot to update the tests after skipping the Subjects page. 

This should be ready to review now.

PR #654, opened 2024-03-11
List Fix

### LINKED ISSUE #649: Renamed interfaces revert back to old name
### Describe the issue

When you add a new data format any previously renamed formats will revert back to their default name

### Steps to Reproduce

https://github.com/NeurodataWithoutBorders/nwb-guide/assets/844306/381de37d-ffb5-4d98-b26e-812acafc8911

See the name of the first format in this video. It is renamed to have a 1 at the end and then reverts to its initial name when a new format is added.

### Operating System

osx

### GUIDE Version

main branch

### Code of Conduct

Yes

Yes

garrettmflynn:
fix #649

### DIFF
--- a/src/renderer/src/stories/List.ts
+++ b/src/renderer/src/stories/List.ts
@@ -205,7 +205,7 @@ export class List extends LitElement {
       this.items.splice(draggedIdx, 1)
       this.items.splice(i, 0, movedItem)
 
-      this.items = [...this.items]
+      this.items = this.items
     }
 
 
@@ -225,7 +225,8 @@ export class List extends LitElement {
     }
 
     add = (item: ListItemType) => {
-      this.items = [...this.items, item]
+      this.items.push(item) // Update original
+      this.items = this.items
     }
 
     #removePlaceholder = () => {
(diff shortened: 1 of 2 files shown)


bendichter:
https://github.com/NeurodataWithoutBorders/nwb-guide/assets/844306/2f65d577-5b52-4ad0-a916-4c2eace6f83c

Two new problems:
1. Any new added interface copies the renamed interface instead of starting with the default name.
2. When you select and then deselect the name field, it adds "_1"

garrettmflynn:
Thank you for thoroughly testing these behaviors! I've fixed these behaviors by copying the item as it is added to the list—as it seems the reference used by the Search component was being modified in the above case.

bendichter:
https://github.com/NeurodataWithoutBorders/nwb-guide/assets/844306/b8feb81e-c0ac-4f12-9435-07718edb6284

OK, those problems are fixed, but now we have a new problem: If you add a 3rd interface of the same type without changing any names, the entire app freezes

garrettmflynn:
Missed an infinite loop in there because of an improperly named variable (`i` when it should have been `kI`) 🫠

I'm updating my environment for the latest `neuroconv` version, but AFAIK this should fix this behavior.

bendichter (approved):

CodyCBakerPhD (approved):

PR #659, opened 2024-03-11
Interactive electrodes

garrettmflynn:
Re-enable interactivity for ElectrodesTable to diagnose ongoing issues with on large tables.

If you want to avoid the deferred table rendering, simply comment out the following line in the GuidedMetadata.js file:
```javascript
if (name === "Electrodes") metadata.deferLoading = true
```


CodyCBakerPhD:
Dang, yeah. It's stalling out

PR #662, opened 2024-03-12
Remove spec CLI

CodyCBakerPhD:
@rly @garrettmflynn What do you all think of just removing these? We've had to do so much manual modification of the spec file it is much easier IMO to just continue to directly modify it instead of using this hack


package.json
```diff
@@ -20,7 +20,6 @@
     "build:mac": "npm run build && npm run build:flask && npm run build:electron:mac",
     "build:linux": "npm run build && npm run build:flask && npm run build:electron:linux",
     "build:flask": "python -m PyInstaller nwb-guide.spec --log-level DEBUG --clean --noconfirm --distpath ./build/flask",
-    "build:flask:spec:base": "pyi-makespec --name nwb-guide --onedir --collect-data jsonschema_specifications --collect-all dandi --collect-all keyrings --collect-all unittest --collect-all nwbinspector --collect-all neuroconv --collect-all pynwb --collect-all hdmf --collect-all hdmf_zarr --collect-all ndx_dandi_icephys --collect-all sklearn --collect-all ci_info ./pyflask/app.py",
     "build:flask:spec": "npm run build:flask:spec:base && python prepare_pyinstaller_spec.py",
```
rly, line 23:
```suggestion
```
Remove this as well?

rly:
I agree - it would be easier to modify `nwb-guide.spec` directly

garrettmflynn (approved):
That's fine with me. Let's do it.

PR #663, opened 2024-03-12
Data Persistence following Subject Table Interactions

garrettmflynn:
While using the Subject Table with the generate test datasets, it became somewhat annoying to lose the associated source data + metadata results (forever...) when changing the subject name.

To address this, this PR improves the persistence of data when you update the subject table:
1. Renaming a single subject will result in a persistent link with existing data
2. Creating a new row and swapping `subject_id` values _and_ maintaining the same `session_id` will result in data persistence on the _same_ `subject_id`
     - Without the same `session_id`, the data will persist on the _new_ `subject_id`

I'm open to suggestions regarding further rules, changes to these, and any bugs you notice!


CodyCBakerPhD:
image

If I change the subject ID of a later row to one of a row above, then delete the row above, I get this (the invalidation is 'subject id already exists')

garrettmflynn:
Good catch. Thanks!

I've made it so that the table revalidates when any `subject_id` is resolved from an error state.

CodyCBakerPhD:
@garrettmflynn Conflict now

CodyCBakerPhD (approved):

CodyCBakerPhD:
@garrettmflynn Failing test now

CodyCBakerPhD (approved):

PR #679, opened 2024-03-15
Render Units Table

garrettmflynn:
This PR allows you to view the units table. Currently, editing is disabled for all related properties to  simplify the user interactions with the table to editing unit properties _outside the GUIDE_.


CodyCBakerPhD:
Attempting this on `dataset_3` of `cellexplorer` yields

```
  File "D:\anaconda3\envs\nwb-guide\lib\site-packages\flask\app.py", line 1484, in full_dispatch_request
    rv = self.dispatch_request()
  File "D:\anaconda3\envs\nwb-guide\lib\site-packages\flask\app.py", line 1469, in dispatch_request
  File "D:\anaconda3\envs\nwb-guide\lib\site-packages\flask_restx\api.py", line 404, in wrapper
  File "D:\anaconda3\envs\nwb-guide\lib\site-packages\flask\views.py", line 109, in view
  File "D:\anaconda3\envs\nwb-guide\lib\site-packages\flask_restx\resource.py", line 46, in dispatch_request
    resp = meth(*args, **kwargs)
  File "D:\GitHub\nwb-guide\pyflask\apis\neuroconv.py", line 97, in post
    return get_metadata_schema(neuroconv_api.payload.get("source_data"), neuroconv_api.payload.get("interfaces"))
  File "D:\GitHub\nwb-guide\pyflask\manageNeuroconv\manage_neuroconv.py", line 361, in get_metadata_schema
    schema["properties"]["Ecephys"]["required"].append("Electrodes")
KeyError: 'required'
```

CodyCBakerPhD:
Default description for `brain_area` should replace 'electrode' with 'unit'

UnitColumns description should be editable so I can change it from "No description"

Since we're not allowing direct editing of table content, can disallow 'data type' from unit column table

The more I think about it the more I think this section should just be for control over unit table column descriptions; in your opinion, is it even useful to render the values of those columns on the page if they can't be easily controlled? This is really more of a behavior relegated to the 'preview' pages

garrettmflynn:
IMO it might still be useful to see the data to provide an appropriate column description—otherwise I think you're right in framing the purpose of the Units information this way.

garrettmflynn:
Updated! 

BTW I realized now that we're never setting the new descriptions in NeuroConv. How would I actually do this?

garrettmflynn:
How should we reconcile the fact that there could be multiple ElectrodeColumns/UnitColumns tables across different interfaces, each with their own values for this?

garrettmflynn:
(2) is going to be the least confusing to implement and interact with, but would also result in the inability to add new interface-specific columns.

Is this alright? Otherwise, we could also declare an additional property (e.g. `interfaces`) that indicates the interface(s) that any particular column applies to—and which defaults to all if blank.

garrettmflynn:
Starting to work with both a global table and subtables will contain the same challenges as (1) involving some reconciliation of duplicated columns.

CodyCBakerPhD:
> some reconciliation of duplicated columns.

How so?

To add a new column I would add the name, description, and data type to the global column 'header' table

Then to add new column values on a single electrode table I would just add it to that particular table

For validation, you would then check that every electrode table column exists on the 'header' table - with the other approach, you would still check if that column exists, just not on the global 'header' table, instead the local 'header' table (as it is on `main` now)

garrettmflynn:
Ah I see. So previously I'd automatically add the new column to the particular interface's Electrodes table—and since these new values would all be blank, this would register as invalid. As such, users were given a clear indication that they had missing data.

If we're comfortable removing that signal, then a global ElectrodeColumns table would only be used to validate whether an extra column *is not described*, though we wouldn't be able to provide feedback on whether a particular column is not found on an Electrodes table that the user expects it to be in—as our expectation that a new column should be present at all is governed by its existence on the table in the first place.

garrettmflynn:
Updated both the Units and Electrodes tables to refer to global columns tables!

CodyCBakerPhD:
I would place the `ElectrodeGroup` section after `Device` (and rename `Device` to `Devices` plural; ditto `ElectrodeGroups`)

Overall I think this looks and feels better

Also these might be problems with SI side but we'd still be responsible for parsing them better... The Units table for the Phy on the tutorial data has invalidations for `Name` columns (integers, should be parsed as strings) and KSLabels and KSLabel Rpeat (might only show up in SI version `0.100.2`) show up as 'objects' instead of strings (or maybe they are meant to be a list of strings for more than one label per unit, even though in tutorial example each only has one element

garrettmflynn:
Fixed a few things, but I'm having issues setting data back to the Units.

Every property gives me the following error:
```
'0' is not in list
```

Where 0 is the unit name.

Any idea how to get around this?

CodyCBakerPhD:
> Where 0 is the unit name.

Sounds like it's acting as if unit IDs still behave like `int` rather than `str` , so `'0' not in list` might be true but maybe `0 in list`? Or in other words, acting based off of indices not string identifiers

garrettmflynn:
Ah duh. Thanks!

pyflask/manageNeuroconv/manage_neuroconv.py
```diff
@@ -987,6 +1141,98 @@ def get_recording_interface_properties(recording_interface) -> Dict[str, Any]:
...
+
+    unit_ids = sorting.get_unit_ids()
+
+    table = list()
+    for unit_id in unit_ids:
+
+        unit_column = dict()
+
+        for property_name in properties:
+            if property_name is "name":
```
CodyCBakerPhD, line 1218:
```suggestion
            if property_name == "name":
```

CodyCBakerPhD:
Still having the problem with KSLabel on Phy interface - also all units tables seem to have a 'Quality' column that is all empty on the unit table, with no way to change it from that invalid status

garrettmflynn:
Huh I'm not seeing that. Maybe I'm not using the same data?

Can you point me directly to the source data you're using?

CodyCBakerPhD:
At stated above, it's the tutorial data

You might need SI 0.100.2 to see it

garrettmflynn:
Huh still no luck. Updated my environment and regenerated the tutorial data.

Here's my confirmation that I'm running the correct version of SpikeInterface
Screenshot 2024-03-25 at 9 30 25 AM

CodyCBakerPhD:
Ahhh my bad, it's the actual example phy data not the tutorial

CodyCBakerPhD:
In the autogenerated pipeline 'SpikeGLX-Phy'

garrettmflynn:
Gotcha. Fixed!

CodyCBakerPhD:
OK thanks, that looks better now

The unit column table doesn't seem to be saving any of my changes to the descriptions though; even if I hit 'save' at the top, if I proceed to the next page then come back it restored to the default. Any ideas?

garrettmflynn:
Weird I don't have the same problem. 

Screenshot 2024-03-25 at 1 23 33 PM

I've updated the previous "name" column to "unit_name" to see if that description would update—but it seems like I'm doing that wrong or that isn't possible.

CodyCBakerPhD:
@garrettmflynn Small conflict here

garrettmflynn:
Still haven't gotten around to those PC-specific issues with the description that we noticed—but the rest should be reviewable.

CodyCBakerPhD:
Are you able to reproduce them locally?

Also, failing tests on this branch

garrettmflynn:
Running into some issues on my PC. Updated the environment and ensured that I have the same testing data as my Mac, but this error is being thrown when going off the Source Data page of the SpikeGLX-Phy test pipeline.
Screenshot 2024-03-28 142220

CodyCBakerPhD:
versions of neo and SI?

CodyCBakerPhD:
Otherwise maybe redownload the files from S3 or otherwise make sure you're pointing to the right files. The nSavedChannels is definitely there

garrettmflynn:
spikeinterface is at 0.100.3, while neo is 0.13.0. I'll re-download the data and try again, as this is using the test pipeline without modifications (which works fine elsewhere) and manually re-specifying the source data didn't work.

It's also quite odd because the Exclude Cluster Groups option for Phy is back to being improperly specified and flagged with that yellow warning.

Screenshot 2024-03-28 163107

garrettmflynn:
Redownloading fixed the Source Data issue. But for some reason the Ecephys metadata is excluded from the schema returned from the GUIDE server, and the returned Ecephys appears to be unprocessed (i.e. before we updated to include electrode data, even).

I'm on the `units-table` branch and have swapped several times. The server runs appropriately after running `npm start`.

Just sharing because it's such weird behavior and I have no idea where to start to debug. If you have any thoughts, let me know.

If nothing comes to mind, though, don't worry about it. I'll slog through it as soon as I can.

CodyCBakerPhD:
> Just sharing because it's such weird behavior and I have no idea where to start to debug. If you have any thoughts, let me know.

In these events I don't just do a complete environment wipe but also a complete local `git` wipe

IDK if related, but above I did find it weird to see you using the classic command shell instead of Anaconda prompt

garrettmflynn:
Okay I will try that.

Oddly, running NPM start changes how my Anaconda prompt tab renders to appear like a classic command prompt. Though I tested again this morning and ensured that I used the Anaconda prompt, but this does not change the behavior.

garrettmflynn:
The full Git wipe worked! Thanks for sticking with me on this lol.

Definitely less acquainted with common Windows issues than I should be.

CodyCBakerPhD:
I too did a complete wipe (including NWB_GUIDE home directory)

The original invalidation issues are indeed gone now, but made a video of the description thing

https://github.com/NeurodataWithoutBorders/nwb-guide/assets/51133164/28e11673-b6b3-49e3-b2a6-b09868c8ac91


It's pretty minor though, I'm OK with tabling as a low priority thing to figure out in a followup if it's too hard to pin down

CodyCBakerPhD (approved):

PR #674, opened 2024-03-14
Fix nested input rendering

garrettmflynn:
This PR fixes a small rendering issue with the nested JSONSchemaInput properties (e.g. `file_paths` for the VideoInterface), which convert lists of only one element to single inputs.

PR #680, opened 2024-03-15
Do not skip on pages on storybook

garrettmflynn:
May have some relevance to the rendering change noticed on #674


CodyCBakerPhD:
I also just disabled squashing, which may have been messing it up too

CodyCBakerPhD (approved):

PR #686, opened 2024-03-18
Add json encoding to fstring autocompletion

### LINKED ISSUE #672: Error with datetime object in Locate Data, Linux version
### Describe the issue

When adding a folder with date and f-string for getting the date at the Locate Data screen, the API endpoint /locate fails. It seems to be because the DeepDict object stores the date as a datetime object which can't be serialized to JSON format.

image

### Steps to Reproduce

1. Create a new conversation pipeline
2. Check both 'Would you like to locate these sessions programmatically' - Yes
3. Video - VideoInterface format
4. Choose base directory and string path
5. When it tries to automatically list the files, it fails

### Operating System

Ubuntu 20.04.6 LTS

### GUIDE Version

linux-latest

### Code of Conduct

Yes

Yes

CodyCBakerPhD:
fix #672

### DIFF
--- a/pyflask/apis/neuroconv.py
+++ b/pyflask/apis/neuroconv.py
@@ -68,7 +68,7 @@ class Schemas(Resource):
 
 
 @neuroconv_api.route("/locate")
-class Locate(Resource):
+class LocateData(Resource):
     @neuroconv_api.doc(responses={200: "Success", 400: "Bad Request", 500: "Internal server error"})
     def post(self):
         try:
@@ -79,7 +79,7 @@ class Locate(Resource):
 
 
 @neuroconv_api.route("/locate/autocomplete")
-class Locate(Resource):
+class AutoCompleteFormatString(Resource):
     @neuroconv_api.doc(responses={200: "Success", 400: "Bad Request", 500: "Internal server error"})
     def post(self):
         try:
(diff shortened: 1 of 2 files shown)


garrettmflynn:
Assuming you confirmed that this allows for proper interaction with the test-case defined in the related Issue?

CodyCBakerPhD:
@garrettmflynn I'm testing it now - it certainly resolves the JSON serialization issue but then triggers other issues - can you give it a look and you probably know better what needs to interact properly

Note that for SpikeGLX prior to latest commit I was able to proceed all the way through and it found the data correctly however I could not test that the session start time makes it through to the final stage because the SpikeGLX session time takes precedence over the f-string one

So I tried again with a copy of the tutorial data renamed along the lines of `..../dataset/mouse2/mouse2_Session2_210324/mouse2_Session2_210324_phy` (since Phy doesn't have session start time in the source data) but now it's having trouble both auto identifying locations (gives 'undefined' as the format string after 'succeeding' autocomplete) and manually specifying it makes a weird subject metadata table

garrettmflynn:
Updated. Whenever you're encoding into JSON format using a custom encoder, you'll have to reload to an object using `json.loads` to ensure the object is passed appropriately to the frontend (instead of double-encoded and received as a JSON string after JSON parsing).

CodyCBakerPhD:
@garrettmflynn Working great for me!

garrettmflynn (approved):
Perfect. Let's get this through

PR #688, opened 2024-03-18
Hide multi-instance sidebar when only a single session

### LINKED ISSUE #685: [Feature]: Single session - Source data and File Metadata

image

For this page, since there is only one session, there is no need for the "Sessions" menu. I think this should be removed, and we should just have the "Phy Sorting" and "SpikeGLX Converter" accordions. The same logic applies to the "File Metadata" page.

Yes.

### Code of Conduct

Yes

Yes

garrettmflynn:
fix #685 

Since the Instance Managers on the source data and file metadata pages have their header with internal controls, I've decided to implement a solution between the suggestions of @oruebel and @bendichter that hides the multi-instance sidebar while maintaining the general structure of the Instance Manager. 

Screenshot 2024-03-18 at 9 40 20 AM

### DIFF
--- a/src/renderer/src/stories/InstanceManager.js
+++ b/src/renderer/src/stories/InstanceManager.js
@@ -51,6 +51,10 @@ export class InstanceManager extends LitElement {
                 justify-content: space-between;
             }
 
+            #instance-sidebar[hidden] {
+                border: unset;
+            }
+
             #instance-header {
                 background: #e5e5e5;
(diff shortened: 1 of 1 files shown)


bendichter:
see https://github.com/NeurodataWithoutBorders/nwb-guide/issues/689, which is related

bendichter (approved):

CodyCBakerPhD (approved):

PR #695, opened 2024-03-20
Inspect progress bars with a global handler

garrettmflynn:
This PR implements what was discussed in https://github.com/catalystneuro/tqdm_publisher/pull/46 where the `MessageAnnouncer` class used by the GUIDE is superceded by a queue manager provided by `tqdm_publisher`.

This update allows utilities that ingest the handler (e.g. NWB Inspector, DANDI, etc.) to simple expect a dictionary of `additional_metadata` to pass alongside the update provided by `TQDMProgressSubscriber`.

Uses https://github.com/catalystneuro/tqdm_publisher/pull/49


pyflask/manageNeuroconv/manage_neuroconv.py
```diff
@@ -685,10 +684,10 @@ def upload_project_to_dandi(
 
 # Create an events endpoint
 def listen_to_neuroconv_events():
-    messages = announcer.listen()  # returns a queue.Queue
+    messages = progress_handler.listen()  # returns a queue.Queue
     while True:
```
CodyCBakerPhD, line 688: Where is `listen_to_neuroconv_events` called from in the rest of the code?

CodyCBakerPhD:
In general looks much simpler for the GUIDE side, and I like the abstraction

Some comments on the other repo about final refinement

pyflask/manageNeuroconv/manage_neuroconv.py
garrettmflynn: https://github.com/NeurodataWithoutBorders/nwb-guide/blob/309727948ac8fe222a73ab7c9bda92902682edd9/pyflask/apis/neuroconv.py#L286

This is where it gets defined as an event stream for an EventSource on the browser.

CodyCBakerPhD: Great, and can you point me to an example of where/how that endpoint gets used in a particular case?

garrettmflynn: So the connection is established here: https://github.com/NeurodataWithoutBorders/nwb-guide/blob/309727948ac8fe222a73ab7c9bda92902682edd9/src/renderer/src/stories/utils/progress.js#L89

My intention is for all progress bar popups to be instantiated using `createProgressPopup` above this: https://github.com/NeurodataWithoutBorders/nwb-guide/blob/309727948ac8fe222a73ab7c9bda92902682edd9/src/renderer/src/stories/utils/progress.js#L6

This automatically manages the complicated bits of ensuring that the SSE messages apply to the appropriate request.

garrettmflynn:
This feature is implemented without this PR. Though this particular addition has not been replaced.

We are currently using a GUIDE-internal `MessageAnnouncer` class, whereas this PR uses the `TQDMProgressHandler` class from `tqdm_publisher` to ensure that messages are handled serially.

The intention of this PR was to move more responsibilities to `tqdm_publisher`, though I'm not personally invested in which solution we decide to use.

src/pyflask/manageNeuroconv/manage_neuroconv.py
```diff
@@ -1096,11 +1091,16 @@ def on_progress_update(message):
...
     from pickle import PicklingError
 
+    request_id = payload.get("request_id", None)
+
     from nwbinspector import load_config
     from nwbinspector.inspector_tools import format_messages, get_report_header
     from nwbinspector.nwbinspector import InspectorOutputJSONEncoder
 
     kwargs = dict(
+        n_jobs=-2,  # uses number of CPU - 1
```
CodyCBakerPhD, line 1101: Looks like `n_jobs` is already in the `payload?`

(no `payload` dumped to logs)

[2024-05-29_11-25-30.log](https://github.com/NeurodataWithoutBorders/nwb-guide/files/15486630/2024-05-29_11-25-30.log)

garrettmflynn: Ah right, thanks! This is actually all redundant anyways—so I've just removed all `payload` preprocessing here.

CodyCBakerPhD:
This appears to break the parallel bars; I'm unsure if it's parallelizing without informing, but I no longer see the sub-progress bars completing per file over a directory

garrettmflynn:
Good catch, thank you. There was still a leftover `announcer` declaration in the NeuroConv namespace.

CodyCBakerPhD (approved):

PR #705, opened 2024-03-26
Validate C Elegans Sex Correctly

garrettmflynn:
This PR allows us to validate C Elegans sex correctly by grouping the `sex` and `species` properties on the Subjects table to validate with each other. 

They now both validate when one changes. 

As far as I can tell, Handsontable doesn't allow for different dropdown configurations to be used per-cell—only per-column—so all options have to be shown all the time.


CodyCBakerPhD:
image

The invalidation message is incorrect in this case, see what it should be in the Inspector: https://github.com/NeurodataWithoutBorders/nwbinspector/blob/dev/src/nwbinspector/checks/nwbfile_metadata.py#L211

Also why is what I have selected there invalid?

garrettmflynn:
Ah I see. Works one way (Sex then Species) but not the other.

garrettmflynn:
Should be fixed now!

CodyCBakerPhD (approved):

Issue #748, opened 2024-04-16
[Enhancement]: Change "File Path" header to "Path to SpikeGLX ap.bin or lf.bin file"


rly:
### What would you like to see added to the NWB GUIDE?

On the source data page for SpikeGLX Converter and Recording, there is a description below the file selector box:
<img width="984" alt="image" src="https://github.com/NeurodataWithoutBorders/nwb-guide/assets/310197/3764d358-209f-4e16-8079-5898ee3e9cdf">

In my personal experience and experience observing the user tests, people click the box and are briefly confused before reading the smaller gray text below the box (and sometimes they don't even read that text - the gray makes it look like it is not particularly important and the fact that it is after the box suggests it is not relevant to the action of clicking the box). The headers "Folder Path" and "File Path" are generic and not very useful. I suggest that if the description exists, it replaces the header instead of being text after the box, or there is a new field per interface that provides a more descriptive replacement of the header. This affects only the handful of interfaces that have a description.

### Do you have any interest in helping implement the feature?

No.

### Code of Conduct

Yes

### Did you confirm this feature was not already reported?

Yes

rly:
In general, I wonder whether the gray description should be moved above the text box / input field.

CodyCBakerPhD:
These were all updated on the NeuroConv side

PR #756, opened 2024-04-17
Organize description and title rendering

garrettmflynn:
This PR attempts to address #748, but diverges from the suggestions introduced there.

IMO directly using those descriptions as titles is risky because `description` tends to be a long multi-word string whereas `title` tends to be fairly concise. Especially with multiple interfaces, this will likely be a lot of large, bolded text to focus on—whereas using the small gray text for descriptions allows for reducing emphasis.

To test out some ideas, though, I've implemented the following:
1. Do not use a title if the description is provided (i.e. only show the file box + description below). Otherwise use a default description about the filesystem path type expected
2. Move the description directly beneath the title if the latter exists


CodyCBakerPhD:
@garrettmflynn @rly I'm leaving his between you two as a post-release improvement

CodyCBakerPhD:
@rly Still interested in this?

CodyCBakerPhD:
From discussion; closing for now

A lot of this has been greatly improved on the description level already, on top of Garretts concerns this might to be the most stable approach

PR #762, opened 2024-04-22
Time alignment

garrettmflynn:
This PR implements time alignment.

@CodyCBakerPhD Would be great to get your help on a few final things:
1. Can you point me to an example for setting up a Dummy Recording for a sorting interface so that we can simply set timestamps / starting time?
2. The current iteration isn't properly setting timestamps on the preview conversion. Is there a reason this is happening?

Besides the above caveats, this is properly updating the time alignment information and requesting it back on the Source Data popup.


CodyCBakerPhD:
> Can you point me to an example for setting up a Dummy Recording for a sorting interface so that we can simply set timestamps / starting time?

Not a 'dummy' example, but a proper test case from example data: https://github.com/catalystneuro/neuroconv/blob/179711773fc67a2a733ebc6522afa0bea6f2f305/tests/test_on_data/test_sorting_interfaces.py#L29-L33

https://github.com/catalystneuro/neuroconv/blob/179711773fc67a2a733ebc6522afa0bea6f2f305/src/neuroconv/tools/testing/data_interface_mixins.py#L586-L589

Easy way to truly do a 'dummy' one is the use the official Mock one: https://github.com/catalystneuro/neuroconv/blob/179711773fc67a2a733ebc6522afa0bea6f2f305/tests/test_ecephys/test_ecephys_interfaces.py#L127-L131


> The current iteration isn't properly setting timestamps on the preview conversion. Is there a reason this is happening?

I'll look into it

garrettmflynn:
And just to confirm, we actually do want to allow users to specify timestamps or a start time for Sorting Interfaces, independent of any active Recording Interfaces?

Mentioned this to Ben at the developer hackathon and he was a bit confused about what this actually would mean.

CodyCBakerPhD:
If the exact recording interface used to generate the sorting is present in the pipeline, then they should link the sorting to the recording and adjust timestamps or shift starting time of the recording

Otherwise, they still need to be able to set timestamps and shift starting time _as if_ they had chosen to include the original recording

garrettmflynn:
Ah gotcha. Makes sense now

garrettmflynn:
So this is practically ready for review.

Now that I've expanded my tests beyond the SpikeGLX-Phy test case, though, I've realized that the Phy Sorting Interface in the tutorial data seems to be locked to the SpikeGLX Recording Interface regardless of any user-defined parameters. Is this an expected behavior?

Screenshot 2024-04-24 at 2 44 37 PM

This was the previous behavior before https://github.com/catalystneuro/neuroconv/pull/823—though an error was thrown about the `None` and `int` addition. If providing a mock recording interface, the specified start time would take effect but it wasn't clear what it was shifting.

garrettmflynn:
Also confirmed that this time offset on the Recording Interface does **not** show up in stub files but can be seen from Neurosift after a full conversion.

CodyCBakerPhD:
@garrettmflynn I don't see the `link recording` button on my end

image

garrettmflynn:
This is by design. I've implemented [a check that determines what interfaces are compatible with a given SortingInterface](https://github.com/NeurodataWithoutBorders/nwb-guide/blob/2bea2ba5ab3bcfcc1a4e91d4fd852c07a5a6e03b/pyflask/manageNeuroconv/manage_neuroconv.py#L755-L764) before display, ensuring that users only select valid options.

garrettmflynn:
It's possible to disable the option—possibly with a related message on hover—instead of removing it completely. Would that be preferable?

CodyCBakerPhD:
> This is by design. I've implemented [a check that determines what interfaces are compatible with a given SortingInterface](https://github.com/NeurodataWithoutBorders/nwb-guide/blob/2bea2ba5ab3bcfcc1a4e91d4fd852c07a5a6e03b/pyflask/manageNeuroconv/manage_neuroconv.py#L755-L764) before display, ensuring that users only select valid options.

Ahh, right you are - silly me I was using a nonsensible pipeline

That part is working as intended then

CodyCBakerPhD:
A couple things I noticed testing this

i) the option of linking between compatible sorting and recording is now failing for the tutorial data (no button shows up)

ii) I set timestamps for Suite2p (single plane; 250 frames); they seem to have 'uploaded' to the interface fine because the alignment bar did say the proper start and stop interval according to the timestamps; but then they do not show up even in the final converted file

iii) Nothing seems to be applying to stub mode, I'm investigating that one myself

garrettmflynn:
Updated so that (i) works! Can you try again? Would be interested to get the file you're using too so I can replicate.
- When I try to do (I), I am getting an error during the conversion—but I can successfully associate the two.

Just FYI I was able to use the Suite2P test pipeline and align to an arbitrary start time, which I could observe across stub and full conversions.

src/pyflask/manageNeuroconv/manage_neuroconv.py
```diff
@@ -674,25 +680,116 @@ def validate_metadata(metadata: dict, check_function_name: str) -> dict:
...
+                    elif method == "linked":
+                        interface.register_recording(converter.data_interface_objects[value])
+
+                    elif method == "start":
+                        interface.set_aligned_starting_time(
+                            value
+                        )  # For sorting interfaces, an empty array will still be returned
+
+                except Exception as e:
+                    errors[name] = str(e)
```
CodyCBakerPhD, line 736: I'm almost done refactoring a bit here and there on the backend for this

But how did you want the frontend to handle this error step given the recent advances in logging?

CodyCBakerPhD: I.e., can we just let them raise?

CodyCBakerPhD:
I discovered the issue with recording interfaces specifically, it's on the NeuroConv side so I'll push a fix over there but it shouldn't hold up this end of things

One question about error handling and then this should be good to go

CodyCBakerPhD:
FYI the issue I had with timestamps was a non-issue; the timestamps vector I was trying to set was regular, and NeuroConv magic automatically detected that and set to a regular rate. There's ongoing discussions over there about whether that should even happen at  all though

src/pyflask/manageNeuroconv/manage_neuroconv.py
garrettmflynn: So the reason for this is to render errors occurring in the Time Alignment pop-up next to their respective interfaces. If uncaught, you'd have to fail the entire request—while this allows us to limit failure to the impacted interface.

garrettmflynn: Whether and how to log these errors in the log file is a separate—and more interesting—question. Is that the desired outcome for you?

CodyCBakerPhD: Hmmm.. I see

OK we'll tackle that in a follow-up. Would probably opt to dump the list of compiled errors all at once

CodyCBakerPhD (approved):

PR #764, opened 2024-04-24
Update mac workflows to use arm64 requirements

rly:
The GitHub Action `Build-and-deploy-mac.yml` is currently failing, possibly because on GitHub Actions, `os: macos-latest` now defaults to a M1 runner: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories

Here, I updated the workflows to use `requirements-MAC-arm64.yml`

If we want to keep testing on a non-arm64 runner, we can pin to `os: macos-13` instead, or we can test on both arm64 and non-arm64.

### DIFF
--- a/.github/workflows/Build-and-deploy-mac.yml
+++ b/.github/workflows/Build-and-deploy-mac.yml
@@ -22,7 +22,7 @@ jobs:
       - uses: conda-incubator/setup-miniconda@v2
         with:
           activate-environment: nwb-guide
-          environment-file: environments/environment-MAC.yml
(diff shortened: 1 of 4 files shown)


CodyCBakerPhD:
> If we want to keep testing on a non-arm64 runner, we can pin to os: macos-13 instead, or we can test on both arm64 and non-arm64.

Yes, please - can you add that to the matrix?

CodyCBakerPhD (approved):

PR #770, opened 2024-05-02
Correctly handle converter electrodes

garrettmflynn:
This PR fixes the issue where the SpikeGLXRecordingInterface and SpikeGLXConverter have inconsistent behaviors and the latter cannot be converted.

The solution was to iterate over the interfaces of a sub-converter instead of checking it directly for relevant metadata.


CodyCBakerPhD:
Looks good - probably waiting on the big e2e PR for tests?

garrettmflynn:
That would be a great test case. Can add to the YAML here

garrettmflynn:
Any idea why these results might be returned though? 

Screenshot 2024-05-13 at 1 37 21 PM

CodyCBakerPhD:
@garrettmflynn Yes, that makes perfect sense because NIDQ electrodes aren't on a shank - and I'm fairly sure from digging into that sub interface that its individual metadata schema should not require it

Should be able to test that out on an individual NIDQ-only pipeline

The real problem is how NWB makes one common table for all electrodes, but the new ecephys extension will hopefully resolve that (and hopefully be integrated to core to resolve) - regardless, NeuroConv can currently resolve that issue for us

This would be a good thing for us to pair program a debug and potential refactor in a longer (~2 hour?) meeting tomorrow

garrettmflynn:
Awesome. Looking forward to it

garrettmflynn:
Fixed in tandem with https://github.com/catalystneuro/neuroconv/pull/860#event-12821303025

CodyCBakerPhD (approved):

PR #771, opened 2024-05-02
Complete conversion with only a sorting interface

garrettmflynn:
This PR fixes an issue discovered with @stephprince where users are not able to advance past the File Metadata page when they've only selected a sorting interface. This was because electrode information was not being ignored and triggered JSON Schema validation incorrectly.


CodyCBakerPhD:
Possible to add an end-to-end test for this case, since it did apparently pop up unpredictably?

garrettmflynn:
This seems like a perfect justification to get #580 through. Do you agree?

Otherwise, I'd essentially complete this for a single generated test pipeline when we could cover our bases more generally.

CodyCBakerPhD:
@garrettmflynn Absolutely. Let me know when that ones ready

CodyCBakerPhD:
@garrettmflynn So question is, would you prefer this in before or after the big e2e test PR?

garrettmflynn:
After is fine. Started working on it today, shouldn't be terribly complicated.

CodyCBakerPhD:
> After is fine. Started working on it today, shouldn't be terribly complicated.

Apparently the e2e PR is pointing to this PR, so this one would go it before

That's also probably the source of the changed files conflation

How do you want to handle this?

garrettmflynn:
Yep sounds right. Fixed that conflation

Whether we merge in the E2E updates into this branch before or after merging to main should be fine. Assumed we'd want to do the former since that would "add the test" to this PR —though it's all really the same.

CodyCBakerPhD (approved):

CodyCBakerPhD:
Keeping things linearly progressive might be cleanest then, getting this through followed by e2e, then on to other stuff

PR #772, opened 2024-05-06
Improve tutorial clarity

garrettmflynn:
This PR provides a wide range of improvements to our tutorials based on the Dev Hackathon user tests. 

Highlights include:
1. Explicitly named test data folders for each tutorial section (i.e. `single_session_data`/`multi_session_dataset` instead of `data`/`dataset`
2. Providing notice to create a DANDI account on the tutorial homepage
3. Linking the tutorial in the GUIDE internal documentation page
4. Align user tests and documentation with workflow changes

PR #776, opened 2024-05-13
Small Tests Improvements

garrettmflynn:
This PR ensures that test data is generated now that the new paths are provided. 

Additionally, this also introduces changes to stop rapidly (2s < 5min) in cases where certain cases where the E2E tests fail before DANDI Upload.

### DIFF
--- a/tests/e2e/config.ts
+++ b/tests/e2e/config.ts
@@ -102,7 +102,7 @@ export const subjectInfo = {
   }
 }
 
-export const regenerateTestData = !existsSync(testDataRootPath) || false // Generate only if doesn't exist
+export const regenerateTestData = !existsSync(testDataPath) || !existsSync(testDatasetPath) || false // Generate only if doesn't exist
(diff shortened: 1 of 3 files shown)


CodyCBakerPhD:
@garrettmflynn One conflict to resolve with e2e tests here

CodyCBakerPhD:
Also, are the test modifications here (or any current e2e tests) directly testing the item fixed by this PR?

garrettmflynn:
Yes and no. These behaviors can be observed downstream of certain changes like https://github.com/NeurodataWithoutBorders/nwb-guide/pull/772 (updated data save location) and failed tests before DANDI Upload. 

However, the CI either avoids these issues/fixes altogether (i.e. there is always a clean filesystem, so the data generation check was always working fine) OR already shows the failure—but simply takes a different amount of time to exit.

CodyCBakerPhD (approved):

PR #777, opened 2024-05-13
Improve Metadata Validation and Display

garrettmflynn:
In response to several issues raised during Dev Hackathon user tests, this PR introduces improved the validation and rendering of metadata values.

In particular, the following behaviors were introduces:
1. Clear files explicitly instead of assuming users want to clear the file selector if canceling
2. Add ESC shortcut for modals
3. Highlight text that can be edited in place
4. Reduce padding on search options
5. Limit date selection for subjects between 1900 and the current date
6. Add units to JSON Schema inputs (e.g. `kg` for subject weight)
7. Hide subject metadata from global popup so that it can only be edited on the Subject Table page.


src/renderer/src/stories/SimpleTable.js
```diff
@@ -995,6 +995,9 @@ export class SimpleTable extends LitElement {
 
         return html`
             ${this.#context}
+            <p style="margin: 0; margin-bottom: 10px">
+                <small style="color: gray;">Right click to add or remove rows.</small>
+            </p>
```
CodyCBakerPhD, line 1000: We can delete the above description now that this is hard-coded, right?

CodyCBakerPhD:
How many explicit tests of these new behaviors do you think we can include - or is the fact that the e2e work evidence that they don't per se 'break' anything as is?

garrettmflynn:
These are largely rendering / styling updates, so I don't believe it is essential to add new tests.

It's correct to say that nothing breaks because E2E tests are successful.

If you'd like, we _could_ add a test that clears the filesystem selector outside of the E2E test since this behavior isn't ever used. But that'd be the biggest one to include. It's simply an `onclick` event decoupled to any other part of the system—only the component itself—so you wouldn't really expect it to fail.

src/renderer/src/stories/SimpleTable.js
garrettmflynn: Other way around. Injected the dynamic description into this part

CodyCBakerPhD: Bah, just when I thought I resolved a merge conflict successfully on my own 😂

CodyCBakerPhD:
> These are largely rendering / styling updates

Or in other words, the chromatic stories _are_ the more appropriate tests for these changes?

CodyCBakerPhD (approved):

PR #780, opened 2024-05-15
Limit file extension selections

### LINKED ISSUE #170: [Idea] File selector, by default, only looks for 'valid' files
For the source data selection step for each interface, we usually know what the supported or common file suffixes are for that format (if a file based format) and likewise the folder structure required by the interface if it's a folder-based format

So it would be cool if the file selector, when pointed to a large folder with a bunch of files, filtered out all the ones not one of the designated suffixes

This would require a small addition to NeuroConv to actually write down that suffix relationship as a class attribute on each interface, similar to how we handle keywords right now

garrettmflynn:
fix #170


CodyCBakerPhD:
Tested this out on SpikeGLX-Phy pipeline (not the converter one)

image

Despite the suffixes from NeuroConv only having a single period: https://github.com/catalystneuro/neuroconv/blob/main/src/neuroconv/datainterfaces/ecephys/spikeglx/spikeglxdatainterface.py#L22

It seems two are added in the selector? Or does the format the 'extensions' of the `options.filters` expect them to not have leading periods?

Also, there is no option for manually choosing 'all files' as there normally is in such a file window - do you know how to add that?

schemas/source-data.schema.ts
```diff
@@ -29,6 +30,11 @@ export default function preprocessSourceDataSchema (schema) {
 
             }
 
+            if (files) {
+                const base = singleLocationInfo ? files : files.items
+                if (!base.accept && info.suffixes) base.accept = info.suffixes
+            }
```
CodyCBakerPhD, line 36: What does the `accept` property of `files` (or `files.items`) mean?

src/renderer/src/stories/FileSystemSelector.js
```diff
@@ -125,6 +127,11 @@ export class FilesystemSelector extends LitElement {
 
     #useElectronDialog = async (type) => {
         const options = { ...this.dialogOptions };
+
+        if (!options.filters && this.accept) {
```
CodyCBakerPhD, line 131: What are cases where `!options.filters`?

src/renderer/src/stories/FileSystemSelector.js
```diff
@@ -125,6 +127,11 @@ export class FilesystemSelector extends LitElement {
...
+            options.filters = [{ name: "Selected Files", extensions: this.accept }];
```
CodyCBakerPhD, line 132: Maybe just need to add another element here?

```suggestion
            options.filters = [{ name: "Selected Files", extensions: this.accept }, {name: "All Files", extensions: ["*"]}];
```

src/renderer/src/stories/JSONSchemaInput.js
```diff
@@ -893,6 +893,7 @@ export class JSONSchemaInput extends LitElement {
             const filesystemSelectorElement = new FilesystemSelector({
                 type: format,
                 value: this.value,
+                accept: schema.accept,
```
CodyCBakerPhD, line 896: Similar to above, what does the `accept` field of the `schema` here mean?

CodyCBakerPhD:
I don't suppose there's a way to add a small vite test for this component behavior?

garrettmflynn:
No, we can't even run an E2E test because this is reliant on a native browser pop-up.

src/renderer/src/stories/FileSystemSelector.js
garrettmflynn: Yep this allows for changing the filter. Thanks!

garrettmflynn:
> It seems two are added in the selector? Or does the format the 'extensions' of the options.filters expect them to not have leading periods?

Huh Mac didn't care. But you're right, this expects no leading period.

src/renderer/src/stories/FileSystemSelector.js
garrettmflynn: Usually this is not true. But if someone wants to use this component in isolation, it'd be possible for them to override the default filters behavior (based on the `accept` field of the schema) with their own declaration

schemas/source-data.schema.ts
garrettmflynn: `accept` is an extension of the JSON Schema to support configuration of the `[<input type="file">](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file)` element. Without this, there is no way to pass this information into the JSON Schema Form in a way that remains associated with a single interface's file property.

There are a few of these extensions that have been used across the GUIDE (e.g. `order`, `enumLabels`, etc.)

src/renderer/src/stories/FileSystemSelector.js
```diff
@@ -142,18 +149,28 @@ export class FilesystemSelector extends LitElement {
...
             this.#onThrow("Incorrect filesystem object", `Please provide a <b>${this.type}</b> instead.`);
+
+        if (this.accept) {
+            const ext = value.split(".").pop();
+            if (!this.accept.includes(ext))
+                this.#onThrow(
+                    "Incorrect file extension",
+                    `Please provide a file with the extension <b>${this.accept.join(", ")}</b> instead.`
+                );
+        }
```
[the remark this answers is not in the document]
garrettmflynn: Yeah now that we're loosening the selection and allowing different filters, this is correct. We can just remove this check altogether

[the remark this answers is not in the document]
garrettmflynn: Or we could provide matching as a warning. What do you think?

src/renderer/src/stories/JSONSchemaInput.js
garrettmflynn: Here's the analogous attribute on the `<input type="file">` element: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/accept

CodyCBakerPhD: Thanks 👍

src/renderer/src/stories/FileSystemSelector.js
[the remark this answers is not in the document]
CodyCBakerPhD: What would the warning look like? I'd probably have to see how it feels and think about how often it might occur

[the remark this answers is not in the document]
garrettmflynn: Just like the warnings displayed for an improper session start time, etc.

If you think this might trip up users, it's probably best to cut it. I'd just wanted to minimize any inconsistencies between the native pop-up selection and the drag-and-drop selection. Though now that there are two filters, we technically allow anything

CodyCBakerPhD (approved):

CodyCBakerPhD:
Works great!

src/renderer/src/stories/FileSystemSelector.js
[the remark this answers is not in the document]
CodyCBakerPhD: I think it might, only because I don't count on NeuroConv handling every possible edge case and don't want to unduly frighten or confuse users

So I say remove entirely, no warning or error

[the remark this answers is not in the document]
garrettmflynn: Alright removed!

CodyCBakerPhD (approved):

PR #784, opened 2024-05-18
cleanup: typehinting and import optimization

bendichter:


CodyCBakerPhD (approved):

CodyCBakerPhD:
What's weird is that we had `isort` in the `pyproject` configuration, just didn't have it enabled via pre-commit

PR #786, opened 2024-05-20
Add auto-update workflow

garrettmflynn:
This PR implements a user-driven update workflow for the GUIDE that notifies users of a new release on the Settings page. This should only show if a new release is actually present compared to the current version.
Screenshot 2024-05-20 at 11 52 16 AM

Clicking the info icon will display a record of the release notes. 
Screenshot 2024-05-20 at 1 12 39 PM

The Update button then triggers the `electron-updater` package to download the latest release (with a related progress bar) and asks the user if they'd like to restart the application now (activating the downloaded release) or defer this until a manual close.

To test, I'd suggest that you create a local build and go through this workflow. The success of this process on Mac depends on proper code-signing, but I believe you should be able to complete this on a Windows PC.


src/renderer/src/stories/ProgressBar.ts
```diff
@@ -5,13 +5,37 @@ import { LitElement, html, css, unsafeCSS } from 'lit';
...
+    const units = ['B', 'KB', 'MB', 'GB', 'TB', 'PB', 'EB', 'ZB', 'YB'];
+
+    // Initialize the index to 0
+    let index = 0;
+
+    // Convert the size to a floating point number
+    size = parseFloat(size);
+
+    // Loop until the size is less than 1024 and increment the unit
+    while (size >= 1024 && index < units.length - 1) {
```
CodyCBakerPhD, line 29: We tend to use base 10 by default

Also base 2 would be consistent with adding `iB` to all the units above

CodyCBakerPhD:
I can take a crack at rebasing this after the next 2 frontend refactors are done

@garrettmflynn the next one was the separation of frontend tests and stories from other assets/pages, correct?

src/electron/frontend/core/pages.js
```diff
@@ -1,32 +1,33 @@
...
-import { sections } from "./components/pages/globals";
-import { GuidedSubjectsPage } from "./components/pages/guided-mode/setup/GuidedSubjects";
-import { GuidedSourceDataPage } from "./components/pages/guided-mode/data/GuidedSourceData";
-import { GuidedMetadataPage } from "./components/pages/guided-mode/data/GuidedMetadata";
-import { GuidedUploadPage } from "./components/pages/guided-mode/options/GuidedUpload";
-import { GuidedResultsPage } from "./components/pages/guided-mode/results/GuidedResults";
-import { Dashboard } from "./components/Dashboard";
-import { GuidedStubPreviewPage } from "./components/pages/guided-mode/options/GuidedStubPreview";
-import { GuidedInspectorPage } from "./components/pages/guided-mode/options/GuidedInspectorPage";
+import { GettingStartedPage } from "./stories/pages/getting-started/GettingStarted";
```
CodyCBakerPhD, line 1: @garrettmflynn Was the `GettingStarted` page intended to be removed from `src/electron/frontend/core/pages.js`? Just checking

garrettmflynn: Yeah it's been unused for quite some time. We could even remove it from the source files

src/renderer/src/stories/ProgressBar.ts
garrettmflynn: Whichever you'd prefer. I think it'd be preferable to avoid a network request—though I also understand the desire for consistency.

src/electron/frontend/core/pages.js
CodyCBakerPhD: OK I just removed it

src/renderer/src/stories/ProgressBar.ts
CodyCBakerPhD: Maybe mess with it in a follow-up if we have time

Swapped it to base 10 at least here

CodyCBakerPhD:
OK this PR is rebased; however, it always looks as if it triggered some problem in the Windows test (perhaps that same you alluded to about modals popping up and closing?)

CodyCBakerPhD:
Worked great on Windows!

garrettmflynn:
> Worked great on Windows!

Where did it do this? I have no idea what's going on here 😅

CodyCBakerPhD:
> Where did it do this? I have no idea what's going on here 😅

I mean the full build and distributable on Windows; it showed the auto-update thing in settings and it all looked perfect; auto-updating installed the current released version in the previous installation directory

CodyCBakerPhD (approved):

garrettmflynn:
Ah awesome!

PR #791, opened 2024-05-22
Flask-side refactors

CodyCBakerPhD:
Enhancing the Swagger with more docstrings, flatter code, more centralized error messages, and probably cleaner submodule localization

### DIFF
--- a/pyflask/app.py
+++ b/pyflask/app.py
@@ -12,14 +12,12 @@ from pathlib import Path
 from signal import SIGINT
 from urllib.parse import unquote
 
-from errorHandlers import notBadRequestException
-
 # https://stackoverflow.com/questions/32672596/pyinstaller-loads-script-multiple-times#comment103216434_32677108
 multiprocessing.freeze_support()
 
 
 from apis import data_api, neuroconv_api, startup_api
-from flask import Flask, request, send_file, send_from_directory
+from flask import Flask, Response, request, send_file, send_from_directory
 from flask_cors import CORS
 from flask_restx import Api, Resource
 from manageNeuroconv.info import (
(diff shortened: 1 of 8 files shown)


pyflask/app.py
```diff
@@ -54,46 +53,85 @@
 api.add_namespace(startup_api)
 api.add_namespace(neuroconv_api)
 api.add_namespace(data_api)
-api.init_app(app)
+api.init_app(flask_app)
+
+# 'nwbfile_registry' is a global list that keeps track of all NWB files that have been registered with the server
+nwbfile_registry = []
```
CodyCBakerPhD, line 59: Actually, I do remember why this is a `dict` (faster lookup via hash map) so I'll change that back after we discuss if/how it's used

pyflask/app.py
```diff
@@ -54,46 +53,85 @@
...
-                path = f"/{path}"
-            return send_file(path)
-        else:
-            app.abort(404, "Resource is not accessible.")
+@flask_app.route(rule="/files/<file_path:path>", methods=["GET", "POST"])
+@api.doc(
+    description="Handle adding and fetching NWB files from the global file registry.",
+    responses=server_error_responses(codes=[200, 400, 404, 500]),
+)
+def handle_file_request(file_path: str) -> Union[str, Response, None]:
```
CodyCBakerPhD, line 98: @garrettmflynn Perhaps you can help clarify this endpoint

From https://github.com/search?q=repo%3ANeurodataWithoutBorders%2Fnwb-guide+%2Ffiles&type=code it seems it is only used in one place (the preview page for serving things to Neurosift) and only used as a POST and does not re-use the registry in that way (only purpose seem to be to get and return the `base_url` of the path)

Is there anywhere that uses the GET, or the other endpoint (https://github.com/NeurodataWithoutBorders/nwb-guide/pull/791/files#diff-d8bafe4fe7c605b4e3a4a4f41367e387da934fae16aa61026e33413861366228R89) that refers to the global registry?

garrettmflynn: Neurosift implicitly uses the GET endpoint after [setting the `url` parameter](https://github.com/NeurodataWithoutBorders/nwb-guide/blob/9960b422556f10a82f8ac638452f4fabe187c2bc/src/renderer/src/stories/pages/preview/PreviewPage.js#L23).

The idea is to only allow access to files the GUIDE knows the absolute path to.

PR #792, opened 2024-05-22
Reorganize core folders

garrettmflynn:
This PR reorganizes the core folder structure of the GUIDE to more clearly separate `src` files


CodyCBakerPhD:
Doing more thorough testing later to be sure

garrettmflynn:
Gotcha. Thank you!

electron.vite.config.js
```diff
@@ -2,11 +2,36 @@ import { defineConfig } from "electron-vite";
...
     renderer: {
+        root: "./src/electron/renderer",
         plugins: [ViteYaml()],
+        build: {
+            rollupOptions: {
+                input: {
+                    index: resolve(__dirname, "index.html"),
+                },
+            },
+        },
```
CodyCBakerPhD, line 35:
```suggestion
        root: "./src/electron/renderer",
        plugins: [ViteYaml()],
        build: {
            rollupOptions: {
                input: {
                    index: resolve(__dirname, "src/electron/renderer/index.html"),
                },
            },
        },
```

@garrettmflynn I'm encountering trouble here doing a local build (`npm run build:win`)

 Without this change everything worked fine in dev mode (`npm start`)

But when I tried to build and test that build, that there was a rollup issue finding the `index.html` file - making this change fixed it and doesn't seem to break dev mode either (which only seems sensitive to the `root`)

However, once I overcome that index issue I then run into 

```bash
cannot find specified resource "electron/renderer/assets/app-icon/logo-guide-draft.ico", nor relative to "D:\GitHub\nwb-guide\build", neither relative to project dir ("D:\GitHub\nwb-guide")
```

and the resulting `build` folder does not contain the `renderer` subfolder either

Any ideas?

CodyCBakerPhD:
Also I just noticed something in the CI - what are the 'Example Pipeline Tests' testing if they don't have any example data? Just the tutorials?

CodyCBakerPhD:
Hmm.. I guess it just skips everything? https://github.com/NeurodataWithoutBorders/nwb-guide/actions/runs/9198157905/job/25300243822?pr=792#step:10:175

garrettmflynn:
Oh no! Yep you got it. I forgot to add the GIN download steps to the workflow after testing locally...

CodyCBakerPhD:
> Oh no! Yep you got it. I forgot to add the GIN download steps to the workflow after testing locally...

We can deal with that in a follow-up then

Current problem holding up PR is for building whereas the pipeline tests use dev mode, right?

garrettmflynn:
Correct. Trying to figure out where the `renderer` folder went now.

electron.vite.config.js
garrettmflynn: Looks like it was being output in the `root` location (i.e. `/src/electron/renderer/build`). Updated the configuration to explicitly output to the repo build folder.

garrettmflynn: Full build renders well!

CodyCBakerPhD:
OK, I still get the same warning when building on windows

```
cannot find specified resource "electron/renderer/assets/app-icon/logo-guide-draft.ico", nor relative to "D:\GitHub\nwb-guide\build", neither relative to project dir ("D:\GitHub\nwb-guide")
```

(which is true, there is no 'electron' folder in 'renderer' in 'build') and the `.exe` in the build folder fails

But the `dist` works just fine so getting this through

CodyCBakerPhD (approved):

garrettmflynn:
Ah figured it out. Just a simple bad path in the `package.json` file. Will submit a follow-up

PR #793, opened 2024-05-24
Fix installer icon after refactor

garrettmflynn:
Fix installer icon path after #792


CodyCBakerPhD:
Ahhh I didn't think to look in the `package.json`

CodyCBakerPhD (approved):

PR #795, opened 2024-05-24
Download GIN Data for Pipeline Tests

garrettmflynn:
This PR fixes the issue noted in #792 where the pipeline tests don't actually test anything because there is no test data found. 

@CodyCBakerPhD Would you mind adding all the appropriate secrets for this workflow to run?


CodyCBakerPhD:
@garrettmflynn OK I got this to a more streamlined point in terms of deployment, display, etc.

However, the pipeline tests are still skipping everything; perhaps a data path needs to be configured in some way?

Would it be possible to adjust those tests to fail instead of skip when files are missing (ideally with an informative error message of 'data not found' per test)?

Feel free to take over this PR again to try to get the CI working as expected

garrettmflynn:
Yeah the data needs to go to a very specific location in the filesystem. I've made it so that the test should fail if it isn't there (instead of skipping).

garrettmflynn:
Are you / can you place the testing data in `~/NWB_GUIDE/test-data/GIN`?

garrettmflynn:
I think I have to clear / update the cache for the updates to take effect. How would I do this?

CodyCBakerPhD:
> I think I have to clear / update the cache for the updates to take effect. How would I do this?

Actions tab has a subgroup for 'caches' from which you can delete them, you should have access I believe - but I cleaned it up for you now

CodyCBakerPhD:
Huh, e2e test reports and a successful conversion but errors at the end due to some expectation? https://github.com/NeurodataWithoutBorders/nwb-guide/actions/runs/9262160305/job/25478666338?pr=795#step:17:324

garrettmflynn:
This is the same error I've been seeing. Can you confirm that the test data was successfully downloaded in the GIN subfolder?

CodyCBakerPhD:
OK caches are properly downloading and saving now - before it was because the path at that location did not exist (AWS CLI does not make it for us automatically)

CodyCBakerPhD:
Half of the pipelines run fine now - some path error now it looks like for Windows: https://github.com/NeurodataWithoutBorders/nwb-guide/actions/runs/9263643652/job/25482471310?pr=795#step:16:205

And some assertion issue in Mac (Intel)? https://github.com/NeurodataWithoutBorders/nwb-guide/actions/runs/9263643652/job/25482471193?pr=795#step:16:4794

CodyCBakerPhD:
Problem with Windows seems to be with the caches not saving any of the content in the folder despite the transfer going smoothly

Would it be possible to specify the base path given to the tests as relative? The CI usually plays better with that than with `~`, especially for windows

CodyCBakerPhD:
For example, can the tests use a base path that is specified as an environment variable, or if that variable doesn't exist then follow existing behavior?

garrettmflynn:
Hmm this is really odd. Any idea why it can't find / won't create the `.env` file? https://github.com/NeurodataWithoutBorders/nwb-guide/actions/runs/9290704253/job/25567601636?pr=795

CodyCBakerPhD:
Is there any way to fetch environment variables without a .env file?

CodyCBakerPhD:
If we can't get this working on Windows by EOD I say we just disable pipelines for windows to get this in and then work on Windows in followup

.github/workflows/test_env_creation.yml
```diff
@@ -0,0 +1,66 @@
...
+        uses: actions/checkout@v2
+
+      - name: Set working directory as environment variable
+        run: echo "CURRENT_WORKING_DIR=$(pwd)" >> $GITHUB_ENV
+        if: runner.os != 'Windows'
+
+      - name: Set working directory as environment variable (Windows)
+        run: echo "CURRENT_WORKING_DIR=%cd%" >> $env:GITHUB_ENV
+        shell: cmd
+        if: runner.os == 'Windows'
```
CodyCBakerPhD, line 43: For this style, I would recommend this syntax at the top of the file

```yaml
env:
  CURRENT_WORKING_DIR: ${{ github.workspace }}
```

garrettmflynn: Oh sure. I'm hoping to remove this ASAP.

It shows that we can, in fact, save a .env file and display the value on Windows. The `Set working directory as environment variable` option was simply to see if we can access something like this cross-platform. Will try this global env variable out!

Just switched the test pipeline workflow to use the exact same configuration for saving the `.env` file. Should tell us whether we're configuring GitHub Actions wrong or the test pipeline action is somehow not compatible with saving an `.env` file.

CodyCBakerPhD: > It shows that we can, in fact, save a .env file and display the value on Windows. The Set working directory as environment variable option was simply to see if we can access something like this cross-platform. Will try this global env variable out!

I saw that... CI can be really confusing sometimes

One thing it might be is the conda activation step, which swaps to a different shell type (and maybe the `.env` thing is only available from the default `bash`??)

garrettmflynn: Hmm although is this only happening in this PR?

garrettmflynn: Nevermind, this was just a consequence of not registering the DANDI API Key secret

CodyCBakerPhD:
OK just gonna get this in, continue work on Windows over on #815

CodyCBakerPhD (approved):

PR #797, opened 2024-05-25
[Refactor Flask I] Remove previous error filter

CodyCBakerPhD:
Splitting up a few things to try to get to the bottom of the problem on #796 

I don't think the diff from this changelog really does anything, right?


garrettmflynn (approved):
Looks good. Just removes `notBadRequestException` references, the parent file, and related imports.

PR #799, opened 2024-05-27
[Refactor Flask II] Break up top-level

CodyCBakerPhD:
Rebase and splinter of #796 specific to offloading various endpoints from the main `app.py` into distinct namespaces


prettier.config.js
```diff
@@ -1,16 +1,16 @@
...
-    singleQuote: false,
-    tabWidth: 4, // to match Python
-    trailingComma: "es5",
-    useTabs: false,
-    vueIndentScriptAndStyle: false,
+  arrowParens: "always",
+  bracketSameLine: false,
+  bracketSpacing: true,
+  embeddedLanguageFormatting: "auto",
+  endOfLine: "crlf",
```
CodyCBakerPhD, line 6: Testing to see if this is maybe what was throwing it off

CodyCBakerPhD: Moved to #800

CodyCBakerPhD:
Well, for reasons that are currently beyond me, our attempt last week to simplify the serving of paths to neurosift has been causing issues so I reverted to the previous structure and we can tackle it in a follow-up

If I go for the `flask_restx.Resource` structure inside a separate namespace (and adjust all fetches to that new namespace accordingly) I get

image

whereas it should normally look like

image

PR #800, opened 2024-05-27
Fix Prettier EOL style to be consistent across platforms

CodyCBakerPhD:
I think this might be the reason I was seeing those local vs webhook differences

Basically the 'auto' setting differs across platforms and the bot of course only runs on linux


garrettmflynn (approved):
Ah gotcha. Seems fine to me.

PR #815, opened 2024-05-30
Restore windows pipelines

CodyCBakerPhD:
Restoring the previous state of attempting to fix Windows pipelines

There also seems to be a new failure in the Windows Dev suite following merge of https://github.com/NeurodataWithoutBorders/nwb-guide/pull/786

At time of  merge it was only seen in the LiveServices tests (and thus disregarded as an impersistent live service specific problem) and not in the dev tests; but it's been seen in the dev tests for Windows ever since as well


CodyCBakerPhD:
Looks like this mistakenly had something setup to point to the parent as a remote; moved to #816

PR #817, opened 2024-05-30
Add daily notification

CodyCBakerPhD:
So maintainers can be notified of failures in daily workflows


garrettmflynn (approved):
Awesome thanks!

PR #790, opened 2024-05-20
Custom DateTime Selector

garrettmflynn:
This PR addresses #591, specifically [starting a draft PR that shows how to get started with a fix](https://github.com/NeurodataWithoutBorders/nwb-guide/issues/591#issuecomment-2117798813).

In particular, it creates a relevant Storybook story to highlight the existing `DateTimeSelector` component that can be modified to replace the native date-time popup with an element that optionally includes time.
Screenshot 2024-05-20 at 4 12 39 PM

PR #820, opened 2024-06-03
Timezone Control in the GUIDE

garrettmflynn:
This PR allows users to control the timezone used across the GUIDE from the Workflow page. Currently, this PR only allows you to _select_ a timezone—not to actually apply this timezone to the resulting data.

@CodyCBakerPhD Still not 100% sure what the best way to propagate the timezone to the appropriate places in the backend? Would you be able to indicate how I should pass a timezone to NWB Inspector (e.g. [using the `validate_nwbfile_metadata` function](https://github.com/NeurodataWithoutBorders/nwb-guide/blob/224675f69e7a40fbded63ea6db6426203cb179bd/src/pyflask/manageNeuroconv/manage_neuroconv.py#L656-L667)) as well as to NeuroConv during the conversion process?

This PR originated from a discussion following the merge of #790.


src/schemas/timezone.schema.ts
```diff
@@ -0,0 +1,41 @@
...
+    'Europe/Moscow': '+03:00',
+    'Asia/Dubai': '+04:00',
+    'Asia/Karachi': '+05:00',
+    'Asia/Dhaka': '+06:00',
+    'Asia/Jakarta': '+07:00',
+    'Asia/Shanghai': '+08:00',
+    'Asia/Tokyo': '+09:00',
+    'Australia/Sydney': '+10:00',
+    'Pacific/Auckland': '+12:00'
+};
```
bendichter, line 26: I think this is likely to confuse people due to daylight savings time. Most of these timezone names actually have different offsets depending on the day of the year. zoneinfo takes care of this automatically- you input the name of the timezone and it will check the date so that the correct offset is applied. I wish we could apply a simplifying approach like this but honestly I think this is going to confuse a lot of people if implemented this way

bendichter: I honestly think a very long dropdown would be better. I know it's not ideal but it is at least accurate

bendichter:
I guess a simple alternative would be to just provide offsets without names

CodyCBakerPhD:
> I guess a simple alternative would be to just provide offsets without names

I think this might be best overall

I really worry about

> I honestly think a very long dropdown would be better. I know it's not ideal but it is at least accurate

Because we already get feedback in user tests about information overload

And basic google searche of 'what time zone am i in' (maybe we can have a description hyperlink https://www.google.com/search?q=what+tiemzone+am+i+in for text 'find my timezone' somewhere?) reports the offset for GMT, so I don't think it's unreasonable for users to be responsible for looking up their own timezone if they don't know it

src/schemas/timezone.schema.ts
CodyCBakerPhD, line 26: To answer

> @CodyCBakerPhD Still not 100% sure what the best way to propagate the timezone to the appropriate places in the backend?

I would make the inverse of this lookup on the Python side (`offset_to_zoneinfo_name`), and then it's just `tzinfo=ZoneInfo(offset_to_zoneinfo_name[offset_from_payload])` when making the session start time `datetime`

garrettmflynn: Just pushed the option where we are rendering the browser's acceptable timezones as a native `<select>` element.

Screenshot 2024-06-03 at 12 21 00 PM

How's this? Would you prefer it to be searchable? What about the formats used by `zoneinfo` that aren't in `America/Chicago` format?

CodyCBakerPhD: This approach would absolutely need to be searchable (probably as our loose selector); it's overwhelming scrolling through a list that massive

I'd still recommend we simplify with a basic numeric scroller, that's many fewer options and removes the complexity of tying the number to a specific place (which has a somewhat low chance of being _exactly_the city they are located in)

CodyCBakerPhD: Nevermind, here's an even easier method if you just have a selector of +/- `00:00` through `12:00`

Get the ISO format of the timestamp without the timezone set, something like

```python
my_timestamp = "2024-06-03T14:15:06"
```

then inject the offset into the end of the string and reform a datetime from the ISO

```python
my_datetime = datetime.datetime.fromisoformat("2024-06-03T14:15:06-07:00")
```

It's a shame this is so roundabout, but PyNWB complains about the type otherwise

garrettmflynn: Beautiful. This works great as an end-to-end solution. Thank you!

garrettmflynn:
@CodyCBakerPhD Can you take a look at this? Almost ready, just think we need to adjust NWB Inspector to allow for adjusted timestamps. At present, a timestamp that is correctly rejected will not have a reaction when the GMT suffix is added.

CodyCBakerPhD:
I don't see a timezone area on the session start time selector for the file metadata page

image

Was there something else you wanted me to look at?

garrettmflynn:
Oh sorry, this just implements a global configurable timezone on the Workflow page.

Implementing a selector-level selector will be significantly more complicated, if only for the back-and-forth. Should squeeze that in with an optional time—as those UI components are going to have to play nice together.

rly:
Could we make the user's current time zone appear first in the list? That's the most common case. 

Secondly, could we enable search on any part of the time zone name? E.g. currently searching "new" does not return anything. I have to start with "America/New"

rly:
I found some interesting ux data on time zone selectors: https://www.nngroup.com/articles/time-zone-selectors/
(TL;DR guide at the bottom)

CodyCBakerPhD:
> Oh sorry, this just implements a global configurable timezone on the Workflow page.

Ah, thanks I missed that

If I click off of the selector mid-choice it autofills with this

image

and I have to delete it to get back to valid choices

I can also proceed to the next pages with `undefined` but if I go back it reverts to `Pacific/Tongatapu` for some reason

I would also make this a required field (red asterisk)

CodyCBakerPhD:
Also when I try to create a non-stub file I get this error displayed

image

But no console.log or log file contains a traceback - is this something on the frontend?

CodyCBakerPhD:
(the neurosift preview of the stub file looks great, the timezone is set correctly too)

garrettmflynn:
> Could we make the user's current time zone appear first in the list? That's the most common case.

@rly Assuming it isn't sufficient to have the user's current timezone be the default then? As far as I read, the article you'd sent suggested we minimize reordering and otherwise provide clear indicators if so (e.g. offset, region, etc.)

garrettmflynn:
@CodyCBakerPhD Fixed the conversion block.

> If I click off of the selector mid-choice it autofills with this

As for this, have you found a reliable way to reproduce this? I've been trying to click off after opening the Search dropdown, but I'm unable to get `undefined` to render. I've seen this intermittently before, so hopefully we'll be able to lock it down.

garrettmflynn:
Also I've updated the timezone dropdown to organize timezones in alphabetical order by group. This is now also searchable by city name as well as region.
Screenshot 2024-06-04 at 7 54 25 AM

garrettmflynn:
Update again so the browser handles the subtitle rendering. Instead of the raw text, I've chosen to render the short and long timezone name.

Screenshot 2024-06-04 at 8 24 52 AM

CodyCBakerPhD:
> As for this, have you found a reliable way to reproduce this? I've been trying to click off after opening the Search dropdown, but I'm unable to get undefined to render. I've seen this intermittently before, so hopefully we'll be able to lock it down.

Not since your latest pushes

I can also write the file successfully now

CodyCBakerPhD:
Working great on my end, just need to get tests passing and should be good to go

garrettmflynn:
Just to check, did we get the atomic NWB Inspector checks to play nice with the timezone offset?

CodyCBakerPhD:
@garrettmflynn One conflict now

garrettmflynn:
@CodyCBakerPhD Switched things around to rely on Python for timezone information.

Can you update this to pass the global `timezone` payload value in `convert_to_nwb` to the correct place in NeuroConv?

BTW you'll need to add `tzlocal` to your environment for this to work.

environments/environment-Linux.yml
```diff
@@ -21,3 +21,4 @@ dependencies:
       - pytest-cov == 4.1.0
       - scikit-learn == 1.4.0
       - tqdm_publisher >= 0.0.1
+      - tzlocal >= 5.2
```
CodyCBakerPhD, line 24: We don't need `tzlocal` for anything here

src/pyflask/namespaces/time.py
```diff
@@ -0,0 +1,31 @@
...
+    @time_namespace.doc(
+        description="Request the available timezones on the system.",
+    )
+    def get(self) -> Union[Dict[str, int], None]:
+        from zoneinfo import available_timezones
+
+        return list(available_timezones())
+
+
+@time_namespace.route("/timezone")
```
CodyCBakerPhD, line 22:
```suggestion
@time_namespace.route("/local_timezone")
```

environments/environment-Linux.yml
CodyCBakerPhD: Hmm nevermind, though `datetime.astimezone()` does return a correct local time it does not return the name of the corresponding zoneinfo

src/pyflask/namespaces/time.py
```diff
@@ -0,0 +1,31 @@
...
+class GetTimezones(flask_restx.Resource):
+
+    @time_namespace.doc(
+        description="Request the current timezone on the system.",
+    )
+    def get(self) -> Union[Dict[str, int], None]:
+        from tzlocal import get_localzone
+
+        return str(get_localzone())
```
CodyCBakerPhD, line 31:
```suggestion
        return get_localzone_name()
```

I think this is probably what you want right? The zoneinfo string key

src/pyflask/namespaces/system.py
```diff
@@ -20,3 +20,26 @@ def get(self) -> Union[Dict[str, int], None]:
         logical = cpu_count()
 
         return dict(physical=physical, logical=logical)
+
+@system_namespace.route("/all_timezones")
```
CodyCBakerPhD, line 24: attn @garrettmflynn I changed the source and names of these endpoints

src/pyflask/manageNeuroconv/manage_neuroconv.py
```diff
@@ -991,6 +991,11 @@ def update_conversion_progress(message):
 
                 del ecephys_metadata["ElectrodeColumns"]
 
+        # Correct timezone in metadata fields
+        resolved_metadata["NWBFile"]["session_start_time"].replace(tzinfo=zoneinfo.ZoneInfo(info["timezone"]))
+        if "date_of_birth" in resolved_metadata["Subject"]:
+            resolved_metadata["Subject"]["date_of_birth"].replace(tzinfo=zoneinfo.ZoneInfo(info["timezone"]))
```
CodyCBakerPhD, line 997: @garrettmflynn Assuming `info["timezone"]` is that thing selected on the pre-form page, this is how it should be used to modify those time fields prior to creating the file

I believe this helper is used in both preview and full conversion mode

src/pyflask/namespaces/system.py
CodyCBakerPhD: I went ahead and adjusted the fetch calls

src/pyflask/manageNeuroconv/manage_neuroconv.py
CodyCBakerPhD: Looks like it works!

garrettmflynn:
Looks good! That change is on me, I've added it back now that we're pulling from the Python list.

Screenshot 2024-06-04 at 4 37 17 PM

CodyCBakerPhD:
Oky doky, last thing is just fixing the tests then this is good to go

CodyCBakerPhD (approved):

CodyCBakerPhD:
@garrettmflynn According to chromatic, this PR also introduces the 'edit default values' button to all source data pages'?

There also seem to be 3 missing story pages: https://www.chromatic.com/build?appId=640a4d93e34604f275368983&number=997

Any idea what that's about?

garrettmflynn:
Yes working on it now. Seems to be something we didn't catch during the refactor after trying to remove the dev (web) mode.

The stories rely on storyStates.ts and saving / loading progress files—which was, in large part, removed.

Some of this was that I wasn't providing a timezone properly without backend access. But there's more to the story than that.

garrettmflynn:
Should be fixed. A lot of this also had to do with the addition of the Workflow (Preform) page.

CodyCBakerPhD (approved):

CodyCBakerPhD:
Great, thanks!

Issue #814, opened 2024-05-29
Improve NIH Award Number Handling


garrettmflynn:
### Describe the issue

We discovered during the Developer Hackathon tests that an incorrect NIH Award Number will still allow you to submit, resulting in the registration of your Dandiset but a notice that the ID isn’t assigned to your account.
- This behavior may be related to #758, though I'm sure it doesn't help embargoed Dandiset registration that a completely incorrect Award can be provided.

An additional issue was discovered that [copying an Award Number straight off the NIH Reporter](https://reporter.nih.gov/search/foCeX4tOvEW6_DyBAgyg_A/project-details/7227029) does not work because of improper spacing. The GUIDE should have a way to parse this directly.

### Steps to Reproduce

1. Complete a full conversion on a DANDI Upload workflow
2. Click Create New Dandiset button
3. Click that you would like to embargo your Dandiset
4. Provide any arbitrary text in the NIH Award Number slot

### Operating System

Mac M2

### GUIDE Version

https://github.com/NeurodataWithoutBorders/nwb-guide/commit/b60abb0e64021d0056ce49fd78695c1e76baf0c0

### Code of Conduct

Yes

### Did you confirm this issue was not already reported?

Yes

garrettmflynn:
> An additional issue was discovered that [copying an Award Number straight off the NIH Reporter](https://reporter.nih.gov/search/foCeX4tOvEW6_DyBAgyg_A/project-details/7227029) does not work because of improper spacing. The GUIDE should have a way to parse this directly.

Skipping this since DANDI itself does not accept these directly.

Issue #758, opened 2024-04-18
Cannot see my embargoed dandisets


CodyCBakerPhD:
### Describe the issue

Ran into this with a user the other day and just reproduced myself

Even when creating a dandiset in the GUIDE (or on the website) in embargo mode, I cannot see it in the loose selector as the target for upload

Also can't see any of my older ones that were previously created

### Steps to Reproduce

create an embargoed dandiset

try to upload to that dandiset in the GUIDE

### Operating System

M1 Mac

### GUIDE Version

0.0.15 and dev

### Code of Conduct

Yes

### Did you confirm this issue was not already reported?

Yes

PR #825, opened 2024-06-03
Remove embargo option from Dandiset creation

garrettmflynn:
As a result of #814 and #758, we are removing embargo from the first release.


CodyCBakerPhD:
BTW I heard this might be due to a relatively recent (past month or something?) refactoring of the embargo mechanism itself

In perspective, these types of live service features are going to be the main focus of maintenance going forward since we can tightly control most other package versioning

PR #831, opened 2024-06-05
Allow validating timezone with Inspector calls

garrettmflynn:
Uses https://github.com/NeurodataWithoutBorders/nwbinspector/pull/471 to properly validate timezone with Inspector checks.

### DIFF
--- a/src/electron/frontend/core/components/pages/guided-mode/data/GuidedMetadata.js
+++ b/src/electron/frontend/core/components/pages/guided-mode/data/GuidedMetadata.js
@@ -333,7 +333,7 @@ export class GuidedMetadataPage extends ManagedPage {
 
             onUpdate: () => (this.unsavedUpdates = "conversions"),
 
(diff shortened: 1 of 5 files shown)


CodyCBakerPhD:
Would you recommend this get merged before or after the parent?

garrettmflynn:
Hmm depends on your answer to this question. With any timezone awareness, we have to decide how we want to limit the user's selection of future dates—if at all.

Right now, the date selector uses the current time in local time—but this isn't timezone aware so, for instance, I would not be able to choose a future time tomorrow for a project run in Europe.

This is a general problem across both PRs. Though the consequences are limited by Ben's earlier note that data from different timezones will rarely be converted the same day.

CodyCBakerPhD:
Yeah that's really an edge case so I'll check this out in more detail tomorrow

CodyCBakerPhD (approved):

PR #833, opened 2024-06-05
Improve NWB Inspector calls

CodyCBakerPhD:
Getting this started in draft

Will replace #550 

Broader idea is to suppress various check functions for guided mode (things that are beyond user control) but not for the standalone page


CodyCBakerPhD:
Not going to have enough time to get this through; I'll add it to the compiled list though

PR #837, opened 2024-06-05
Fix Embargo

garrettmflynn:
This PR is an attempt to fix all issues relating to embargoed Dandisets.

I've been able to isolate the embargoed Dandiset issue to the fact that requesting `dandisets/?user=me` does _not_ appear to return Dandisets that are embargoed. I'll start scouring the docs for an alternative request.

Once this is complete, we can move on to #814.


garrettmflynn:
A simple fix. Turns out this request turns everything _except_ embargoed Dandisets by default. Need to add another search query (e.g. `dandisets/?user=me&embargoed=true`) to the end of it.

garrettmflynn:
@CodyCBakerPhD Aside from the NIH Award Number stuff, can you confirm this is behaving as expected?

CodyCBakerPhD:
When I try to create one one on staging I get this error

```javascript
dandi.es.js:199 Uncaught (in promise) Error: API is not authorized. Please provide a valid token.
    at K.authorize (dandi.es.js:199:15)
    at async K.init (dandi.es.js:182:7)
    at async Button.onClick [as __onClick] (UploadsPage.js:122:17)
```

But it's definitely the right credentials for staging

CodyCBakerPhD:
I also can't see any of my embargoed Dandisets on the loose selector when my API keys are set

garrettmflynn:
Sorry! Try again

CodyCBakerPhD:
OK I can now see them in the selector

One thing, not sure if you want to deal with it in this PR or a follow-up : the NIH grant number validation did not occur until I hit 'Submit' (I clicked off it and waited to see if that would trigger but it did not) and then it popped up very briefly but submission of the modal finished (would have expected the modal to stay open if there was an error) before I could fix it

However, the dandiset did get created and seems fine. Maybe it would have problems at some level without the right kind of grant ID format, but the basic feature did work

garrettmflynn:
Let's handle the NIH grant number stuff before merging this PR, as this will enable embargo again. I'll start this in a linked follow-up.

garrettmflynn:
Alright invalid NIH grant numbers are properly blocked now! I'm going to skip the auto-parsing of the non-delimited string copied off NIH Reporter because DANDI itself does not accept this format directly.

CodyCBakerPhD:
Cool

CI tests failing now - you wanted to work on NIH number formatting on a separate branch/PR that gets merged into this one before merging either to main?

garrettmflynn:
Blocking Dandiset creation when the user provides an invalid NIH grant number has been implemented here. I'm actually going to skip the number formatting since DANDI does not allow this. Would it be correct here to assume we want to maintain parity between ourselves and DANDI?

CodyCBakerPhD:
I think it's probably fine at this point, the other idea was a 'nicety'

As per meeting; we can merge this whenever you're ready but I won't cut a new release until after user days

garrettmflynn:
Sounds good to me. Will need your review here

CodyCBakerPhD (approved):

CodyCBakerPhD (approved):

Issue #840, opened 2024-06-07
Generating test dataset on Mac arm64 build fails with "No module named 'spikeinterface'"


rly:
### Describe the issue

```
index-178c0fa4.js:7278 [main-process]: [2024-06-07 09:55:16,250] ERROR in app: Exception on /data/generate [POST]
Traceback (most recent call last):
  File "flask/app.py", line 1484, in full_dispatch_request
  File "flask/app.py", line 1469, in dispatch_request
  File "flask_restx/api.py", line 404, in wrapper
  File "flask/views.py", line 109, in view
  File "flask_restx/resource.py", line 46, in dispatch_request
  File "namespaces/data.py", line 19, in post
  File "manageNeuroconv/manage_neuroconv.py", line 1637, in generate_test_data
ModuleNotFoundError: No module named 'spikeinterface'
```

### Steps to Reproduce

Install 1.0 test release

### Operating System

Mac OS with Mac M1

### GUIDE Version

1.0 test

### Code of Conduct

Yes

### Did you confirm this issue was not already reported?

Yes

rly:
Strangely a local build works on my system.

I will do some testing. Might just require adding
```
tmp_ret = collect_all('spikeinterface')
datas += tmp_ret[0]; binaries += tmp_ret[1]; hiddenimports += tmp_ret[2]
```
to `nwb-guide.spec`

rly:
Fixed by #841

PR #841, opened 2024-06-07
Use neuroconv[full] and update Mac build workflow to use x64

rly:
Attempted fix of #840.

Watching https://github.com/NeurodataWithoutBorders/nwb-guide/actions/runs/9420722499

### DIFF
--- a/environments/environment-Linux.yml
+++ b/environments/environment-Linux.yml
@@ -15,7 +15,7 @@ dependencies:
       - flask == 2.3.2
       - flask-cors == 4.0.0
       - flask_restx == 1.1.0
-      - neuroconv >= 0.4.10
+      - neuroconv[full] >= 0.4.10
       - dandi >= 0.60.0
       - pytest == 7.4.0
       - pytest-cov == 4.1.0
--- a/environments/environment-Linux.yml
+++ b/environments/environment-Linux.yml
@@ -15,7 +15,7 @@ dependencies:
       - flask == 2.3.2
       - flask-cors == 4.0.0
       - flask_restx == 1.1.0
-      - neuroconv[full] >= 0.4.10
+      - neuroconv @ git+https://github.com/catalystneuro/neuroconv.git@main#neuroconv[full]
       - dandi >= 0.60.0
       - pytest == 7.4.0
       - pytest-cov == 4.1.0
--- a/environments/environment-MAC-intel.yml
+++ b/environments/environment-MAC-intel.yml
@@ -18,7 +18,7 @@ dependencies:
       - flask == 2.3.2
       - flask-cors == 4.0.0
       - flask_restx == 1.1.0
-      - neuroconv[full] >= 0.4.10
+      - neuroconv @ git+https://github.com/catalystneuro/neuroconv.git@main#neuroconv[full]
       - dandi >= 0.60.0
       - pytest == 7.4.0
       - pytest-cov == 4.1.0
--- a/environments/environment-MAC-intel.yml
+++ b/environments/environment-MAC-intel.yml
@@ -18,7 +18,7 @@ dependencies:
       - flask == 2.3.2
       - flask-cors == 4.0.0
       - flask_restx == 1.1.0
-      - neuroconv >= 0.4.10
+      - neuroconv[full] >= 0.4.10
       - dandi >= 0.60.0
       - pytest == 7.4.0
       - pytest-cov == 4.1.0
--- a/environments/environment-Windows.yml
+++ b/environments/environment-Windows.yml
@@ -18,7 +18,7 @@ dependencies:
       - flask == 2.3.2
       - flask-cors === 3.0.10
       - flask_restx == 1.1.0
-      - neuroconv >= 0.4.10
+      - neuroconv[full] >= 0.4.10
       - dandi >= 0.60.0
       - pytest == 7.2.2
       - pytest-cov == 4.1.0
--- a/environments/environment-Windows.yml
+++ b/environments/environment-Windows.yml
@@ -18,7 +18,7 @@ dependencies:
       - flask == 2.3.2
       - flask-cors === 3.0.10
       - flask_restx == 1.1.0
-      - neuroconv[full] >= 0.4.10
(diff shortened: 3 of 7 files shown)


rly:
That change did not work. A recent update to the env installed the latest `neuroconv` instead of `neuroconv[full]`. `[full]` should install spikeinterface and other packages. Let's try that.

CodyCBakerPhD (approved):

rly:
I also set debugging environment variables in the Mac release workflow to help debug build & deploy issues like when electron-builder reports:
```
Above command failed, retrying 5 more times
Above command failed, retrying 4 more times
Above command failed, retrying 3 more times
```
and is doing something for 5-10 minutes without output...

It may have to do with an issue running `hdiutil` on Macs: https://github.com/electron-userland/electron-builder/issues/4606

rly:
I switched the Mac release workflow to use macos-13 because this error started to appear after the GitHub Actions macos-latest switched to point to an M1 runner: 
```
PyInstaller.utils.osx.IncompatibleBinaryArchError: /Users/runner/miniconda3/envs/nwb-guide/lib/python3.9/site-packages/sonpy/darwin/sonpy.so is incompatible with target arch arm64 (has arch: x86_64)!
```

After I updated my environment locally, I also started to get this error when I could totally build GUIDE before! Maybe I had removed this file from my environment previously. Or maybe something changed with how sonpy was getting installed. I'm not sure... 

In any case, building on GItHub Actions with macos-13 works. And building locally with these files:
```
$CONDA_BASE_DIR/envs/nwb-guide/lib/python3.9/site-packages/sonpy/darwin/sonpy.so
$CONDA_BASE_DIR/envs/nwb-guide/lib/python3.9/site-packages/sonpy/linux/sonpy.so
```
removed also works, though I suspect that reading from Spike2 files would not work without these files.

rly:
Notes to self:
- The build tests and release workflow using the macos-latest (arm64) runner [work when the environment installs neuroconv from github](https://github.com/NeurodataWithoutBorders/nwb-guide/actions/runs/9427191709/job/25971002796?pr=841) but not [when installing the latest neuroconv from pypi](https://github.com/NeurodataWithoutBorders/nwb-guide/actions/runs/9426458817/job/25969404079?pr=841). The neuroconv package should be the same except neuroconv from github has a version bump which should not matter. The error is:
  ```
  PyInstaller.utils.osx.IncompatibleBinaryArchError: /Users/runner/miniconda3/envs/nwb-guide/lib/python3.9/site-packages/sonpy/darwin/sonpy.so is incompatible with target arch arm64 (has arch: x86_64)!
  ```
  - When installing from github, there is no directory `$CONDA_PREFIX/lib/python3.9/site-packages/sonpy` and no directory "sonpy" anywhere in the conda environment.
  - When installing from pypi, `$CONDA_PREFIX/lib/python3.9/site-packages/sonpy` exists and the problematic `.so` files exist. How strange.
  - Then I remembered that we added this code to `neuroconv/setup.py`: https://github.com/catalystneuro/neuroconv/blob/main/setup.py#L57-L64
  - I believe this if statement is entered when building the wheel and running `setup.py` on my Mac M1, but if the wheel is already built, and was not built on a Mac M1, then this if statement was not entered and sonpy is included in the wheel.

Proposed solution:
- When running tests on Mac arm64, install neuroconv from github. This will pull the latest from the main branch. Tests should pass because sonpy will not be installed.
- When building and deploying, use a Mac x64 runner for now so that Spike2RecordingInterface is available to Mac x64 users. sonpy will be installed and will be compatible with the x64 OS. I do not know whether the arm64 app built with the x64 .so file can use the Spike2RecordingInterface. If we want to switch to a Mac arm64 runner to build both the x64 and arm64 apps, then we have to install neuroconv from github and Spike2RecordingInterface would not be available on both x64 and arm64. Alternatively, we could have separate workflows build the x64 and arm64 apps.

environments/environment-Linux.yml
```diff
@@ -15,7 +15,7 @@ dependencies:
       - flask == 2.3.2
       - flask-cors == 4.0.0
       - flask_restx == 1.1.0
-      - neuroconv >= 0.4.10
+      - neuroconv[full] >= 0.4.10
```
CodyCBakerPhD, line 18:
```suggestion
      - neuroconv @ git+https://github.com/catalystneuro/neuroconv.git@main#neuroconv[full]
```

environments/environment-MAC-intel.yml
```diff
@@ -18,7 +18,7 @@ dependencies:
       - flask == 2.3.2
       - flask-cors == 4.0.0
       - flask_restx == 1.1.0
-      - neuroconv >= 0.4.10
+      - neuroconv[full] >= 0.4.10
```
CodyCBakerPhD, line 21:
```suggestion
      - neuroconv @ git+https://github.com/catalystneuro/neuroconv.git@main#neuroconv[full]
```

environments/environment-Windows.yml
```diff
@@ -18,7 +18,7 @@ dependencies:
       - flask == 2.3.2
       - flask-cors === 3.0.10
       - flask_restx == 1.1.0
-      - neuroconv >= 0.4.10
+      - neuroconv[full] >= 0.4.10
```
CodyCBakerPhD, line 21:
```suggestion
      - neuroconv @ git+https://github.com/catalystneuro/neuroconv.git@main#neuroconv[full]
```

CodyCBakerPhD:
I'm honestly fine with just reverting all environments to use the git install path for NeuroConv, I didn't expect it to cause this many problems

nwb-guide.spec
```diff
@@ -16,7 +16,6 @@ hiddenimports = [
     'email_validator',
     *collect_submodules('scipy.special.cython_special'),
     *collect_submodules('scipy.special._cdflib'),
-    *os.path.join(os.path.dirname(scipy.__file__), '.libs')
```
CodyCBakerPhD, line 19: Should we restore this too?

rly: No, this resulted in failures to import "/", then "U", then "s", "e", "r", "s", ... because of the asterisk on a string path. Given that this line didn't work in the first place, I think we can remove it.

PR #851, opened 2024-06-10
Attempt to suppress reroute of switcher JSON for dev

CodyCBakerPhD:
Trying to fix the switcher when using `latest` (dev)


CodyCBakerPhD:
It _looks_ like it might work on https://nwb-guide--851.org.readthedocs.build/en/851/ but not sure if that will apply to `main` / 
 `latest`

WDYT?

rly (approved):

rly:
Worth a try!

Issue #879, opened 2024-06-24
MEARec not working in prod


bendichter:
### Describe the issue

<img width="1273" alt="image" src="https://github.com/NeurodataWithoutBorders/nwb-guide/assets/844306/9d53563d-8df2-4e8a-8d1a-cc2f5345654e">

### Steps to Reproduce

Try to convert from our example data

### Operating System

osx

### GUIDE Version

1.0.1

### Code of Conduct

Yes

### Did you confirm this issue was not already reported?

Yes

CodyCBakerPhD:
closed by #880

PR #880, opened 2024-06-24
Fix MEArec build

CodyCBakerPhD:
#879 

Added test case to pipelines

Ensured it works on local build


bendichter:
@CodyCBakerPhD would this work?

```python
for pkg in (
    'dandi',
    'keyrings',
    'unittest',
    'nwbinspector',
    'neuroconv',
    'pynwb',
    'hdmf',
    'hdmf_zarr',
    'ndx_dandi_icephys',
    'sklearn',
    'ci_info',
    'tifffile',
    'dlc2nwb',
    'sleap_io',
    'ndx_pose',
    'tzdata',
    'elephant',
):
    tmp_ret = collect_all(pkg)
    datas += tmp_ret[0]; binaries += tmp_ret[1]; hiddenimports += tmp_ret[2]
```

CodyCBakerPhD:
It should, went ahead and enhanced that

The only thing is if you ever regenerate the spec file via PyInstaller CLI, it'll get rid of that (the flat version is what is auto-compiled by it)

PR #886, opened 2024-06-27
Update switcher for 1.0.2

rly:
Before release of 1.0.2

### DIFF
--- a/docs/_static/switcher.json
+++ b/docs/_static/switcher.json
@@ -5,6 +5,12 @@
         "preferred": false
     },
     {
+        "name": "1.0.2",
+        "version": "v1.0.2",
+        "url": "https://nwb-guide.readthedocs.io/en/v1.0.2/",
+        "preferred": true
+    },
+    {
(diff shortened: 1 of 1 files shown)


CodyCBakerPhD:
Exactly my suspicion as well

Need to change https://github.com/NeurodataWithoutBorders/nwb-guide/pull/886/files#diff-4e8e667a6b53889132b09d57fb7cdb38aa0d577fd5a7406592b2d6fdfaf850eeR17 to `false` since the switcher can only have one `preferred` at a time ([source](https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/version-dropdown.html#add-a-json-file-to-define-your-switcher-s-versions))

However, we would only swap the preference immediately before release, WDYT?

rly:
Yes, we should swap immediately before release. I was going to propose a release ahead of the user test which was cancelled yesterday. The next one is Tuesday. What do you think about merging this and making a release today / this weekend?

PR #889, opened 2024-07-02
Escape HTML tags in error messages on Source Data Information screen

dhmjhu:
# Summary
On the Source Data Information screen, not all parts of error messages returned by the server are properly escaped.

# Example
Trying to select a Kilosort directory with an invalid `params.py` resulted in the following message:
> **Request Failed**
> invalid syntax (, line 3)

Screenshot of error message

The actual response from the server was:
```json
{"message": "invalid syntax (<string>, line 3)", "type": "SyntaxError"}
```

# Cause
While angle brackets in `text` are converted to HTML entities, `type` is left alone.
https://github.com/NeurodataWithoutBorders/nwb-guide/blob/29599c137dce1b49c9fdcb149c204cc88db7e7cf/src/electron/frontend/core/components/pages/guided-mode/data/GuidedSourceData.js#L123-L130

# Fix
Performing the same replacements on `type` should resolve this.

I have not tested this change, as I don't have an appropriate environment set up right now.

### DIFF
--- a/src/electron/frontend/core/components/pages/guided-mode/data/GuidedSourceData.js
+++ b/src/electron/frontend/core/components/pages/guided-mode/data/GuidedSourceData.js
@@ -121,13 +121,14 @@ export class GuidedSourceDataPage extends ManagedPage {
 
                     if (result.message) {
                         const [type, ...splitText] = result.message.split(":");
+                        const escapedType = type.replaceAll("<", "&lt").replaceAll(">", "&gt");
                         const text = splitText.length
                             ? splitText.join(":").replaceAll("<", "&lt").replaceAll(">", "&gt")
                             : result.traceback
                               ? `<small><pre>${result.traceback.trim().split("\n").slice(-2)[0].trim()}</pre></small>`
                               : "";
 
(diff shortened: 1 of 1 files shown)


CodyCBakerPhD:
Thanks for catching this and pushing a fix!

> I have not tested this change, as I don't have an appropriate environment set up right now.

Please give it a try and let us know (preferable with screenshot of new error to match expectation), since I don't have a version of the data with invalid `params` myself

The developer installation instructions can be found here: https://nwb-guide.readthedocs.io/en/stable/developer_guide.html#installation

rly:
Thanks @dhmjhu for the PR and minimal example! I was able to demonstrate the fix:

image

rly (approved):

CodyCBakerPhD:
@rly Looks like you should generate and set a new token for chromatic

CodyCBakerPhD:
I would propose merging this to a copy of current main (`staging`) then merging that into `main` after chromatic checks pass

I've adjusted the target of this PR

rly:
Good idea. Let's do that.

PR #898, opened 2024-07-29
Try updating lock file

CodyCBakerPhD:
To fix installation issues without suppressing post-install scripts (which are necessary for app to launch)


CodyCBakerPhD:
That's a no as well

PR #900, opened 2024-07-29
Disable mac on all tests

CodyCBakerPhD:
What is odd is that the app seems to work fine on Mac (arm64 at least) locally, it's just in CI where the problem is seen...

Propose disabling this until an `npm` expert can dig deeper


CodyCBakerPhD:
Warning: I did just now observe the same stall on Ubuntu (https://github.com/NeurodataWithoutBorders/nwb-guide/actions/runs/10139908661/attempts/1?pr=900) but that is very random. Retrying with a fresh cache seemed to fix the problem.

PR #904, opened 2024-07-30
Fix missing asterisk for dandiset license

### LINKED ISSUE #903: License option on "Create Dandiset" form missing asterisk to indicate it is required
### Describe the issue

image

The description also suggests you can select a **set** of licenses when you can only select one here. The [dandi schema suggests](https://github.com/dandi/dandi-schema/blob/be57e138505819a46475f8f5a590e6f6b00cd1b2/dandischema/models.py#L1503-L1509) you can provide more than one but I think 1) most users would not do that and 2) it would be super confusing given these two options. So we should continue allowing only one.

### Steps to Reproduce

Create a dandiset with UI

### Operating System

Mac OS with Mac M1

### GUIDE Version

main

### Code of Conduct

Yes

Yes

rly:
Fix #903

### DIFF
--- a/src/schemas/json/dandi/create.json
+++ b/src/schemas/json/dandi/create.json
@@ -37,29 +37,24 @@
         },
 
         "license": {
-            "type": "array",
-            "description": "Provide a set of licenses for this Dandiset. Review the individual licenses and select the one that best fits your needs.",
-            "items": {
-                "type": "string",
-                "enumLinks": {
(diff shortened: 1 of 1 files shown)


rly:
The current schema set up allows for a multi-select but then says "maxItems: 1" which counteracts that. If we are only going to allow one license, then we do not need to set up the schema as a multi-select.

CodyCBakerPhD:
LGTM - thanks

CodyCBakerPhD (approved):

PR #906, opened 2024-08-04
Simplify backend dependencies

CodyCBakerPhD:
Build tests just started failing due to a new hidden library in `matplotlib`, which we shouldn't even need as far as I know to run GUIDE (I tried removing locally and ran through a bunch of pipelines, including tutorial generation, and everything worked fine)

I think `spikeinterface[qualitymetrics]` in `neuroconv[testing]` (a part of `neuroconv[full]`) might be the source of that, so testing if this simplifies things

In the same spirit, there is no reason to bundle `pytest` or `pytest-cov` with our actual official build releases, so I'm suggesting only installing it in the CI where it is needed to run tests

### DIFF
--- a/.github/workflows/testing_dev.yml
+++ b/.github/workflows/testing_dev.yml
@@ -74,6 +74,9 @@ jobs:
       - name: Install GUIDE
         run: npm ci --verbose
 
+      - name: Install testing dependencies
+        run: pip install pytest pytest-cov
+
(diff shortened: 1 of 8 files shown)


CodyCBakerPhD:
Well I still have no idea where `matplotlib` is being installed from out of that, but I still think trimming the testing requirements out of package setup for GUIDE is in our best interest for everything outside of CI

rly (approved):

PR #910, opened 2024-08-13
Update puppeteer, re-enable mac CI, split e2e CI

rly:
Replace #902


behavior_testing_data
CodyCBakerPhD, line 1: What are these submodules? I can't follow them anywhere

package.json
```diff
@@ -156,7 +156,7 @@
     "lit": "^2.6.1",
     "lottie-web": "^5.9.5",
     "notyf": "^3.9.0",
-    "puppeteer": "^21.7.0",
+    "puppeteer": "^22.15.0",
```
CodyCBakerPhD, line 159: So was this the magical fix after all? Or they just fixed something on the puppeteer side?

.github/workflows/e2e_tests_on_pull_request.yml
CodyCBakerPhD, line 1: Can you recombine this with the main deployment again since it doesn't look like anything changed with how they are called?

The suggestion to peel out e2e step was on the lower workflow level, so that 'dev tests' only run vitetest and pytest, and puppeteer calls on pipelines with example data could be a separate `npm run` call

behavior_testing_data
rly: Oh! That was not intentional.

.github/workflows/e2e_tests_on_pull_request.yml
rly: Ohhh I see.

package.json
rly: Yes. Puppeteer calls a live service to download chrome. They updated that live service in some way such that older versions crash / stall when calling it.

behavior_testing_data
CodyCBakerPhD: That said, did you actually make some GitHub repositories with our example testing data? And maybe git LFS? Because we've all wanted that for a really long time lol

.github/workflows/e2e_tests_on_pull_request.yml
CodyCBakerPhD: Just something to keep in mind if this happens again. Fingers crossed it doesn't though

behavior_testing_data
rly: I guess I accidentally added the testing data git repos as submodules. How would you want these repos/submodules to work?

CodyCBakerPhD: > I guess I accidentally added the testing data git repos as submodules.

But that's what I'm confused about - there aren't any git repos currently (well, technically git-annex via GIN)

> How would you want these repos/submodules to work?

Identically to the GIN repos, but better in all the ways that GitHub is better than GIN as a platform (and likely using Git LFS for any files that are larger than a few MB)

CodyCBakerPhD (approved):

PR #918, opened 2024-08-23
Update developer guide, removed unused code and CSS

rly:
In reviewing the code base, I updated the developer guide and removed code and CSS that was used in SODA but not in GUIDE. In particular, the Guided Home Page manages uploaded vs non-uploaded datasets and has support for many tabs that are also radio buttons, but GUIDE does not track uploaded vs non-uploaded and has just the one tab "Existing Conversions". So I simplified the code. We can always reference SODA if we want to use any of those features, but I do not expect that we will. 

There is still a lot more CSS left that is unused; I just removed what stood out as immediately removable after a search through the repo. I also cross-checked most of them with the SODA repo to make sure they were used in custom elements in their app (as opposed to an imported component).


CodyCBakerPhD (approved):

CodyCBakerPhD (changes requested):
Actually looks like Windows e2e live tests are failing due to a frontend thing? Are you sure everything removed here is unused?

rly:
Hm. I will investigate.

CodyCBakerPhD:
@rly Nvm, it's failing on `main` as well. Must be something else

CodyCBakerPhD (approved):

PR #920, opened 2024-08-23
Update notification email

CodyCBakerPhD:
I don't mind having mine exposed instead of as a secret since it makes it easier to debug

Didn't notice dailies were failing before


.github/workflows/daily_tests.yml
```diff
@@ -53,6 +53,6 @@ jobs:
           username: ${{ secrets.MAIL_USERNAME }}
           password: ${{ secrets.MAIL_PASSWORD }}
           subject: NWB GUIDE Daily Test Failure
-          to: ${{ secrets.DAILY_FAILURE_EMAIL_LIST }}
+          to: cody.c.baker.phd@gmail.com
```
rly, line 56:
```suggestion
          to: cody.c.baker.phd@gmail.com, rly@lbl.gov
```
Can we add mine as well?

CodyCBakerPhD: Of course!

rly (approved):

PR #923, opened 2024-08-31
Debug DANDI upload argument defaults

### LINKED ISSUE #922: Error uploading data
### Describe the issue

image

### Steps to Reproduce

Try to upload any NWB files

### Operating System

OSX

### GUIDE Version

1.0.3

### Code of Conduct

Yes

Yes

CodyCBakerPhD:
fix #922 

Not sure when or how this changed

I tried it out with `v1.0.0` of the GUIDE and this issue did not exist at that time

But in latest release, the values being sent back from the frontend were `null` (even if they visually displayed as the default values of `1`), causing `number_of_jobs` to set to `None` from the top level of all Python argument passing, which for parallelization usually means 'use all jobs'

So I was able to replicate the issue locally and then tested that this fix resolves the issue on a local build

@rly and I discussed having a pre-release TODO checklist (including RTD switcher bumps) - this would be a good thing to add to that list


src/pyflask/manageNeuroconv/manage_neuroconv.py
```diff
@@ -1277,7 +1277,9 @@ def upload_multiple_filesystem_objects_to_dandi(**kwargs) -> list[Path]:
     del innerKwargs["filesystem_paths"]
     innerKwargs["nwb_folder_path"] = tmp_folder_path
     results = upload_folder_to_dandi(**innerKwargs)
-    rmtree(tmp_folder_path)
+
+    rmtree(tmp_folder_path, ignore_errors=True)
```
CodyCBakerPhD, line 1281: Also fixed this, which can happen if you enable 'cleanup' option but cleanup succeeds in earlier stage (earlier stage is known to fail stochastically on certain platforms, which is why we try it again here)

bendichter (approved):

Issue #941, opened 2024-10-31
Daily ExampleData tests failing


rly:
### Describe the issue

Daily ExampleData tests failing have failed 3 days in a row for the same reasons.

https://github.com/NeurodataWithoutBorders/nwb-guide/actions/runs/11613365359/job/32347559180
macos-latest: `ModuleNotFoundError: No module named 'ndx_pose'`
macos-13: `ModuleNotFoundError: No module named 'tables'`
windows-latest: `ModuleNotFoundError: No module named 'tables'`

### Steps to Reproduce

See GitHub actions logs

### Operating System

Mac, Windows

### GUIDE Version

main

### Code of Conduct

Yes

### Did you confirm this issue was not already reported?

Yes

PR #942, opened 2024-11-04
Quick fix for werkzeug/flask incompatibility

rly:
Most GitHub Actions are failing because flask depends on werkzeug, and werkzeug 3.0+ is being installed which deprecates the `__version__` attribute used by flask 2.3.2. We could upgrade flask to 3.0+, but that is a major change that will require some time to carefully test. While that happens, I suggest we pin werkzeug < 3.0.


pauladkisson (approved):
Example Data Tests are failing still, but looks like that's a separate issue (#941). So this LGTM.

PR #944, opened 2024-11-12
Fix setting dandiset license to array

### LINKED ISSUE #943: license metadata populated incorrectly
### Describe the issue

When you create a new dandiset, the license metadata is created incorrectly

image

It should be sent as a list with one element, but it is sent as a string, so the API instead breaks up the string and has a license for each letter.

### Steps to Reproduce

Just go through the GUIDE tutorial and create a new dandiset on staging

### Operating System

OSX

### GUIDE Version

1.0.4

### Code of Conduct

Yes

Yes

rly:
Fix #943


rly:
Confirmed that this is fixed in local build:
https://gui-staging.dandiarchive.org/dandiset/215660

image

pauladkisson (approved):

PR #917, opened 2024-08-21
port over other PR

CodyCBakerPhD:
replace #916 

GitHub was being slow to change the merge target (took like 10 minutes to allow merge) so in less time than that I just ported over the essential changes

Let's see what CI says

PR #952, opened 2025-01-20
neuroconv update - WIP

luiztauffer:


rly:
Thanks @luiztauffer . Please see also WIP in #917 in case that is useful.

rly:
On further review, I believe this can be replaced by #917. We will not be able to update directly to NeuroConv 0.6.7 just yet. See #956. We'll do it iteratively.

PR #975, opened 2025-03-07
Clear Neurosift storage when rendering Neurosift

### LINKED ISSUE #974: Clear cache when serving NWB files to Neurosift and other services
### Describe the issue

A user I am working with was iteratively creating NWB files and using NWB GUIDE > Neurosift to visualize the file. The NWB file is only ~75 MB. The user would modify the file but the changes were not reflected in NWB GUIDE > Neurosift. The file was indeed changed as confirmed by PyNWB. It turns out that Electron has a cache and something, probably the NWB file, is being cached and served when the file with the same path (but different create/modify date and size) is requested.

See https://stackoverflow.com/a/44675132/20177

Deleting `/Users/<user>/Library/Application Support/NWB GUIDE/Cache` was insufficient, but deleting `/Users/<user>/Library/Application Support/NWB GUIDE/` worked.

It might be that if the file is much larger, it would not be cached.

GUIDE should clear the Electron cache (or not use the cache) when serving NWB files to Neurosift and other services like DANDI Upload.

### Steps to Reproduce

^

### Operating System

Mac

### GUIDE Version

1.0.4

### Code of Conduct

Yes

Yes

rly:
Fix #974

I tested this locally. The neurosift storage data is cleared within the same session and across sessions whenever a new file is rendered.


rly:
Good idea. With help from Claude Sonnet, I added a unit/component test that tests whether the `deleteDatabase` method is called when the neurosift component is updated with a URL set.

I also converted `Neurosift.js` to a typescript file to resolve a linter warning about importing a js file into a ts file in the Neurosift component test

pauladkisson (approved):
Looks good!

PR #986, opened 2025-05-14
Use random endpoint for neurosift files to fix incorrect caching

### LINKED ISSUE #974: Clear cache when serving NWB files to Neurosift and other services

rly:
Fix #974

The new endpoint is `/files/<string:file_id>` where file_id is a string with 8 random letters and numbers, which is highly unlikely to repeat before the Neurosift cache resets (currently after 1000 entries)

See comments in #974 for rationale of using a random string instead of other schemes.

Clearing the Neurosift cache database doesn't work from the GUIDE domain so I removed that code and test.


pauladkisson (commented):
Random string seems less transparent, what about using an index + date modified instead?

rly:
Great idea! I made that change.

pauladkisson (approved):
Not sure why the e2eLiveServices tests are failing, but otherwise looks good.

rly:
They are failing due to #987 . I'll address that separately.

Issue #957, opened 2025-02-01
Support NeuroConv 0.6.0 and 0.6.1: https://github.com/catalystneuro/neuroconv/releases/tag/v0.6.0 https://github.com/catalystneuro/neuroconv/releases/tag/v0.6.1


rly:

rly:
TODO:
- [ ] Add and test MedPCInterface
- [ ] Add and test TDTFiberPhotometryInterface
- [ ] Test Plexon2

rly:
The MedPC Interface is not so straightforward to implement (https://github.com/NeurodataWithoutBorders/nwb-guide/issues/998) so I propose we punt on it for now.

rly:
Plexon2 support will be added when NeuroConv 0.6.5 support is added due to issues with the PL2 interface in NeuroConv 0.6.0 and challenges installing NeuroConv 0.6.2 to 0.6.4 properly. (#1001)

PR #997, opened 2025-07-01
Handle new interface format keywords from NeuroConv 0.6.0

rly:
Working on #957. 

Address https://github.com/NeurodataWithoutBorders/nwb-guide/issues/957#issuecomment-3022092942

- Generates JSON Schema for interfaces using NeuroConv 0.6.0 which updates `format: file` -> `format: file-path` and `format: directory` -> `format: directory-path` among other changes.
- Updates GUIDE code to detect `file-path` and `directory-path` instead of `file` and `directory` for displaying a file selector vs directory selector
- Updates internal GUIDE code to use `format: file-path` and `format: directory-path`. I think I caught all the instances.

Alternatively, I could have converted the JSON Schema on read to change `file-path` back to `file` and not change the rest of GUIDE, but 1) I think `file-path` is more informative, and 2) that conversion feels less clean.

Other changes in the interfaces will be addressed in a separate PR to keep this one clean.


stories/inputs/interface_schemas/AlphaOmegaRecordingInterface.json
```diff
@@ -2,24 +2,24 @@
...
                 "folder_path": {
-                    "format": "directory",
+                    "format": "directory-path",
                     "type": "string",
                     "description": "Path to the folder of .mpx files."
                 },
                 "verbose": {
-                    "type": "boolean",
-                    "default": true
+                    "default": true,
```
pauladkisson, line 12: Is there a reason these need to be switched?

pauladkisson (approved):
Looks good!

stories/inputs/interface_schemas/AlphaOmegaRecordingInterface.json
rly: I'm not sure why the ordering changed. I just regenerated all of the JSON Schema - NeuroConv (or an updated jsonschema package) made these changes.

PR #1001, opened 2025-07-02
Restrict Plexon2RecordingInterface to Windows

### LINKED ISSUE #915: [Feature]: Support reading pl2

We should allow this only on Windows computers. NeuroConv can support conversion on linux and mac, but it requires wine, which is too big for this app. 

blocked by #914

Yes.

### Code of Conduct

Yes

Yes

rly:
Fix #915

TODO:
- Restrict usage to Windows distributions


rly:
Installing neuroconv 0.6.2 from pypi does not install its extras (see https://github.com/catalystneuro/neuroconv/issues/1116). This is fixed in neuroconv 0.6.5. So I think it would be best to punt PL2 support to when neuroconv 0.6.5 support is added.

bendichter (approved):

PR #1057, opened 2026-02-13
Download pre-generated tutorial data instead of generating it

bendichter:
## Problem
Tutorial test data generation runs every E2E CI run, spending ~2-3 minutes fitting PCA across 50 units × 385 channels. This is unnecessary since the data is deterministic (seeded).

## Solution
Pre-generated tutorial data is now hosted as GitHub release assets and downloaded instead of generated:
- **Single-session data** (SpikeGLX + Phy): 36MB compressed
- **Multi-session dataset** (2 subjects × 2 sessions): 113MB compressed

### Changes
1. **Backend**: New `download_test_data()` and `download_test_dataset()` functions + `/data/download` and `/data/download/dataset` API endpoints
2. **Frontend**: App tries downloading first, falls back to generation if offline or download fails
3. **CI**: `ExampleDataCache` workflow caches tutorial data from GitHub release; E2E workflow restores it before tests
4. **Release**: [tutorial-test-data-v1](https://github.com/NeurodataWithoutBorders/nwb-guide/releases/tag/tutorial-test-data-v1) hosts the compressed archives

### Benefits
- E2E tests skip data generation entirely (data already present)
- App users get tutorial data in seconds instead of minutes
- Generation code still works as fallback (no breaking change)
- Data is versioned via release tags — bump tag when generation code changes


rly:
Tutorial generation now takes 13 seconds on my Mac M1. Do we still want to cache the tutorial data and use that in tests?

PR #1080, opened 2026-07-08
Fix daily CI failures from spikeinterface/scipy dependency drift

rly:
## Problem

The **Daily Tests** workflow had been failing every day, and the `Deploy` (PR) workflow had failed on every PR since mid-March 2026. The failures trace to dependency drift in `spikeinterface`, its transitive `scipy`, and `pandas`.

### 1. `DevTests` / `BuildTests`: spikeinterface low `freq_min` guard

Recent `spikeinterface` added a `highpass_check` that raises `ValueError` when `bandpass_filter` receives a `freq_min` below its threshold:

```
ValueError: The freq_min (0.5 Hz) is too low ... You can set 'ignore_low_freq_error=True' to bypass this error...
```

`generate_test_data` intentionally uses `freq_min=0.5` Hz for the synthetic LFP band, so it now passes `ignore_low_freq_error=True`.

### 2. `BuildTests`: scipy `array_api_compat` moved, plus a scipy floor

`scipy` relocated `array_api_compat` from `scipy._lib` to `scipy._external` in **scipy 1.18.0**. The PyInstaller spec pinned the old path, so the built executable crashed with:

```
ModuleNotFoundError: No module named 'scipy._external.array_api_compat.numpy.fft'
```

The spec now bundles `scipy._external.array_api_compat.numpy.fft`, and the environment files floor `scipy >= 1.18.0` so the bundled path is guaranteed to exist (spikeinterface's own scipy floor is far below this).

### 3. spikeinterface version floor

`spikeinterface` was only a transitive dependency of `neuroconv` (which requires `>= 0.103.0`), so the installed version drifted. All four environment files now floor it at `>= 0.104.8`, the version providing `ignore_low_freq_error`. A floor (not an exact pin) keeps the daily tests able to catch future drift.

### 4. `E2ELiveServices`: pandas 3.0 read-only array in the Phy extractor

`POST /neuroconv/metadata` returned 500 when instantiating a `PhySortingInterface`:

```
spikeinterface/extractors/phykilosortextractors.py, in __init__
    unit_ids[i] = new_si_id
ValueError: assignment destination is read-only
```

**pandas 3.0** makes Copy-on-Write permanent, so `DataFrame.values` returns a read-only array, and spikeinterface's Phy extractor assigns into it in place. The environment files now cap `pandas < 3` until spikeinterface stops mutating the read-only array. Reported upstream: SpikeInterface/spikeinterface#4687.

## Verification

- `pytest src/pyflask/tests/` passes (previously `test_generate_test_data` failed).
- All `Deploy` CI jobs are green across ubuntu, macOS (arm64 + intel), and Windows: `DevTests`, `BuildTests`, `LiveServices`, `E2ELiveServices`, and `ExampleDataTests`.
- The pandas fix resolved the entire `E2ELiveServices` failure: the earlier frontend `TypeError`s (e.g. reading `form` on an undefined modal) were downstream of the metadata 500, not a separate issue.

### DIFF
--- a/environments/environment-Linux.yml
+++ b/environments/environment-Linux.yml
@@ -16,7 +16,7 @@ dependencies:
       - flask_restx == 1.1.0
       - werkzeug < 3.0 # werkzeug 3.0 deprecates features used by flask 2.3.2. Remove this when updating flask.
       - neuroconv[dandi,compressors,ecephys,ophys,behavior,text] == 0.9.3
-      - spikeinterface == 0.104.8 # Pin the ecephys backend (transitive dependency of neuroconv) used by tutorial data generation
+      - spikeinterface >= 0.104.8 # ecephys backend (transitive dependency of neuroconv); 0.104.8+ supports ignore_low_freq_error used by tutorial data generation
       - scikit-learn == 1.6.1 # Tutorial data generation
       - tqdm_publisher == 0.1.1 # Progress bars
       - tzlocal == 5.3.1 # Frontend timezone handling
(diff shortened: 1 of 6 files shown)


pauladkisson (approved):
Looks good to me.

PR #1082, opened 2026-07-13
Migrate Python environment management from conda to uv

rly:
## Summary

Replaces conda with [uv](https://docs.astral.sh/uv/) for the Python environment. The four `environments/environment-*.yml` files become a single uv project (`pyproject.toml` + committed `uv.lock`), with the interpreter pinned via `.python-version` (3.12.9). The conda-forge binaries (`numpy`, `lxml`, `pyedflib`, `pytables`) now install as PyPI wheels, which exist for all supported platforms.

## Changes

- **`pyproject.toml`**: runtime deps under `[project]`; build/test tooling (PyInstaller, pytest) in a `dev` group; version `dynamic` so `package.json` stays authoritative. Transitive workarounds as uv `constraint-dependencies`: `spikeinterface >= 0.104.8`, `scipy >= 1.18.0`, `pandas < 3` (mirroring #1080's env-file floors), and `ruamel.yaml.clib != 0.2.13` (Intel-mac build). `pandas < 3` avoids a read-only `DataFrame.values` crash in spikeinterface's Phy extractor (upstream SpikeInterface/spikeinterface#4687).
- **CI/deploy workflows** (7): `setup-miniconda` -> `astral-sh/setup-uv` + `uv sync --locked` + `.venv` on `PATH`; run shell `bash -l {0}` -> `bash` (login shell ran macOS `path_helper`, which shadowed the venv's `python`); `npm` downloads cached via `cache: 'npm'`; `matplotlib` removal made resilient and applied in deploy builds too. Node.js still via `actions/setup-node`.
- **`docs/developer_guide.rst`**: uv-based setup + Node.js prerequisite.
- **`.gitignore`**: `.venv/`, `*.egg-info/`.
- Deleted the four `environments/environment-*.yml` files.

## Intel E2E flakiness

`E2ELiveServices / macos-15-intel` is intermittently red due to Chromium/Electron subprocess crashes on that runner (pre-existing, fails on the conda #1080 branch too; not a migration regression). Mitigated by wrapping only that runner's E2E in `nick-fields/retry` (5 attempts), pinned to the v4.0.0 commit.

## Validation

Verified on **macOS arm64**: `uv sync --locked`, pyflask tests (5/5), `npm run build:flask`, and the bundled executable boots and loads all 72 neuroconv interfaces. Windows, Linux, and Intel-macOS are validated by CI.

## Notes for reviewers

- `flask-cors` unified to `4.0.0` (Windows conda env used `3.0.10`, an apparent accidental divergence).
- Transitive scientific packages are now frozen by `uv.lock`; a follow-up can deliberately bump the stack.


.github/workflows/build_and_deploy_mac.yml
```diff
@@ -8,33 +8,44 @@ on:
...
       - name: Use Node.js 20
         uses: actions/setup-node@v6
         with:
           node-version: 20
+          cache: 'npm'
 
       - name: Install package.json modules and their dependencies (not via `npm ci`)
         run: npm install --verbose
 
+      # Keep the released bundle consistent with the one exercised by the build tests.
```
pauladkisson, line 45: Why do you have to manually remove Matplotlib?

.github/workflows/testing_dev_e2e_with_live_services.yml
```diff
@@ -17,70 +17,67 @@ jobs:
...
         run: |
           touch .env
           echo DANDI_SANDBOX_API_KEY=${{ secrets.DANDI_SANDBOX_API_KEY }} >> .env
 
-      - if: matrix.os != 'ubuntu-latest'
+      # The macOS x64 runner intermittently crashes Chromium/Electron subprocesses (GPU and
+      # network-service), which fails the E2E run. Retry only on that runner.
+      - if: matrix.os == 'macos-15-intel'
+        name: Run tests (retried on the flaky x64 runner)
+        uses: nick-fields/retry@ad984534de44a9489a53aefd81eb77f87c70dc60 # v4.0.0
```
pauladkisson, line 78: What does this job do?

pauladkisson (commented):
Looks good, just some clarifying comments.

.github/workflows/testing_dev_e2e_with_live_services.yml
pauladkisson: And also, why can't you just `@v4`?

.github/workflows/build_and_deploy_mac.yml
rly: Matplotlib gets pulled in from other dependencies and something in its hidden libraries breaks the packaged build, so we remove it from the build, and to keep the tests consistent, we remove it from tests as well. I added a comment about this to all workflows that do it.

.github/workflows/testing_dev_e2e_with_live_services.yml
rly: The job retries the below command (`npm run coverage:tutorial`) if it fails, with 5 max attempts and a timeout of 25 minutes, in case it gets stuck.

Recent best practices for github actions, especially non-official actions, are to pin a specific commit instead of the mutable version tag.

Last year, a popular GitHub Action `tj-actions/changed-files` was compromised and most release tags were updated to point to a new commit with code that leaked secrets. https://www.stepsecurity.io/blog/harden-runner-detection-tj-actions-changed-files-action-is-compromised.

To prevent these attacks, GitHub now supports ["immutable releases"](https://docs.github.com/en/code-security/concepts/supply-chain-security/immutable-releases), but adoption has been slow across the actions community. Surprisingly, not even the github-maintained core "actions/checkout" action is immutable: https://github.com/actions/checkout/issues/2316 . The setup-uv action has started to make [immutable releases](https://github.com/astral-sh/setup-uv/releases), and so pinning to a version tag there is fine (note that they stopped publishing "floating tags" (v4, v4.0) for similar reasons). Until actions are immutable, pinning to a specific release commit gives the highest security possible, at the costs of the workflow being less readable and the action not auto-updated. I think in a year, actions will be immutable, and we can return to tag-based pointers.

I did not update the action pins in all the workflows in this PR, just the changed ones. The others ones will be updated in a separate PR.

pauladkisson: Huh, good to know. Thanks for the info!

pauladkisson (approved):

rly:
Thanks for the review @pauladkisson !

PR #1084, opened 2026-07-15
Diagnose macOS Intel E2E flake (Electron quitting mid-test)

rly:
## Purpose

Temporary **diagnostic** branch (not for merge) to pin the root cause of the long-standing `E2ELiveServices / macos-15-intel` flakiness.

## Background

The intel E2E job fails intermittently (~4/5 runs on the #1080 conda branch, and similarly on the uv migration branch #1082). It is **pre-existing and dependency-independent**. The failure mechanism: Electron quits gracefully mid-test (`app.on("before-quit")` → `exitPyProc`, `[electron] Exited with code 0`), which tears down the Flask backend and cascades every remaining test to failure (`Failed to fetch`, `dashboard.page` undefined, etc.). It is **not** the backend hot-reload watcher (ruled out with a targeted diagnostic) and **not** electron-vite reload (Electron does not restart after the exit).

## What this adds

Under the test suite only (`runByTestSuite`), logs:
- which signal (`SIGINT`/`SIGTERM`/`SIGHUP`) the Electron main process receives, if any (while preserving graceful shutdown);
- `render-process-gone` / `child-process-gone` events (renderer/GPU crashes);
- a stack trace when `before-quit` fires.

Once CI on the intel runner reveals the trigger, the real fix will replace these diagnostics.

### DIFF
--- a/.github/workflows/deploy_tests_on_pull_request.yml
+++ b/.github/workflows/deploy_tests_on_pull_request.yml
@@ -9,32 +9,9 @@ concurrency:
 
 jobs:
 
-  DevTests:
-    uses: ./.github/workflows/testing_dev.yml
-    secrets:
-      CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
-
-  LiveServices:
-    uses: ./.github/workflows/testing_dev_with_live_services.yml
-    secrets:
-      CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
-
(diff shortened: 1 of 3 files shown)


rly:
Closing: served its purpose. Diagnosis: the macos-15-intel E2E flake is Chromium/Electron subprocess instability on that runner (network-service and GPU-process crashes), producing two modes — app quits gracefully mid-test, or the window never opens ("Network service crashed" → beforeStart timeout). It is pre-existing (fails on conda and uv alike) and not dependency-related. The fix is a job-level E2E retry, which will be developed on a fresh branch off main.

